Issue #40, opened 2017-10-09
Include chainstate validation in block validation


thepiwo:
- this will only work when adding new blocks to the chain, verifying older blocks would result in failure
- verify if the local calculated chainstate, including the changes by transactions in the new block, will match the chainstate roothash in the block header
- TODO @thepiwo: add more details on how to match validity

PR #45, opened 2017-10-10
[GH-#32] Coinbase Transactions

### LINKED ISSUE #32: Coinbase Transaction
- [x] define some hardcoded coinbase transaction value (e.g. 100 tokens)
 - [x] when mining a block add a coinbase transaction to the miners public key, it has no from_acc and no signature
 - [x] check coinbase transaction sum when checking block validity, sum of coinbase transactions values should not exceed the coinbase transaction value
 - a coinbase transaction should never be in the transaction pool

cheezus1:
closes #32

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,5 @@
+language: elixir
+elixir:
+  - 1.5.2
+otp_release:
+  - 20.0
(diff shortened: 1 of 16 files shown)


d-velev:
This also solves #40

thepiwo:
@d-velev can you try to fix the issues after merging master?

PR #49, opened 2017-10-12
[GH-#35] removed aecore.block.* and aecore.txs.tx

### LINKED ISSUE #35: Clean unused functions from Aecore.Block and Structures
maybe remove Aecore.Block.* at all and only use the structures

thepiwo:
resolves #35


apps/aecore/lib/aecore/structures/tx_data.ex
```diff
@@ -15,14 +15,19 @@ defmodule Aecore.Structures.TxData do
...
+             :from_acc,
+             :to_acc,
+             :value]
   use ExConstructor
 
-  @spec create() :: tx_data()
-  def create do
-    TxData.new(%{})
+  @spec create(map()) :: {:ok, %TxData{}}
+  def create(%{from_acc: from_pubkey, to_acc: to_pubkey, value: value}) do
```
d-velev: Why do we need to pass a map to this function? Seems better to do it like in Header.create()

thepiwo: Yes you are correct, I'll change it

thepiwo:
I merge this with master, once #45 is merged

Issue #52, opened 2017-10-17
HTTP-Client Ping


thepiwo:
Use [HTTPoison](https://github.com/edgurgel/httpoison) for this

 - Build a sample client to query the `GET /ping` from #41 expect `pong`

Issue #41, opened 2017-10-09
HTTP-Endpoint Pong


thepiwo:
We should try building this using the [Phoenix Framework](http://phoenixframework.org/)

- Build sample endpoint `GET /ping` return `pong`

Issue #69, opened 2017-10-20
provide get_info endpoint


thepiwo:
- [x] add endpoint to get the clients info, e.g:

```
{
    "current_block_version" : 1,
    "current_block_height" : 2,
    "current_block_hash" : xyz,
    "genesis_block_hash" : abc,
    "peer_count" : 8,
    "difficulty_target" : 21,
    "public_key": pubkey
}
```
- [x] replace ping/pong with get_info, only add peer if genesis block has is correct

PR #71, opened 2017-10-20
GH-53 Create PeerManager and add new Peer

### LINKED ISSUE #53: Create PeerManager and add new Peer
- [x] Build PeerManager GenServer
- [x] `add_peer` function to expect a ip address
- [x] on adding peer `/ping` to see if it is alive, if so, add to a list of peers
- [x] `check_peers` function to check all peers in list if they are still alive, remove if not

d-velev:
resolves #53 , #52 , #41


apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -0,0 +1,63 @@
...
+      {:reply, :ok, List.delete(peers, uri)}
+    else
+      {:reply, :error, peers}
+    end
+  end
+
+  def handle_call(:check_peers, _from, peers) do
+    updated_peers = Enum.filter(peers, fn(peer) ->
+      reply = Client.ping_uri(peer)
+      :ok = reply
```
thepiwo: shouldn't this be `==`

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -0,0 +1,63 @@
...
+
+  @spec all_peers() :: list()
+  def all_peers() do
+    GenServer.call(__MODULE__, :all_peers)
+  end
+
+  def handle_call({:add_peer,uri}, _from, peers) do
+    case(Client.ping_uri(uri)) do
+      :ok ->
+        {:reply, :ok, [uri | peers]}
```
thepiwo: can you avoid adding the same peer, twice

apps/aehttpserver/config/config.exs
```diff
@@ -0,0 +1,23 @@
...
+# and its dependencies with the aid of the Mix.Config module.
+#
+# This configuration file is loaded before any dependency and
+# is restricted to this project.
+use Mix.Config
+
+# Configures the endpoint
+config :aehttpserver, Aehttpserver.Endpoint,
+  url: [host: "localhost"],
+  secret_key_base: "5OCr9Ljxr0s8cSx36fmjKgi6hWETJeoxH+jMvBSNDCO/pt0Zqo1S0+EGGJ2VyOoa",
```
thepiwo: Is this auto-generated? If we don't need it, please remove it

apps/aehttpserver/web/controllers/page_controller.ex
```diff
@@ -0,0 +1,7 @@
+defmodule Aehttpserver.PageController do
```
thepiwo: we don't need this

thepiwo:
please remove all code used for templating, as I see we don't need `gettext` at all.
`apps/aehttpserver/priv/` and `apps/aehttpserver/web/` should mostly be unnecessary.

d-velev:
don't merge, requested changes will be applied in #69

Issue #57, opened 2017-10-18
Test multiple transactions in one block


thepiwo:
all transactions should be in the same block

- [x] account A has 100 tokens, spends 100 to B should succeed
- [x] account A has 100 tokens, spends 110 to B should be invalid
- [x] account A has 100 tokens, spends 40 to B, and two times 30 to C should succeed
- [x] account A has 100 tokens, spends 50 to B, and two times 30 to C, last transaction to C should be invalid, others be included
- [x] account A has 100 tokens, spends 100 to B, B spends 100 to C should succeed if (still) undefined is correct

thepiwo:
do this after #70 so the transaction ordering will be explicitly defined

PR #93, opened 2017-10-25
[GH-#70] Increment transaction nonce for every outgoing transaction

### LINKED ISSUE #70: increment transaction nonce for every outgoing transaction
- [x] start first transaction from one account with nonce 0
- [x] for every outgoing transaction increment the nonce
- [x] keep the current nonce in chain_state
- [x] only allow higher nonces when adding transactions to a block
- [x] order transactions from same account by nonce when taking them out of the transaction pool

cheezus1:
closes #70 #57

### DIFF
--- a/apps/aecore/lib/aecore/chain/chain_state.ex
+++ b/apps/aecore/lib/aecore/chain/chain_state.ex
@@ -18,13 +18,13 @@ defmodule Aecore.Chain.ChainState do
         updated_block_state =
           cond do
             transaction.data.from_acc != nil ->
-              update_block_state(block_state, transaction.data.from_acc, -transaction.data.value)
+              update_block_state(block_state, transaction.data.from_acc, -transaction.data.value, 1)
 
             true ->
               block_state
           end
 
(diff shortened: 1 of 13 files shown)


thepiwo:
The nonce in the chain_state should be set to the highest nonce of the account included in a block, if the first transaction of an account is nonce 100, following transactions with nonces <= 100 should be invalid.

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -91,16 +100,34 @@ defmodule Aecore.Chain.ChainState do
...
+                          nonce:   new_nonce}
     Map.put(block_state_filled_empty, account, new_account_state)
   end
 
   @spec reduce_map_list(list()) :: map()
   defp reduce_map_list(list) do
     List.foldl(list, %{}, fn x, acc ->
       Map.merge(x, acc, fn _key, v1, v2 ->
-        %{balance: v1.balance + v2.balance, nonce: v1.nonce + v2.nonce}
+        new_nonce = cond do
```
thepiwo: can you remove the duplicate code with L41

PR #94, opened 2017-10-25
GH-68 Broadcasting new transactions to other peers

### LINKED ISSUE #68: broadcast new transactions to all peers
- [x] when adding a transaction to the pool send it to every peer in peermanager
 - [x] receive a transaction from another peer and add it to the pool if its not included already

HristianHristov:
Closes #68


apps/aecore/lib/aecore/utils/serialization.ex
```diff
@@ -39,4 +39,20 @@ defmodule Aecore.Utils.Serialization do
         Base.decode16!(data)
     end
   end
+
+  @spec txs(map(), :serialize | :deserialize) :: map() | {:error, term()}
+  def txs(tx, direction) do
+    case direction do
+      :serialize -> 
+        tx 
+        |> :erlang.term_to_binary()
```
thepiwo: like this all of the transaction will be a base64 string, only the binary fields should be handled like this, so a readable json format is used. See the Serialize.block() function.

apps/aehttpclient/lib/client.ex
```diff
@@ -31,4 +31,13 @@ defmodule Aehttpclient.Client do
         :error
     end
   end
+
+  @doc """
+  Send newest transactions to a peer
+  """
+  @spec broadcast_tx(String.t, map()) :: {:ok, map()} | {:error, term()}
+  def broadcast_tx(uri, tx) do
```
thepiwo: rename to `send_tx`

PR #99, opened 2017-10-27
GH-89 endpoint to fetch balance from chainstate account

### LINKED ISSUE #89: endpoint to fetch balance from chainstate account
- [x] GET `/balance/:account:` endpoint, returning number of tokens from chainstate
- [x] if unknown, return 0

d-velev:
closes #89

### DIFF
--- a/apps/aehttpclient/lib/client.ex
+++ b/apps/aehttpclient/lib/client.ex
@@ -27,6 +27,10 @@ defmodule Aehttpclient.Client do
     Enum.each(peers, fn{peer, _} -> Peers.add_peer(peer) end)
   end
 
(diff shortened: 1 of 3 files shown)


apps/aehttpclient/lib/client.ex
```diff
@@ -38,13 +42,19 @@ defmodule Aehttpclient.Client do
...
+            standard_response(body)
         end
       {:ok, %HTTPoison.Response{status_code: 404}} ->
         :error
       {:error, %HTTPoison.Error{}} ->
         :error
     end
   end
+
+  def standard_response(body) do
```
thepiwo: `json_response` would be a better name, keep to semantic naming

PR #100, opened 2017-10-30
GH-90 endpoint to get transactions from pool concerning account

### LINKED ISSUE #90: endpoint to get transactions from pool concerning account
- [x] GET `/tx_pool/:account:` returning all transactions having from_acc or to_acc
- [x] return empty list if none found

d-velev:
closes #90


apps/aehttpclient/lib/client.ex
```diff
@@ -41,10 +49,12 @@ defmodule Aehttpclient.Client do
           :info ->
             response = Poison.decode!(body, keys: :atoms!)
             {:ok, response}
-          :peers ->
```
thepiwo: why is `:peers` and `:balance` removed?

d-velev: :peers and :balance will be handled by the default json_response()

thepiwo: great

thepiwo:
can you remove the compiler warnings as well?

thepiwo:
can you add a test as well for getting an existinng tx from the pool and one where there is none transaction?

PR #101, opened 2017-10-30
[GH-#83] Added fee to transactions

### LINKED ISSUE #83: add fee to transaction
- [x] add fee value to the transaction
- [x] the fee will should be reduced from `from_acc` e.g. balance 100 --> `from_acc` 99 `fee` 1 
- [x] the miner should be able to collect all fees from transactions in the block in addition to the coinbase

cheezus1:
closes #83


apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -79,6 +82,25 @@ defmodule Aecore.Chain.ChainState do
       Enum.all?()
   end
 
+  @spec add_fees_to_block_state(list(), map()) :: map()
+  def add_fees_to_block_state(txs, block_state) do
+    pubkey = elem(Keys.pubkey(), 1)
```
thepiwo: Why is the pubkey accessed here?

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -79,6 +82,25 @@ defmodule Aecore.Chain.ChainState do
       Enum.all?()
   end
 
+  @spec add_fees_to_block_state(list(), map()) :: map()
+  def add_fees_to_block_state(txs, block_state) do
```
thepiwo: What does this function do?

cheezus1: it adds the sum of the fees to the given account in the block state

thepiwo:
I think the `sum_coinbase_transactions` is wrong, it should add only coinbase transactions values and compare with the `Miner.coinbase_transaction_value()` + sum of tx fees

PR #102, opened 2017-10-30
[GH-75] Sending/Receiving Blocks

### LINKED ISSUE #75: broadcast new block to all peers
- [x] If a block is mined, broadcast it to all peer 
- [x] if a block is received from other peers, validate it and add it to the chain
- [x] after validation broadcast it further to other clients

meivantodorov:
### General description

- Validated blocks are sent to other peers via http request asynchronously
- Received blocks are validated and added to the chain if validation is successful 
- Improved the flow for sending asyc request.
 
closes #75


apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -109,6 +111,11 @@ defmodule Aecore.Chain.Worker do
         |> BlockValidation.block_header_hash()
         |> Base.encode16()}, total tokens: #{total_tokens}"
       end)
+
+      ## Block was validated, now we can send it to other peers
+      serialized_block = Serialization.block(b, :serialize)
```
thepiwo: the serialization should not bein the chain

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -109,17 +123,28 @@ defmodule Aecore.Peers.Worker do
...
     {:noreply, peers}
   end
+
   def handle_cast(_any, peers) do
-    {:noreply, peers}
+    Logger.info("[Peers] Unhandled cast message:  #{inspect(peers)}")
+  end
+
+  ## Internal functions
+  defp send_to_peers(:post, uri, data, []) do
```
thepiwo: there sould be no spectial case, `for peer <- peers do` should work with empty list. You can log to debug after sending to how many peers it was sent

apps/aecore/lib/aecore/txs/pool/worker.ex
```diff
@@ -44,8 +45,8 @@ defmodule Aecore.Txs.Pool.Worker do
     if is_tx_valid do
       updated_pool = Map.put_new(tx_pool, :crypto.hash(:sha256, :erlang.term_to_binary(tx)), tx)
       case tx_pool == updated_pool do
-        true -> Logger.info(" This transaction has already been added")
-        false -> Peers.broadcast_tx(tx)
+        true -> Logger.info(" This transaction already has been added")
+        false -> Peers.async_request({:new_tx, Serialization.tx(tx, :serialize)})
```
thepiwo: do the serialization by case in `async_request`

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -52,6 +52,20 @@ defmodule Aecore.Peers.Worker do
     |> Base.encode16()
   end
 
+  @doc """
+  Every async requests to the peers will be send from here
+  """
+  @spec async_request({type :: atom(), data :: term()}) :: :ok | :error
```
thepiwo: this should have a more semantic correct name, like `broadcast_all_peers`

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -109,17 +123,28 @@ defmodule Aecore.Peers.Worker do
...
   def handle_cast(_any, peers) do
-    {:noreply, peers}
+    Logger.info("[Peers] Unhandled cast message:  #{inspect(peers)}")
+  end
+
+  ## Internal functions
+  defp send_to_peers(:post, uri, data, []) do
+    Logger.warn("Empty peers list")
+  end
+  defp send_to_peers(:post, uri, data, peers) do
```
thepiwo: sending will always happen with `:post`, no need to specify. What is `uri` needed for, each peer should have its own uri

meivantodorov: https://github.com/aeternity/elixir-research/blob/GH-75/apps/aehttpclient/lib/client.ex#L37

thepiwo: I have to look at it in more detail, thanks

thepiwo: Can you rename it to something else? In the peers worker, `uri` is always used as the address of the peer, this is why I was so confused

thepiwo:
can you resolve the compiler warnings, as well?

thepiwo:
When having two nodes, mining on the first one, suspending and then continue mining on the second node `[error] Failed to add block: Incorrect previous hash` occures, but the blocks are still added to the chain and broadcasted.

PR #106, opened 2017-11-01
[GH-#88] Refactor chain to be map of blocks, traverse top to bottom

### LINKED ISSUE #88: refactor chain to be map of blocks, traverse top to bottom
- [x] replace list in chain worker with map
- [x] reference blocks by hash in this map
- [x] keep head_block_hash list additionally to map in genserver state (should be a list because there could be multiple valid head blocks in the future)
- [x] to get list of blocks, traverse from latest, via prev_block_hash to the bottom
- [x] refactor block validation to have only block and chainstate as parameter. get the prev_block by hash from the chain, and the traverse the chain to get blocks for difficulty check

cheezus1:
closes #88


apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -77,57 +85,101 @@ defmodule Aecore.Chain.Worker do
...
-    [prior_block | _] = chain
-    new_block_state = ChainState.calculate_block_state(b.txs)
-    new_chain_state = ChainState.calculate_chain_state(new_block_state, prev_chain_state)
+  def handle_call({:add_block, %Block{} = block}, _from, state) do
+    {chain, chain_state} = state
+    prev_block_chain_state = chain_state[block.header.prev_hash]
+    new_block_state = ChainState.calculate_block_state(block.txs)
+    new_chain_state = ChainState.calculate_chain_state(new_block_state, prev_block_chain_state)
+
+    BlockValidation.validate_block!(block, chain[block.header.prev_hash], new_chain_state)
```
thepiwo: maybe add the `try` back

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -77,57 +85,101 @@ defmodule Aecore.Chain.Worker do
...
     else
       {:reply, {:error, "Block not found"}, state}
     end
   end
 
-  def handle_call(:all_blocks, _from, state) do
-    chain = elem(state, 0)
-    {:reply, chain, state}
+  def handle_call({:get_block_by_hex_hash, hash}, _from, state) do
+    case(Enum.find(elem(state, 0), fn{block_hash, _block} ->
```
thepiwo: use  `{chain, _} = state` instead of `elem(state, 0)`

PR #107, opened 2017-11-02
GH-91 Add list of transactions concerning account to chainstate (block_hash, tx_hash)

### LINKED ISSUE #91: add list of transactions concerning account to chainstate (block_hash, tx_hash)
- [x] Keep the chain_state as it is. `%{acc => %{balance: balance, latest_nonce: latest nonce}}`
- [x] make a txs_index which is `%{acc => [{block_hash, tx_hash}]}` this gets updated on add_block but is not kept in the block itself

d-velev:
closes #91


apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -23,8 +23,8 @@ defmodule Aecore.Chain.Worker do
     genesis_block_map = %{genesis_block_hash => Block.genesis_block()}
     genesis_chain_state = ChainState.calculate_block_state(Block.genesis_block().txs)
     latest_block_chain_state = %{genesis_block_hash => genesis_chain_state}
-
-    initial_state = {genesis_block_map, latest_block_chain_state}
+    txs_index = %{}
```
thepiwo: build the initial index from the `Block.genesis_block().txs` even if there are none

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -164,6 +174,34 @@ defmodule Aecore.Chain.Worker do
     get_blocks(latest_block_hash, latest_block_obj.header.height)
   end
 
+  def acc_txs_info(block) do
```
thepiwo: rename to something more understandable, what it actually does

PR #110, opened 2017-11-02
GH-#79 Ignores own peer in handling peers

### LINKED ISSUE #79: ignore own peer in handling peers
- [x] on peer startup generate a random nonce
- [x] add the nonce to get info
- [ ] ignore all peers with our nonce

HristianHristov:
Resolves #79


apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -15,14 +15,14 @@ defmodule Aecore.Peers.Worker do
   require Logger
 
   def start_link(_args) do
-    GenServer.start_link(__MODULE__, %{}, name: __MODULE__)
+    GenServer.start_link(__MODULE__, %{peers: %{}, nonce: :rand.uniform(1000)}, name: __MODULE__)
```
thepiwo: make the nonce range bigger

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -15,14 +15,14 @@ defmodule Aecore.Peers.Worker do
...
   end
 
   ## Client side
 
   @spec add_peer(term) :: :ok | {:error, term()} | :error
   def add_peer(uri) do
-    GenServer.call(__MODULE__, {:add_peer, uri})
+    GenServer.call(__MODULE__, {:add_peer, uri}, 10000)
```
thepiwo: why is this hardcoded?

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -63,31 +70,38 @@ defmodule Aecore.Peers.Worker do
...
+              updated_peers = Map.put(peers, uri, info.current_block_hash)
+              Logger.info(fn -> "Added #{uri} to the peer list" end)
+              {:reply, :ok, %{state | peers: updated_peers}}
+            else
+              Logger.error(fn ->
+                "Failed to add #{uri}, genesis header hash not valid" end)
+              {:reply, {:error, "Genesis header hash not valid"}, %{state | peers: peers}}
+            end
+          true -> 
+            Logger.error(fn ->
```
thepiwo: should not be error, only debug

apps/aecore/lib/aecore/peers/worker.ex
HristianHristov: Because the request to self takes more than 5000ms, so we had to hardcode a higher value for the sake of the test we make

thepiwo: ah okay, I did mistake it with the nonce

apps/aecore/lib/aecore/peers/worker.ex
thepiwo: something big

PR #117, opened 2017-11-08
[GH-104] Loading previous state on startup

### LINKED ISSUE #104: read blockchain and chainstate to disk on startup

meivantodorov:
closes #104 

- removed the reference from persistence worker.
- loading previous blockchain state on start up


apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -27,7 +28,7 @@ defmodule Aecore.Chain.Worker do
     txs_index = calculate_block_acc_txs_info(Block.genesis_block())
     initial_state = {genesis_block_map, latest_block_chain_state, txs_index}
 
-    {:ok, initial_state}
+    {:ok, initial_state, 2000}
```
cytadela8: Could you elaborate why do we wait 2sec. after start to load the chain from disk? Is this because `Aecore.Persistence.Worker` may not be started?

If so we should find another solution of the problem. I'm not aware of any clean way in Elixir to make `Chain.Worker` depend on `Persistance.Worker`, so I would even suggest merging `Persistance.Worker` with `Chain.Worker`.

meivantodorov: Yes, the reason is `Aecore.Persistence.Worker` may not be started yet. I will search for better solution.

thepiwo: @meivantodorov any progress on this?

apps/aecore/apps/aecore/logs/2017-10-19_info.log
```diff
@@ -0,0 +1,102 @@
+10:39:27.463 [info] Mined block #1 with a difficulty target of 1
```
thepiwo: Don't check in your logs, can you amend commit and force push this commit?

apps/aecore/lib/aecore/utils/persistence.ex
```diff
@@ -0,0 +1,31 @@
+defmodule Aecore.Utils.Persistence do
+  @moduledoc """
+  Store/Restore latest blockchain and chainstate
+  """
+
+  @persistence_table Application.get_env(:aecore, :persistence)[:table]
```
cytadela8: "@" constants are calculated on build time, not runtime. Please make a similiar change to commit 950943113c15842ff9256d348553a1a3d75df562

meivantodorov:
This is the way I stop gracefully the node

:init.stop()


`:init.stop()`

cytadela8:
Now everything is ok.

Issue #115, opened 2017-11-03
getting random peer from peers we connect to


thepiwo:
- [ ] randomize peers to get peers from
 - [ ] add new peers randomly until limit
 - [ ] maybe remove some peers randomly if at limit to get some variation

PR #120, opened 2017-11-10
[GH-#111] After adding a peer add his peers

### LINKED ISSUE #111: after adding a peer add his peers
- [x] get peers from other nodes to keep an preset number of peers in memory; Download peers from every peer, choose peers that we don't know, add to our list.
 - [x] add a maximum of 50 peers to add to the list, after that don't add any new ones

cheezus1:
closes #111

### DIFF
--- a/apps/aecore/lib/aecore/peers/worker.ex
+++ b/apps/aecore/lib/aecore/peers/worker.ex
@@ -48,6 +48,11 @@ defmodule Aecore.Peers.Worker do
     GenServer.call(__MODULE__, :get_peers_nonce)
   end
 
+  @spec add_peers_from_peers() :: :ok
+  def add_peers_from_peers() do
+    GenServer.call(__MODULE__, :add_peers_from_peers)
+  end
+
   @spec genesis_block_header_hash() :: term()
   def genesis_block_header_hash() do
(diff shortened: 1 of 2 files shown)


cytadela8:
Hmm... I didn't know about this issue. This conflicts with my work on #115.

cytadela8:
I suggest, I will try to propose an overall solution to peer sync this evening.

thepiwo:
Yeah maybe it is conflicting, but we need both issues solved. Could you integrate these changes in your work?

cytadela8:
@thepiwo I will try to do that this evening, but at the moment I think this PR conflicts with my design choices. I will try to think how to make everything work this evening and probably I will propose some solution.

cytadela8:
After analysing possibilities I took the test fix and `max_peers` limit from this code. Sadly most of this code conflicts with peer sync design I would like to introduce with #123. Sorry for not realizing the conflict earlier!

thepiwo:
@cytadela8 so we merge this, prior to #123?

cytadela8:
@thepiwo We ignore this. I cherry-picked one commit and rewritten `max_peers`.

thepiwo:
replaced with #123

Issue #118, opened 2017-11-09
Server timeout when checking peers_nonoce


cytadela8:
We recently introduced peers `nonce` check in `Aecore.Peers.Worker.add_peer`. The check was introduced to prevent node from adding itself to list of known peers by making `\info` request to peer that's about to be added and comparing it's nonce to own nonce. This doesn't work as expected, because `\info` requests uses `Aecore.Peers.Worker` so it can't be processed while `:add_peer` call to `Aecore.Peers.Worker` genserver is handled.

Error introduced by 2055e9fb (GH-79)

To reproduce:
```
iex(9)> Aecore.Peers.Worker.add_peer("http://localhost:4000") 
[info] GET /info
[debug] Processing with Aehttpserver.InfoController.info/2
  Parameters: %{}
  Pipelines: [:browser]
:error
iex(10)> [error] GET /info request error
[error] #PID<0.415.0> running Aehttpserver.Endpoint terminated
Server: localhost:4000 (http)
Request: GET /info
** (exit) exited in: GenServer.call(Aecore.Peers.Worker, :get_peers_nonce, 5000)
    ** (EXIT) time out

```

cytadela8:
@HristianHristov take a look

d-velev:
A fix would be to make the nonce for each peer a module attribute in the Peers module and have a getter function for it, instead of having it in the state. I think that will only work if the nonce doesn't change.

cytadela8:
Nonce doesn't change but has to be chosen randomly on start.

HristianHristov:
I will fix it.

HristianHristov:
@cytadela8 Have a look.

d-velev:
Have the peer_nonce attribute in the Peer module and make a normal function that returns it (get_peer_nonce) and have that function called in the info_controller when we want to get the nonce from a peer, that fixes it.

HristianHristov:
@cytadela8 , @d-velev , Think it's done now. How do you think ?

cytadela8:
Looks good. Can you submit PR and I will test and merge later when I'm on computer?

cytadela8:
I just checked and module attributes (`@`) are evaluated compile time. Despite that currently we force users to compile the project for themselves we want to me able to distribute compiled binaries. Having the same nonce on many peers in unacceptable.

cytadela8:
I think the easiest and quite good solution would be to store the nonce in ETS table.

gspasov:
[...]

PR #121, opened 2017-11-10
[GH-#118] Server timeout when checking peers_nonce

HristianHristov:
Resolves GH-#118

### DIFF
--- a/apps/aecore/lib/aecore/peers/worker.ex
+++ b/apps/aecore/lib/aecore/peers/worker.ex
@@ -16,9 +16,10 @@ defmodule Aecore.Peers.Worker do
   require Logger
 
   @mersenne_prime 2147483647
+  @peer_nonce :rand.uniform(@mersenne_prime)
 
   def start_link(_args) do
-    GenServer.start_link(__MODULE__, %{peers: %{}, nonce: :rand.uniform(@mersenne_prime)}, name: __MODULE__)
+    GenServer.start_link(__MODULE__, %{}, name: __MODULE__)
   end
 
   ## Client side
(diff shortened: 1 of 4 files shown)


apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -66,44 +62,52 @@ defmodule Aecore.Peers.Worker do
...
     {:ok, initial_peers}
   end
 
-def handle_call({:add_peer,uri}, _from, %{peers: peers, nonce: own_nonce} = state) do
+def handle_call({:add_peer,uri}, _from, peers) do
     case(Client.get_info(uri)) do
       {:ok, info} ->
-        case own_nonce == info.peer_nonce do
-          false ->  
+        case Aecore.Peers.Worker.get_peer_nonce() == info.peer_nonce do
```
d-velev: we don't have to call the function here, just use @peer_nonce

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -10,21 +10,23 @@ defmodule Aecore.Peers.Worker do
   alias Aecore.Utils.Blockchain.BlockValidation
   alias Aehttpclient.Client, as: HttpClient
   alias Aecore.Utils.Serialization
+  alias Aecore.Peers.Scheduler, as: Scheduler
```
d-velev: we don't need this

apps/aehttpserver/web/controllers/info_controller.ex
```diff
@@ -6,6 +6,7 @@ defmodule Aehttpserver.InfoController do
   alias Aecore.Utils.Blockchain.BlockValidation
   alias Aecore.Keys.Worker, as: Keys
   alias Aecore.Peers.Worker, as: Peers
+  alias Aecore.Peers.Scheduler, as: Scheduler
```
d-velev: we don't need this

cytadela8:
I just checked and `Module attributes are evaluated at compile time, not runtime.`, that makes this solution invalid.

apps/aehttpserver/web/controllers/info_controller.ex
gspasov: We have missed those. A quick fix is on its way.

Issue #119, opened 2017-11-09
Changes for local multinode


cytadela8:
We should make it possible to launch multiple local nodes without conflicts by changing environment variables. To do that we have to make it possible to change following configs with environment variables:
- [x] persistence table path
- [x] node port

thepiwo:
Yes we could make the port phoenix listens on, configurable via environment variable

thepiwo:
@meivantodorov can you check why the persistance breaks when starting multiple nodes? I tried with `PORT=4001 iex -S mix phx.server` and `PORT=4002 iex -S mix phx.server` in a different terminal

Issue #116, opened 2017-11-03
when being added by a new peer, try to add it as well


thepiwo:
- [x] when a peer we don't know yet calls our `/info` find his uri (phenix should have some kind of method) and try to add him as well, if we don't know him and if the genesis_block matches
- [x] maybe check if it is a aeclient type of peer through `server` response header

thepiwo:
@cheezus1 @d-velev whats the status of this?

PR #123, opened 2017-11-11
[GH-115] Peers sync

### LINKED ISSUE #115: getting random peer from peers we connect to

cytadela8:
### This PR solves following problems:
- When we know less then `peers_target_count` peers we add random peers from peers known by peers we know. (We request peers list from peers we know and add random peers from that list)
- Max known peers count is now limited. When we try to add another one, but the max count is reached, we at random: remove one of known or ignore the new one.
- When we know at least `peers_target_count` peers we remove one randomly every "sync loop".

### Stuff missing for ready peers sync:
- [ ] adding peer when unknown node connects to us. (#116)
- [ ] testing (#119)

This replaces #120 
This closes #115


cytadela8:
Please check and advice about code style. This is my first contribution so I probably violated some established rules.

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -72,28 +75,35 @@ defmodule Aecore.Peers.Worker do
...
-        {:reply, :error, %{state | peers: peers}}
+  def handle_call({:add_peer,uri}, _from, %{peers: peers, nonce: own_nonce} = state) do
+    if Map.has_key?(peers, uri) do
+      Logger.debug(fn ->
+              "Skipped adding #{uri}, already known" end)
+      {:reply, {:error, "Peer already known"}, state}
+    else
+      case check_peer(uri, own_nonce) do
+        {:ok, info} ->
+          if map_size(peers) < @peers_max_count
```
thepiwo: can we refactor this if condition into a function, to give it a name

thepiwo:
code style is good so far, we have no specific rules yet

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -189,6 +188,11 @@ defmodule Aecore.Peers.Worker do
     end
   end
 
+  defp should_a_peer_be_added peers_count do
```
cytadela8: @thepiwo Is this what you thought of?

thepiwo: yes great!

PR #124, opened 2017-11-13
GH-#118 Server timeout when checking peers_nonce

### LINKED ISSUE #118: Server timeout when checking peers_nonoce

HristianHristov:
Resolves #118


apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -69,23 +62,42 @@ defmodule Aecore.Peers.Worker do
...
+  def get_peer_nonce() do
+    case :ets.info(:nonce_table) do
+      :undefined -> create_nonce_table()
+      _ -> :table_created
+    end
+    case :ets.lookup(:nonce_table, :nonce) do
+      [] ->
+        rand = :rand.uniform(@mersenne_prime)
+        :ets.insert(:nonce_table, {:nonce, rand})
+        rand
```
cytadela8: I would rename `rand` to `nonce` as it's a little bit confusing.

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -69,23 +62,42 @@ defmodule Aecore.Peers.Worker do
...
+    end
+  end
+
   ## Server side
 
   def init(initial_peers) do
     {:ok, initial_peers}
   end
 
-  def handle_call({:add_peer,uri}, _from, %{peers: peers, nonce: own_nonce} = state) do
```
cytadela8: I would leave the the peers map wrapped in a state map. In future we might want to store more data in state.

cytadela8:
I made the test that should have caught the problem earlier more precise.

apps/aecore/lib/aecore/peers/worker.ex
thepiwo: @HristianHristov will you adjust this as well?

Issue #97, opened 2017-10-25
implement initial sync from latest block to genesis block


thepiwo:
#88 has to be finished first
only do this in pair programming

 - [x] blocks sync:
   1. ask random peers about top and get top block of random peer(s) if we don't have it 
   2. for every block that we don't know the parent ask random peers for parent. Optional: mark block as "bad" if no-one can provide it's parent
   - [x] for every block we get in i., ii.:
      a. "single" validate transactions without chainstate, coinbase sum, version and root_hash(discard if invalid)
      b. get parent if we don't have parent
   3. for every "single" validated block that has not been "fully" validated:
      a. if parent is not validate. Validate parent.
      b. If parent does not exist skip.
      c. validate the block
   4. compare height of topmost block with chainstate height. If it's higher advance chainstate:
      a. generate list of block between chainstate and highest block by traversing from top. If you find an marked as "invalid" block, mark all of it's children as invalid. (all blocks from top block to that block including top)
      b. if you get to genesis while looking for chainstate block: regenerate chainstate from genesis. (We have to have some chainstate snapshots in future to avoid that)
      c. validate blocks in order beginning with child of chainstate block. Update chainstate while doing that.
      d. if validation fails at one point mark the block as "invalid" and mark all it's children as "invalid".
      e. if validation failed at one point try again to get the topmost block and if it's not chainstate block try to advance chainstate as above described.

cytadela8:
This was fixed by #126.

PR #126, opened 2017-11-15
[GH-#97] Implement initial sync from latest block to genesis block

cheezus1:
**Overview:**
When we add a new peer, we get all unknown blocks from the peer, build a chain from them and add the chain to ours, if the earliest block's parent from the new chain is the latest one from our chain. Also, whenever we receive a broadcasted block, we add it to the Sync state and then do the same operation. 

**The rest is described in #97 .**

There is a concurrency problem when two or more miners mine simultaneously. The chain's state can be updated and have a different value than the one we expect (our latest block can end up different than the one our sync worker gets when checking what the highest block of our chain is). This will be solved later on, in a different issue.


apps/aecore/lib/aecore/utils/blockchain/block_validation.ex
```diff
@@ -54,6 +54,27 @@ defmodule Aecore.Utils.Blockchain.BlockValidation do
     end
   end
 
+  @spec single_validate_block(%Block{}) :: {:error, term()} | :ok
```
thepiwo: call this as well in `validate_block` to minimize duplicate code

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -15,30 +17,89 @@ defmodule Aecore.Peers.Sync do
...
+    {:reply, :ok, updated_state}
+  end
+
+  def handle_call(:add_valid_peer_blocks_to_chain, _from, state) do
+    updated_state = single_validate_all_blocks(state)
+
+    filtered_state = Enum.reduce(updated_state, updated_state, fn({block_hash, %{peer: peer}}, acc) ->
+          built_chain = build_chain(acc, {block_hash, peer}, [])
+          add_built_chain(built_chain)
+          remove_added_blocks_from_state(acc)
```
cytadela8: We should combine this two functions into one (`add_built_chain` and `remove_added_blocks_from_state`). Going though whole `acc` every time can be time consuming.

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -98,11 +159,102 @@ defmodule Aecore.Peers.Sync do
...
+            peer_block_hash =
+              BlockValidation.block_header_hash(deserialized_block.header)
+            status =
+              case(HttpClient.get_block({peer_uri, Base.encode16(peer_block_hash)})) do
+                {:ok, _peer_block_parent} ->
+                  :good
+                :error ->
+                  :bad
+              end
+
```
cytadela8: We should *single_validate* the blocks here. This way:
- we don't store info about invalid blocks
- we *single_validate* each block only ones.

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -98,11 +159,102 @@ defmodule Aecore.Peers.Sync do
...
+  defp check_peer_block(peer_uri, block_hash, blocks_with_status) do
+    case Chain.get_block_by_hex_hash(block_hash) do
+      {:error, _} ->
+        case(HttpClient.get_block({peer_uri, block_hash})) do
+          {:ok, peer_block} ->
+            deserialized_block = Serialization.block(peer_block, :deserialize)
+            peer_block_hash =
+              BlockValidation.block_header_hash(deserialized_block.header)
+            status =
+              case(HttpClient.get_block({peer_uri, Base.encode16(peer_block_hash)})) do
```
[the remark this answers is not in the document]
cheezus1: It was correct in a previous commit, we have changed it later with this, now it is ok

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -18,30 +17,66 @@ defmodule Aecore.Peers.Sync do
...
+  def handle_call(:get_state, _from, state) do
+    {:reply, state, state}
+  end
+
+  def handle_call({:add_block_to_state, block_hash, block}, _from, state) do
+    updated_state =
+      case Chain.has_block?(Base.encode16(block_hash)) do
+        true ->
+          state
+        false ->
```
cytadela8: We could move block_single_validate hear so we never store not validated blocks in memory.

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -18,30 +17,66 @@ defmodule Aecore.Peers.Sync do
...
+          Map.put(state, block_hash, block)
```
cytadela8: We should single validate the block here, so we never store not validated blocks in memory. This function is used by aehttpserver and (I think) it doesn't validate the blocks so we can get an invalid block from there.

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -112,5 +147,60 @@ defmodule Aecore.Peers.Sync do
       end
     end)
   end
-end
 
+  # Builds a chain, starting from the given block,
+  # until we reach a block, of which the previous block is the highest in our chain
+  # (that means we can add this chain to ours)
+  defp build_chain(state, block, chain) do
```
cytadela8: I attach my previous comment. This is only optimization so we could leave it as is, but it would be nice to change to the logic suggested in the comment.

> Aside from what I commented, I would suggest to change from build_chain function to sorting by height and then simply going through all blocks just once. The problem with current solution is: if we get a chain of blocks of length n, but we fail to get a parent block of that chain (an peer gives us a lot of unknown blocks, but no parent of those blocks) then we will call build_chain for every block in that chain and go through all of it's parents every time. So on average we will analyse the block in build_chain n/2 times.

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -112,5 +147,60 @@ defmodule Aecore.Peers.Sync do
...
+    case Chain.has_block?(block_hash) do
+      false ->
+        case(HttpClient.get_block({peer_uri, block_hash})) do
+          {:ok, peer_block} ->
+            deserialized_block = Serialization.block(peer_block, :deserialize)
+            try do
+              BlockValidation.single_validate_block(deserialized_block)
+              peer_block_hash =
+                BlockValidation.block_header_hash(deserialized_block.header)
+
```
cytadela8: We should check here if `peer_block_hash` matches `block_hash`.

apps/aecore/lib/aecore/peers/sync.ex
thepiwo: we discussed, not to optimize for now, we can do it at a later point in time, as this would introduce some more complexity for now

apps/aecore/lib/aecore/peers/sync.ex
thepiwo: yes, great find!

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -49,7 +49,13 @@ defmodule Aecore.Peers.Sync do
         true ->
           state
         false ->
-          Map.put(state, block_hash, block)
+          try do
+            BlockValidation.single_validate_block(block)
+            Map.put(state, block_hash, block)
+          catch
+            {:error, message} ->
+              Logger.error(fn -> "Can't add block to Sync state; #{message}" end)
```
cytadela8: What happens with `state` in this case?
After `Logger.error(...)` there should be a line = `state`. So when there is an error in validation `updated_state = state`

apps/aecore/lib/aecore/peers/sync.ex
cytadela8: ok

cytadela8:
Well done!

PR #137, opened 2017-11-21
[GH-#116] When being added by a new peer, try to add it as well

### LINKED ISSUE #116: when being added by a new peer, try to add it as well

d-velev:
closes #116


apps/aehttpserver/web/controllers/info_controller.ex
```diff
@@ -18,11 +20,27 @@ defmodule Aehttpserver.InfoController do
...
+    if(!(peer == host || host == "localhost:4000")) do
+      case(Map.has_key?(Peers.all_peers, peer)) do
+        true ->
+          Logger.info("Peer already in our list")
+        false ->
+          Peers.schedule_add_peer(peer)
+      end
+    end
+
+    conn = Plug.Conn.put_resp_header(conn, "server", "aehttpserver")
```
thepiwo: can you put this to a general place, for all endpoints to be active, or does this have to exactly here?

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -189,10 +203,13 @@ defmodule Aecore.Peers.Worker do
...
         case own_nonce == info.peer_nonce do
           false ->
-            if(info.genesis_block_hash == genesis_block_header_hash()) do
-              {:ok, info}
-            else
-              {:error, "Genesis header hash not valid"}
+            cond do
+              info.genesis_block_hash != genesis_block_header_hash() ->
+                {:error, "Genesis header hash not valid"}
+              !Map.has_key?(info, :server) || info.server != "aehttpserver"->
```
cytadela8: Why actually we block peers that are not "aehttpserver"? If someone develops another AE client it's ok. As for validation if we are adding an AE node: we already check genesis_block_hash. If someone returns valid one they probably are an AE node.

apps/aehttpclient/lib/client.ex
```diff
@@ -44,14 +44,17 @@ defmodule Aehttpclient.Client do
...
+      {:ok, %{body: body, headers: headers, status_code: 200}} ->
         case(identifier) do
           :block ->
             response = Poison.decode!(body, as: %Block{}, keys: :atoms!)
             {:ok, response}
           :info ->
             response = Poison.decode!(body, keys: :atoms!)
-            {:ok, response}
+            {_, server} = Enum.find(headers, fn(header) ->
+              header == {"server", "aehttpserver"} end)
```
cytadela8: We force `server` header to be equal "aehttpserver" also here. I wonder: what happens if it's not? Do we get an exception that then goes back to `Aecore.Peers.Worker` and causes it to restart?

apps/aecore/lib/aecore/peers/worker.ex
thepiwo: for now it is an easy check to filter other clients (e.g. a wallet) to try to connect

apps/aehttpclient/lib/client.ex
thepiwo: @d-velev what do you think?

PR #140, opened 2017-11-22
[GH-#116] Fixed adding a peer back when his port is different than 4000

d-velev:


cytadela8:
That probably may not be connected with this changes, but we are again getting exceptions when two peers talk to each other:

```
[error] Failed to add peer. reason=Request error
[error] GenServer Aecore.Peers.Sync terminating                                      
** (stop) exited in: GenServer.call(Aecore.Peers.Worker, {:add_peer, "localhost:4002"}, 5000)                                                                             
    ** (EXIT) time out
    (elixir) lib/gen_server.ex:774: GenServer.call/3
    (aecore) lib/aecore/peers/sync.ex:105: anonymous fn/3 in Aecore.Peers.Sync.get_newpeers_and_add/1
    (elixir) lib/enum.ex:1826: Enum."-reduce/3-lists^foldl/2-0-"/3
    (aecore) lib/aecore/peers/sync.ex:68: Aecore.Peers.Sync.refill/0
    (aecore) lib/aecore/peers/sync.ex:28: Aecore.Peers.Sync.handle_info/2
    (stdlib) gen_server.erl:616: :gen_server.try_dispatch/4
    (stdlib) gen_server.erl:686: :gen_server.handle_msg/6
Last message: :work
State: %{}
[error] #PID<0.426.0> running Aehttpserver.Endpoint terminated
Server: localhost:4002 (http)
Request: GET /info
** (exit) exited in: GenServer.call(Aecore.Peers.Worker, :all_peers, 5000)
    ** (EXIT) time out
```

cytadela8:
See #141 for detailed explanation of problem

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -219,7 +216,7 @@ defmodule Aecore.Peers.Worker do
                 {:ok, info}
             end
           true ->
-            {:error, :peer_has_own_nonce}
+            {:error, "Equal peer nonces"}
```
thepiwo: why is the `:peer_has_own_nonce` removed, if we don't have it we will log an error for everytime we try to add ourself, we should just ignore this,and assume everything is ok

cytadela8: I agree we should revert that.

thepiwo: please don't merge next time if there are still open comments ;)

cytadela8: If you check the timing you will see I merged one day after my approval and before you commented ;) But I agree it was a mistake.

Issue #141, opened 2017-11-22
Loop when adding peer_add with own uri


cytadela8:
We recently added added code to add a peer that added us (makes an \info request to us). This results in a self adding loop. Lets say we get from somewhere (`\peers` request for example) our own uri. We then call it (`\info`) to check it's nonce (this is the way we try to find out if we are not trying to add ourselves to peer list). That results with `add_peer` call with our own uri from `\info` handler.

Currently there is some code that tries to work around that by checking the uri if it's "localhost:$PORT" where $PORT is our port. This is not the way to go. There are many uri's that may point to our own node and we use nonce to detect them. 

Exception example:
```
[error] Failed to add peer. reason=Request error
[error] GenServer Aecore.Peers.Sync terminating                                      
** (stop) exited in: GenServer.call(Aecore.Peers.Worker, {:add_peer, "localhost:4002"}, 5000)                                                                             
    ** (EXIT) time out
    (elixir) lib/gen_server.ex:774: GenServer.call/3
    (aecore) lib/aecore/peers/sync.ex:105: anonymous fn/3 in Aecore.Peers.Sync.get_newpeers_and_add/1
    (elixir) lib/enum.ex:1826: Enum."-reduce/3-lists^foldl/2-0-"/3
    (aecore) lib/aecore/peers/sync.ex:68: Aecore.Peers.Sync.refill/0
    (aecore) lib/aecore/peers/sync.ex:28: Aecore.Peers.Sync.handle_info/2
    (stdlib) gen_server.erl:616: :gen_server.try_dispatch/4
    (stdlib) gen_server.erl:686: :gen_server.handle_msg/6
    (stdlib) proc_lib.erl:247: :proc_lib.init_p_do_apply/3
Last message: :work
State: %{}
[error] #PID<0.426.0> running Aehttpserver.Endpoint terminated
Server: localhost:4002 (http)
Request: GET /info
** (exit) exited in: GenServer.call(Aecore.Peers.Worker, :all_peers, 5000)
    ** (EXIT) time out
```

PR #145, opened 2017-11-27
[GH-#142] Don't add a peer with the same nonce twice

### LINKED ISSUE #142: Don't add a peer with the same nonce twice
this can happen if you add "localhost" but in discovery the address will resolve to "127.0.0.1" then we will have both peers in our list.

d-velev:
fixes #142, #141


thepiwo:
when having 3 connected nodes and killing one of them it can happen to get `no match of right hand side value` at `{_, info} = Client.get_info(uri)` (Peers.Worker:127)

thepiwo:
@d-velev can we fix this as well?
https://github.com/aeternity/elixir-research/pull/140#pullrequestreview-78857136

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -162,28 +168,42 @@ defmodule Aecore.Peers.Worker do
...
-                random_peer = Enum.random(Map.keys(peers))
-                Logger.debug(fn -> "Max peers reached. #{random_peer} removed" end)
-                Map.delete(peers, random_peer)
-              else
-                peers
-              end
-            updated_peers = Map.put(peers_update1, uri, info.current_block_hash)
-            Logger.info(fn -> "Added #{uri} to the peer list" end)
-            {:reply, :ok, %{state | peers: updated_peers}}
+          if(!Map.has_key?(peers, info.peer_nonce)) do
```
cytadela8: This check should be inside check_peer function. The purpose of `check_peer` function is to separate all of our checks from the main logic. (if inside if, inside if, inside... doesn't "look nice")

d-velev: We need the state for this check and we don't have it in check_peer, passing it just for that seems unnecessary

thepiwo: maybe we just create a separate function for it

apps/aehttpserver/web/controllers/info_controller.ex
```diff
@@ -29,17 +29,13 @@ defmodule Aehttpserver.InfoController do
...
     host = conn.host <> host_port
 
-    if !(peer == host || host == "localhost:" <> host_port) do
-      case(Map.has_key?(Peers.all_peers, peer)) do
-        true ->
-          Logger.info("Peer already in our list")
-        false ->
-          Peers.schedule_add_peer(peer)
-      end
+    if(!(peer_nonce == own_nonce || peer == host || host == "localhost:" <> host_port)) do
```
thepiwo: maybe try check "127.0.0.1:" <> host_port as well

apps/aehttpserver/web/controllers/info_controller.ex
```diff
@@ -29,26 +29,20 @@ defmodule Aehttpserver.InfoController do
...
-      case(Map.has_key?(Peers.all_peers, peer)) do
-        true ->
-          Logger.info("Peer already in our list")
-        false ->
-          Peers.schedule_add_peer(peer)
-      end
-    end
+    if(!(peer_nonce == own_nonce)) do
+      Peers.schedule_add_peer(peer, peer_nonce)
+   end
```
untra: weird indentation here

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -162,28 +175,41 @@ defmodule Aecore.Peers.Worker do
...
+            if should_a_peer_be_added(map_size(peers)) do
+              peers_update1 =
+              if map_size(peers) >= @peers_max_count do
```
untra: I would consider extracting this into its own function to enhance readability:

```elixir
peers_update1 = trim_peers(peers)

...

def trim_peers(peers) do
  if map_size(peers) >= @peers_max_count do
    random_peer = peers |> Map.keys() |> Enum.random()
    Logger.debug(fn -> "Max peers reached. #{random_peer} removed" end)
    Map.delete(peers, random_peer)
  else
    peers
  end
end
```

apps/aehttpserver/web/controllers/info_controller.ex
```diff
@@ -29,26 +29,20 @@ defmodule Aehttpserver.InfoController do
     peer_ip = conn.peer |> elem(0) |> Tuple.to_list |> Enum.join(".")
     peer_port = Plug.Conn.get_req_header(conn, "peer_port") |> Enum.at(0) |> to_string()
     peer_port = ":" <> peer_port
-    host_port = ":" <> to_string(conn.port)
+    peer_nonce = Plug.Conn.get_req_header(conn, "nonce") |> Enum.at(0) |> String.to_integer()
```
cytadela8: Good solution!

apps/aecore/lib/aecore/peers/worker.ex
cytadela8: It's definitely better now. We could pass `peers` to `check_peer` and move the check, but I think it's good enough now.

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -125,39 +123,48 @@ defmodule Aecore.Peers.Worker do
...
   def handle_call(:all_peers, _from, %{peers: peers} = state) do
-    {:reply, peers, %{state | peers: peers}}
+    {:reply, peers, state}
   end
 
   ## Async operations
+  def handle_cast({:schedule_add_peer, uri, nonce}, %{peers: peers} = state) do
+    already_have_peer = Map.has_key?(peers, nonce)
+    state =
+      case already_have_peer do
```
cytadela8: Please change `case` -> `if`.

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -125,39 +123,48 @@ defmodule Aecore.Peers.Worker do
...
       end, peers)
-
     updated_peers =
-      for {peer, current_block_hash} <- filtered_peers, into: %{} do
-        {_, info} = Client.get_info(peer)
-        if info.current_block_hash != current_block_hash do
-          {peer, info.current_block_hash}
+      for {nonce, %{uri: uri, latest_block: latest_block}} <- filtered_peers, into: %{} do
+        {_, info} = Client.get_info(uri)
+        if(info.current_block_hash != latest_block) do
```
cytadela8: Please remove parenthesis.

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -113,44 +113,57 @@ defmodule Aecore.Peers.Worker do
   is updated if the one in the latest GET /info request is different.
   """
   def handle_call(:check_peers, _from, %{peers: peers} = state) do
-    filtered_peers = :maps.filter(fn(peer, _) ->
-        case Client.get_info(peer) do
-          {:ok, info} -> info.genesis_block_hash == genesis_block_header_hash()
-          _ -> false
+    filtered_peers = :maps.filter(fn(_, %{uri: uri}) ->
+        case Client.get_info(uri) do
```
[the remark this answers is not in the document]
d-velev: Don't think it can be done here

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -145,8 +151,10 @@ defmodule Aecore.Peers.Sync do
       #if we have successfully added less then number_of_peers_to_add peers then try to add another one
       if acc < number_of_peers_to_add do
```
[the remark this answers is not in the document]
cytadela8: We can do it like that, but I don't like relying on lazy evaluation for determining if a function with side effects should be invoked.

PR #148, opened 2017-11-27
[GH-#147] get blocks route from_hash with count

### LINKED ISSUE #147: add route to get blocks

thepiwo:
route for https://github.com/aeternity/web-wallet/issues/7

closes #147

### DIFF
--- a/apps/aehttpserver/web/controllers/block_controller.ex
+++ b/apps/aehttpserver/web/controllers/block_controller.ex
@@ -18,6 +18,32 @@ defmodule Aehttpserver.BlockController do
     end
   end
 
+  def get_blocks(conn, params) do
+    latest_block_hash = case Map.get(params, "from_block") do
+      nil ->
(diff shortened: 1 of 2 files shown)


apps/aehttpserver/web/controllers/block_controller.ex
```diff
@@ -18,6 +18,29 @@ defmodule Aehttpserver.BlockController do
...
+      hash ->
+        {_, hash_bin} = Base.decode16(hash)
+        hash_bin
+    end
+
+    count = case Map.get(params, "count") do
+      nil ->
+        100
+      count_string ->
+        {number, _} = Integer.parse(count_string)
```
cytadela8: We could add an limit here.

thepiwo: I think it's ok for now, we have to optimize this anyway in the future (e.g. only give header)

PR #151, opened 2017-11-29
[GH-114] Refactoring of serialization

### LINKED ISSUE #114: refactor serialization to be in aeutil
- [x] Move (de)serialization code to aeutil, for code shared between multiple apps

cytadela8:
This PR:
- creates Aeutil application and moves serialization code to it.
- moves invocations of `Aeutil.Serialization` to `Aehttpclient`
- moves other request related code to `Aehttpclient`

This closes #114


apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -189,15 +189,14 @@ defmodule Aecore.Peers.Sync do
...
-            deserialized_block = Serialization.block(peer_block, :deserialize)
+          {:ok, deserialized_block} ->
             try do
               BlockValidation.single_validate_block(deserialized_block)
               peer_block_hash =
                 BlockValidation.block_header_hash(deserialized_block.header)
 
               if(block_hash == Base.encode16(peer_block_hash)) do
-                check_peer_block(peer_uri, peer_block.header.prev_hash,
+                check_peer_block(peer_uri, Serialization.hex_binary(deserialized_block.header.prev_hash, :serialize),
```
cytadela8: Here I had to do a little bit dirty fix, because we use hex encoded hashes in requests and hex encoded hash is passed to `check_peer_block`

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -191,7 +192,7 @@ defmodule Aecore.Chain.Worker do
       Persistence.write_block_by_hash(block)
 
       ## Block was validated, now we can send it to other peers
-      Peers.broadcast_to_all({:new_block, block})
+      Client.broadcast_block(block)
```
thepiwo: I'd like to divide the client as much from the rest of the code as possible. Only the Peers Worker should know about the HTTP-Client Interface

apps/aecore/lib/aecore/peers/sync.ex
thepiwo: internally I'd prefer to only pass binary, we should only introduce other encodings at the last point needed

apps/aecore/lib/aecore/txs/pool/worker.ex
```diff
@@ -46,7 +47,7 @@ defmodule Aecore.Txs.Pool.Worker do
       updated_pool = Map.put_new(tx_pool, :crypto.hash(:sha256, :erlang.term_to_binary(tx)), tx)
       case tx_pool == updated_pool do
         true -> Logger.info(" This transaction already has been added")
-        false -> Peers.broadcast_to_all({:new_tx, tx})
+        false -> Client.broadcast_tx(tx)
```
thepiwo: Again, only the Peers should know about the Client

thepiwo:
@meivantodorov can you approve and merge?

Issue #138, opened 2017-11-22
make coinbase transaction time locked


thepiwo:
- [x] time lock coinbase transactions for a few blocks (e.g. 10)

can be started after #125

PR #155, opened 2017-11-30
[GH-#125] Add lock time to transaction

### LINKED ISSUE #125: add lock time to transaction
- [x] add `lock_time_block` field to the transaction
- [x] make the transaction amount not spendable, in any block with number prior to the one given in `lock_time_block`

cheezus1:
closes #125, #138


apps/aecore/config/dev.exs
```diff
@@ -63,3 +63,6 @@ config :aecore, :pow,
 config :aecore, :peers,
   peers_target_count: 3,
   peers_max_count: 4
+
+config :aecore, :tx_data,
+  lock_time_block: 10
```
thepiwo: name it `lock_time_coinbase`

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -20,13 +20,17 @@ defmodule Aecore.Chain.ChainState do
...
-                                 transaction.data.nonce)
+                                 transaction.data.nonce, transaction.data.lock_time_block, true)
 
             true ->
               block_state
           end
 
-        update_block_state(updated_block_state, transaction.data.to_acc, transaction.data.value, 0)
+        add_to_amount = latest_block_height + 1 !=
+          transaction.data.lock_time_block - Application.get_env(:aecore, :tx_data)[:lock_time_block]
```
thepiwo: don't get the env here, pass it to the function

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -68,29 +72,59 @@ defmodule Aecore.Chain.ChainState do
 
   @spec calculate_total_tokens(map()) :: integer()
   def calculate_total_tokens(chain_state) do
-    chain_state |>
-      Enum.map(fn{_account, data} -> data.balance end) |>
-      Enum.sum()
+    chain_state
```
thepiwo: maybe return a tuple of {total_tokens, total_unlocked_tokens, total_locked_tokens}

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -68,29 +72,59 @@ defmodule Aecore.Chain.ChainState do
...
+      {locked_amount, updated_locked} =
+          Enum.reduce(locked, {0, []}, fn(%{amount: amount, block: lock_time_block}, {amount_update_value, updated_locked}) ->
+            cond do
+              lock_time_block > new_block_height ->
+                {amount_update_value, updated_locked ++ [%{amount: amount, block: lock_time_block}]}
+              lock_time_block == new_block_height ->
+                {amount_update_value + amount, updated_locked}
+
+              true ->
+                {amount_update_value, updated_locked}
```
cytadela8: This shouldn't happen. Let's log an error here.

apps/aecore/test/aecore_chain_state_test.exs
```diff
@@ -10,41 +10,76 @@ defmodule AecoreChainStateTest do
...
+    lock_time_block = Enum.at(block.txs, 0).data.lock_time_block
+    latest_block = Chain.latest_block()
 
-    assert %{"a" => %{balance:  1, nonce: 102},
-             "b" => %{balance: -2, nonce: 1},
-             "c" => %{balance: -3, nonce: 1}} ==
-      ChainState.calculate_block_state(block.txs)
+    assert %{"a" => %{balance: -9, nonce: 102,
+                      locked: [%{amount: 10, block: lock_time_block}]},
+             "b" => %{nonce: 1, balance: -11,
```
cytadela8: I see that's the same as before, but why do we actually get negative balances?

thepiwo: whoops, good question

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -99,8 +133,16 @@ defmodule Aecore.Chain.ChainState do
         block_state_filled_empty[account].nonce
     end
 
-    new_account_state = %{balance: block_state_filled_empty[account].balance + value,
-                          nonce:   new_nonce}
+    new_locked = if(value > 0) do
```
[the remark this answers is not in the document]
cheezus1: That's fixed

apps/aecore/test/aecore_chain_state_test.exs
cheezus1: Because amounts are spent this block, but the received amounts for each are locked for one block? Correct me if I am wrong

thepiwo: @cytadela8 this is just the block_state indicating the change this block makes to the chain_state, negative values occur for every token spent.

thepiwo:
@cheezus1 is there a test to check if locked token can't actually be spent?

apps/aecore/lib/aecore/chain/chain_state.ex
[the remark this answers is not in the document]
cytadela8: Ok, so if `value < 0` then `add_to_amount` should be always true. If `add_to_amount` is false then the call is a no-op and we should produce an error to logs or maybe raise some exception or assert? @thepiwo What should we do if there in such a case?

apps/aecore/test/aecore_chain_state_test.exs
cytadela8: Ohh, right. This is not the chain_state after the block. It's delta chain_state.

apps/aecore/lib/aecore/chain/chain_state.ex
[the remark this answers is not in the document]
cytadela8: Ok, so if `value < 0` then `add_to_amount` should be always true. If `add_to_amount` is false then the call is a no-op and we should produce an error to logs or maybe raise some exception or assert? @thepiwo What should we do if there in such a case?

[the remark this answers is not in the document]
thepiwo: I guess we should raise an exception, but why handle the case at all?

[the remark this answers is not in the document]
thepiwo: @cheezus1 here is the answer

apps/aecore/test/multiple_transactions_test.exs
```diff
@@ -30,13 +30,17 @@ defmodule MultipleTransactionsTest do
...
+                              Application.get_env(:aecore, :tx_data)[:lock_time_coinbase] + 1)
     assert :ok = Pool.add_transaction(tx)
     Miner.resume()
     Miner.suspend()
     Pool.get_and_empty_pool()
     tx = create_signed_tx(account1, account2, 99,
-                          Map.get(Chain.chain_state, account1_pub_key, %{nonce: 0}).nonce + 1, 1)
+                          Map.get(Chain.chain_state, account1_pub_key, %{nonce: 0}).nonce + 1, 1,
+                          Chain.latest_block().header.height +
+                            Application.get_env(:aecore, :tx_data)[:lock_time_coinbase] + 1)
```
cytadela8: Why do we add locktime to all of the transactions in tests?

apps/aecore/lib/aecore/keys/worker.ex
```diff
@@ -40,7 +41,8 @@ defmodule Aecore.Keys.Worker do
 
   """
   @spec sign_tx(binary(), integer(), integer(), integer(), integer()) :: {:ok, %SignedTx{}}
-  def sign_tx(to_acc, value, nonce, fee, lock_time_block) do
+  def sign_tx(to_acc, value, nonce, fee,
+              lock_time_block \\ Chain.latest_block().header.height + 1) do
```
thepiwo: why do you want to lock a transaction by default? I would set the default to `0`

cytadela8:
@cheezus1 We still have merge conflicts

PR #157, opened 2017-11-30
[GH-113] Sync worker state functions

### LINKED ISSUE #113: sync worker state functions
- [x] is the highest known block from peers equal to our highest block? (or are we in the middle of analyzing new blocks)
 - [x] are all the blocks, until the highest block, fully validated?
 - [x] only allow the miner to run if both are true

d-velev:
do not merge before #142 
closes #113

### DIFF
--- a/apps/aecore/lib/aecore/peers/worker.ex
+++ b/apps/aecore/lib/aecore/peers/worker.ex
@@ -155,7 +155,7 @@ defmodule Aecore.Peers.Worker do
     updated_peers =
       for {nonce, %{uri: uri, latest_block: latest_block}} <- filtered_peers, into: %{} do
         {_, info} = Client.get_info(uri)
-        if(info.current_block_hash != latest_block) do
+        if info.current_block_hash != latest_block do
           {nonce, %{uri: uri, latest_block: info.current_block_hash}}
         else
           {nonce, %{uri: uri, latest_block: latest_block}}
--- a/apps/aecore/lib/aecore/peers/worker.ex
+++ b/apps/aecore/lib/aecore/peers/worker.ex
@@ -114,7 +114,7 @@ defmodule Aecore.Peers.Worker do
           end
         end)
     is_synced =
-      if(Enum.count(peer_uris) == 0) do
+      if(Enum.empty?(peer_uris)) do
         true
       else
         Enum.max(peer_latest_block_heights) <= local_latest_block_height
@@ -185,17 +185,12 @@ defmodule Aecore.Peers.Worker do
   end
 
   def handle_cast({:schedule_add_peer, uri, nonce}, %{peers: peers} = state) do
-    already_have_peer = Map.has_key?(peers, nonce)
-    state =
-      case already_have_peer do
-        true ->
-          state
-        false ->
-          {:reply, _, state} = add_peer(uri, state)
-          state
-      end
-
-    {:noreply, state}
+    if Map.has_key?(peers, nonce) do
+      {:noreply, state}
+    else
+      {:reply, _, newstate} = add_peer(uri, state)
+      {:noreply, newstate}
+    end
   end
 
   def handle_cast(any, state) do
@@ -221,7 +216,7 @@ defmodule Aecore.Peers.Worker do
           if(!Map.has_key?(peers, info.peer_nonce)) do
             if should_a_peer_be_added(map_size(peers)) do
               peers_update1 =
-              if map_size(peers) >= @peers_max_count do
+                if map_size(peers) >= @peers_max_count do
                   random_peer = Enum.random(Map.keys(peers))
                   Logger.debug(fn -> "Max peers reached. #{random_peer} removed" end)
                   Map.delete(peers, random_peer)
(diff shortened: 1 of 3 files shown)


apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -113,44 +144,57 @@ defmodule Aecore.Peers.Worker do
...
+    send_to_peers(type, data, peer_uris)
     {:noreply, state}
   end
 
-  def handle_cast({:schedule_add_peer, uri}, state) do
-    {:reply, _, state} = add_peer(uri, state)
+  def handle_cast({:schedule_add_peer, uri, nonce}, %{peers: peers} = state) do
+    already_have_peer = Map.has_key?(peers, nonce)
+    state =
+      case already_have_peer do
```
untra: In general, I would not use `case ... do` statements to match on booleans. An if/else statement works just fine. 

```elixir
if Map.has_key?(peers, nonce) do
  {:noreply, state}
else
  {:reply, _, newstate} = add_peer(uri, state)
  {:noreply, newstate}
end
```

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -92,7 +99,31 @@ defmodule Aecore.Peers.Worker do
...
+      Enum.map(peer_uris, fn(uri) ->
+          case Client.get_info(uri) do
+            {:ok, info} ->
+              info.current_block_height
+            :error ->
+              0
+          end
+        end)
+    is_synced =
+      if(Enum.count(peer_uris) == 0) do
```
untra: Favor `Enum.empty?` instead of `Enum.count(...) == 0`

`Enum.empty?` works in constant time, whereas `Enum.count` must first count all the elements of a collection, operating in O(N)

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -162,28 +206,43 @@ defmodule Aecore.Peers.Worker do
...
-                peers
-              end
-            updated_peers = Map.put(peers_update1, uri, info.current_block_hash)
-            Logger.info(fn -> "Added #{uri} to the peer list" end)
-            {:reply, :ok, %{state | peers: updated_peers}}
+          if(!Map.has_key?(peers, info.peer_nonce)) do
+            if should_a_peer_be_added(map_size(peers)) do
+              peers_update1 =
+              if map_size(peers) >= @peers_max_count do
+                  random_peer = Enum.random(Map.keys(peers))
```
untra: questionable identing here

thepiwo: @d-velev could you change this not to be nested 3 nested if

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -113,44 +144,57 @@ defmodule Aecore.Peers.Worker do
...
       end, peers)
-
     updated_peers =
-      for {peer, current_block_hash} <- filtered_peers, into: %{} do
-        {_, info} = Client.get_info(peer)
-        if info.current_block_hash != current_block_hash do
-          {peer, info.current_block_hash}
+      for {nonce, %{uri: uri, latest_block: latest_block}} <- filtered_peers, into: %{} do
+        {_, info} = Client.get_info(uri)
+        if(info.current_block_hash != latest_block) do
```
cytadela8: Please remove unnecessary parenthesis.

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -92,7 +99,31 @@ defmodule Aecore.Peers.Worker do
...
 
-  def handle_call({:add_peer, uri}, _from, state) do
+  def handle_call(:is_chain_synced, _from, %{peers: peers} = state) do
+    local_latest_block_height = Chain.latest_block().header.height
+    peer_uris = peers
+      |> Map.values()
+      |> Enum.map(fn(%{uri: uri}) -> uri end)
+    peer_latest_block_heights =
+      Enum.map(peer_uris, fn(uri) ->
+          case Client.get_info(uri) do
```
cytadela8: We actually store in state latest_block_info for every peer. Please use that instead so we won't make any requests when handling this call.

thepiwo: yeah, good find!

d-velev: We only store latest_block_hash, in the state, if we don't get the info here we would have to do check_peers before mining which would do these requests anyway, I think it's fine like it is now but i'm open to suggestions

apps/aecore/lib/aecore/peers/worker.ex
d-velev: which parenthesis are unnecessary?

cytadela8: around statement after if. It should be:
`if info.current_block_hash != latest_block do`

apps/aecore/lib/aecore/peers/worker.ex
thepiwo: @cytadela8 should be fine

thepiwo:
@meivantodorov @andonov can you approve?

Issue #144, opened 2017-11-27
fix Cuckoo Cycle crashes


thepiwo:

thepiwo:
`beam.smp: ./cuckoo/src/cuckoo_miner.hpp:374: void cuckoo_ctx::solution(node_t*, u32, node_t*, u32): Assertion `n==PROOFSIZE' failed.`

PR #158, opened 2017-12-06
[GH-#92] Endpoint to get transactions concerning account from chain

### LINKED ISSUE #92: endpoint to get transactions concerning account from chain
- [x] GET `/tx/:account:` endpoint returning all transactions concerning account
tx can be fetched from chainstate, find block, filter transactions
- [x] GET `/tx/:account:?include_proof=true` include merkle proof and block_hash in result, to get the proof fetch the block by hash, build the tx merkle tree, filter the transaction by hash and create the proof via https://github.com/aeternity/gb_merkle_trees/blob/dev/src/gb_merkle_trees.erl#L128

HristianHristov:
Resolves #92 , Merge after #144


apps/aehttpserver/web/controllers/tx_controller.ex
```diff
@@ -0,0 +1,63 @@
...
+    user_txs = Pool.get_txs_for_address(account_bin, :no_hash)
+    case user_txs do
+      [] -> json(conn, [])
+      _ ->
+        case params["include_proof"]  do
+          "true" ->
+            user_txs_and_hash = Pool.get_txs_for_address(account_bin, :add_hash)
+            merkle_tree = build_tx_tree(user_txs)
+            {size, {key, value, _}} = merkle_tree
+            proof = :gb_merkle_trees.merkle_proof(key, merkle_tree)
```
cytadela8: What's the idea behind this "proof" value? It's calculated based on the transactions we send to the user. An useful proof would be if for every transaction we included a Merkle Proof that proves the transaction was included in a particular block, given the user that verifies the proof has the header for that block.

apps/aecore/lib/aecore/txs/pool/worker.ex
```diff
@@ -40,6 +44,20 @@ defmodule Aecore.Txs.Pool.Worker do
     GenServer.call(__MODULE__, :get_and_empty_pool)
   end
 
+
```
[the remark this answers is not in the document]
thepiwo: The variable should get some semantic correct name, making it self-explaining. e.g. `include_txs_hash` as boolean

apps/aecore/test/aecore_get_txs_for_address.ex
```diff
@@ -0,0 +1,67 @@
...
+
+    {:ok, tx1} = Keys.sign_tx(elem(Keys.pubkey(), 1), 5, 1, 1)
+    {:ok, tx2} = Keys.sign_tx(elem(user_pubkey, 1), 7, 1, 1)
+    {:ok, tx3} = Keys.sign_tx(elem(Keys.pubkey(), 1), 5, 1, 1)
+
+    assert :ok = Pool.add_transaction(tx1)
+    assert :ok = Pool.add_transaction(tx2)
+    assert :ok = Pool.add_transaction(tx3)
+    Miner.resume()
+    :timer.sleep(10000)
```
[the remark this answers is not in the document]
thepiwo: yes, without sleeping should be sufficient

apps/aehttpserver/web/controllers/tx_controller.ex
thepiwo: yes, this should be done as @cytadela8 said. The user will have the possibility to download the block header by hash and verify this proof

apps/aehttpserver/web/controllers/tx_controller.ex
```diff
@@ -2,42 +2,55 @@ defmodule Aehttpserver.TxController do
...
+            for block <- blocks_for_user_txs do
+              build_tx_tree(block.txs)
+            end
+            user_txs_trees = Enum.zip(user_txs, merkle_trees)
+            proof =
+            for user_tx_tree <- user_txs_trees  do
+              {tx, tree} = user_tx_tree
+              {_, {key, value, _}} = tree
+              merkle_proof = :gb_merkle_trees.merkle_proof(key, tree)
+              verification =
```
thepiwo: There should be the merkle path for the user to verify himself

this works for the simple example with only the coinbase transaction:
```
serialized_merkle_proof = Enum.map(Tuple.to_list(merkle_proof), fn merkle_proof_line ->
    Serialization.hex_binary(merkle_proof_line, :serialize)
end)
Map.put_new(tx, :proof, serialized_merkle_proof)
```

But it fails if there are multiple transactions in one block, here is a test case:
```
alias Aecore.Miner.Worker, as: Miner
Miner.resume()
Miner.suspend()
{:ok, to_account} = Keys.pubkey()
{:ok, tx} = Keys.sign_tx(to_account, 5, 1, 1)
Pool.add_transaction(tx)
Miner.resume()
Miner.suspend()
```

the problem is that `tree` is actually a tree of proves and not ` {_, {key, value, _}}` maybe you have to traverse the tree and serialize every key and value to hex

thepiwo:
Except for https://github.com/aeternity/elixir-research/pull/158#discussion_r156311032 it looks really good

apps/aecore/lib/aecore/txs/pool/worker.ex
```diff
@@ -45,16 +46,31 @@ defmodule Aecore.Txs.Pool.Worker do
...
-  def get_txs_for_address(address, option) do
-    GenServer.call(__MODULE__, {:get_txs_for_address,{address, option}})
+  @spec get_txs_for_address(String.t()) :: list()
+  def get_txs_for_address(address) do
+    GenServer.call(__MODULE__, {:get_txs_for_address, address})
+  end
+
+  @spec get_block_by_txs_hash(binary()) :: %Block{}
+  def get_block_by_txs_hash(txs_hash) do
+   GenServer.call(__MODULE__, {:get_block_by_txs_hash, txs_hash})
```
cytadela8: Only one space of indentation after `do` (should be two spaces)

apps/aehttpserver/web/controllers/tx_controller.ex
```diff
@@ -0,0 +1,104 @@
...
+    account_bin =
+      params["account"]
+      |> Base.decode16!()
+    user_txs = Pool.get_txs_for_address(account_bin)
+    case user_txs do
+      [] -> json(conn, [])
+      _ ->
+        case params["include_proof"]  do
+          "true" ->
+            blocks_for_user_txs =
```
cytadela8: We should move all of the proof adding code to a function so we can test it. (and create such tests)

apps/aehttpserver/web/controllers/tx_controller.ex
```diff
@@ -0,0 +1,104 @@
...
+            user_txs_trees = Enum.zip(user_txs, merkle_trees)
+            proof =
+            for user_tx_tree <- user_txs_trees  do
+              {tx, tree} = user_tx_tree
+              transaction = :erlang.term_to_binary(
+                tx
+                |> Map.delete(:txs_hash)
+                |> Map.delete(:block_hash)
+                |> Map.delete(:block_height)
+                |> Map.delete(:signature)
```
cytadela8: Please refactor the code so we separate the transaction data used in hashing from other data. 

The hashing should be moved to a function in aeutil or some better place.

apps/aehttpserver/web/controllers/tx_controller.ex
```diff
@@ -0,0 +1,104 @@
...
+              serialized_merkle_proof = serialize_merkle_proof(merkle_proof, [])
+              verification =
+                case :gb_merkle_trees.verify_merkle_proof(key, transaction, tx.txs_hash, merkle_proof) do
+                  {:ok, :verified} -> :verified
+                  {:error, {:key_hash_mismatch, _}} -> :key_hash_mismatch
+                  {:error, {:value_hash_mismatch, _}} -> :value_hash_mismatch
+                  {:error, {:root_hash_mismatch, _}} -> :root_hash_mismatch
+                end
+              tx
+              |> Map.put_new(:verification, verification)
```
cytadela8: "verification" variable isn't usefull for client. Only it's use is if we add an assert `assert verification == :verified`

apps/aehttpserver/web/controllers/tx_controller.ex
```diff
@@ -0,0 +1,104 @@
...
+                    to_acc: Serialization.hex_binary(tx.to_acc, :serialize),
+                    txs_hash: Serialization.hex_binary(tx.txs_hash, :serialize),
+                    block_hash: Serialization.hex_binary(tx.block_hash, :serialize),
+                    signature: Serialization.hex_binary(tx.signature, :serialize)
+                   } end))
+        end
+    end
+  end
+
+  defp build_tx_tree(txs) do
```
cytadela8: This function also is a copy probably.

apps/aecore/lib/aecore/txs/pool/worker.ex
```diff
@@ -67,4 +116,98 @@ defmodule Aecore.Txs.Pool.Worker do
     {:reply, tx_pool, %{}}
   end
 
+  def get_txs_for_address_with_proof(user_txs) do
```
cytadela8: Proper name would be `add proof to txs`. This function doesn't *get* any txs. It is given txs and adds proof to them.

cytadela8: Proper name would be `add proof to txs`. This function doesn't *get* any txs. It is given txs and adds proof to them.

apps/aecore/lib/aecore/txs/pool/worker.ex
```diff
@@ -67,4 +116,98 @@ defmodule Aecore.Txs.Pool.Worker do
...
+      {tx, tree} = user_tx_tree
+      transaction = :erlang.term_to_binary(
+        tx
+        |> Map.delete(:txs_hash)
+        |> Map.delete(:block_hash)
+        |> Map.delete(:block_height)
+        |> Map.delete(:signature)
+        |> TxData.new()
+      )
+      key = :crypto.hash(:sha256, transaction)
```
cytadela8: We talked about extracting the transaction hashing to `Aecore.Structures.TxData` module or some utility module, remember? ;)

apps/aecore/test/aecore_get_txs_for_address.ex
```diff
@@ -0,0 +1,66 @@
...
+  alias Aecore.Utils.Blockchain.BlockValidation, as: BlockValidation
+  alias Aecore.Txs.Pool.Worker, as: Pool
+
+
+  @tag timeout: 100_000
+  test "get txs for given address test" do
+    {:ok, address} = Keys.pubkey()
+
+    ## Create unique address
+    address_hex =
```
cytadela8: It's ok, but you could have just hardcoded anything.

apps/aecore/lib/aecore/txs/pool/worker.ex
```diff
@@ -67,4 +116,98 @@ defmodule Aecore.Txs.Pool.Worker do
...
+        |> Map.put_new(:signature, tx.signature)| user_txs]
+    end
+
+    check_address_tx(txs, address, user_txs)
+  end
+  defp check_address_tx([], address, user_txs) do
+    user_txs
+  end
+
+  defp build_tx_tree(txs) do
```
cytadela8: This is a copy of `calculate_root_hash` from `Aecore.Chain.BlockValidation` please merge this functions.

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -128,7 +128,7 @@ defmodule Aecore.Chain.BlockValidation do
   end
 
   @spec calculate_root_hash(list()) :: binary()
-  def calculate_root_hash(txs) when length(txs) == 0 do
+  def calculate_root_hash(txs) when txs == [] do
```
thepiwo: prefer Enum.empty?(txs) for this

HristianHristov: Can't invoke Enum.empty? in a guard

thepiwo: ok I don't know

apps/aecore/lib/aecore/txs/pool/worker.ex
```diff
@@ -78,4 +115,81 @@ defmodule Aecore.Txs.Pool.Worker do
...
+    for user_tx_tree <- user_txs_trees  do
+      {tx, tree} = user_tx_tree
+      key = :erlang.term_to_binary(
+        tx
+        |> Map.delete(:txs_hash)
+        |> Map.delete(:block_hash)
+        |> Map.delete(:block_height)
+        |> Map.delete(:signature)
+        |> TxData.new()
+      )
```
thepiwo: this looks weird, can you explain what it does?

HristianHristov: It removes the keys which are not for the transaction and then turn the map into a TxData stucture.

thepiwo: can you refactor `get_txs_for_address` to give you the TxData and not some map, then you don't have to reconstruct your transaction here. If any more attibutes are needed you can return a tuple.

apps/aecore/lib/aecore/txs/pool/worker.ex
```diff
@@ -78,4 +115,81 @@ defmodule Aecore.Txs.Pool.Worker do
...
+  def add_proof_to_txs(user_txs) do
+    blocks_for_user_txs =
+    for user_tx <- user_txs do
+      get_block_by_txs_hash(user_tx.txs_hash)
+    end
+    merkle_trees =
+    for block <- blocks_for_user_txs do
+      BlockValidation.build_merkle_tree(block.txs)
+    end
+    user_txs_trees = Enum.zip(user_txs, merkle_trees)
```
meivantodorov: Please combine all `for do` in one and remove the zipping part

apps/aecore/lib/aecore/txs/pool/worker.ex
```diff
@@ -78,4 +115,81 @@ defmodule Aecore.Txs.Pool.Worker do
...
+    proof =
```
meivantodorov: Variable `proof` it is not used

apps/aecore/lib/aecore/txs/pool/worker.ex
```diff
@@ -78,4 +115,81 @@ defmodule Aecore.Txs.Pool.Worker do
...
+      _ ->
+        block_user_txs =
+      for block_user_txs <- user_txs do
+        block_user_txs
+        |>  Map.put_new(:txs_hash, block.header.txs_hash)
+        |>  Map.put_new(:block_hash, block.header.chain_state_hash)
+        |>  Map.put_new(:block_height, block.header.height)
+      end
+        split_blocks(blocks, address, block_user_txs)
+    end
```
meivantodorov: You can return `block_user_txs` for each case and just call the `split_blocks()` once. Also `if do` can replace `case do` here

apps/aecore/test/aecore_get_txs_for_address.ex
```diff
@@ -0,0 +1,66 @@
...
+        |> TxData.new()
+      )
+      key = TxData.hash_tx(transaction)
+      assert {:ok, :verified} =
+        :gb_merkle_trees.verify_merkle_proof(key,
+          transaction,
+          user_tx_with_proof.txs_hash,
+          user_tx_with_proof.proof)
+    end
+  end
```
meivantodorov: Missing `end` and some more syntax errors that You should fix. Also the file extension should be `.exs`.

apps/aecore/test/aecore_get_txs_for_address.ex
```diff
@@ -0,0 +1,66 @@
...
+    {:ok, tx3} = Keys.sign_tx(elem(Keys.pubkey(), 1), 5, 1, 1)
+
+    assert :ok = Pool.add_transaction(tx1)
+    assert :ok = Pool.add_transaction(tx2)
+    assert :ok = Pool.add_transaction(tx3)
+    Miner.resume()
+    Miner.suspend()
+
+    assert 2 <= :erlang.length(Chain.all_blocks)
+    assert [tx2] = Pool.get_txs_for_address(address_bin)
```
meivantodorov: I think that here you mean `assert [^tx2] = Pool.get_txs_for_address(address_bin)`

thepiwo:
@meivantodorov @cytadela8 can you re-review?

apps/aecore/lib/aecore/txs/pool/worker.ex
```diff
@@ -78,4 +115,81 @@ defmodule Aecore.Txs.Pool.Worker do
     {:reply, tx_pool, %{}}
   end
 
+  @doc """
+  A function that adds a merkle proof for every single transaction
+  """
+
+  @spec add_proof_to_txs(list()) :: list()
+  def add_proof_to_txs(user_txs) do
```
untra: what is `user_txs` a list of? are these `%SignedTx{}` objects?
* If so, can you specify that in the spec? we will also need to update the existing [SignedTx](https://github.com/aeternity/elixir-research/blob/master/apps/aecore/lib/aecore/structures/signed_tx.ex) structure with the appropriate fields (tx.txs_hash, tx.data, etc..)
* If not, what structure is this? Does it need to be defined for this PR?

apps/aecore/lib/aecore/txs/pool/worker.ex
```diff
@@ -78,4 +115,81 @@ defmodule Aecore.Txs.Pool.Worker do
...
+      |> TxData.hash_tx()
+      merkle_proof = :gb_merkle_trees.merkle_proof(key, tree)
+      serialized_merkle_proof = serialize_merkle_proof(merkle_proof, [])
+      Map.put_new(tx, :proof, serialized_merkle_proof)
+    end
+  end
+
+  ## Private functions
+
+  defp split_blocks([block | blocks], address, txs) do
```
untra: `@spec split_blocks(list(%Block{}, String().t, list(%SignedTx{})) :: list(%SignedTx{})`

apps/aecore/lib/aecore/txs/pool/worker.ex
thepiwo: It returns the merkle proof for each transaction as api endpoint to the wallet, so the wallet can in theory verify them

apps/aecore/lib/aecore/txs/pool/worker.ex
```diff
@@ -78,4 +115,81 @@ defmodule Aecore.Txs.Pool.Worker do
...
+      split_blocks(blocks, address, new_txs)
+    end
+
+  end
+
+  defp split_blocks([], _address, txs) do
+    txs
+  end
+
+  defp check_address_tx([tx | txs], address, user_txs) do
```
untra: @spec check_address_tx(list(%SignedTx{}, String().t, list(%SignedTx{})) :: list(%SignedTx{})

apps/aecore/lib/aecore/txs/pool/worker.ex
thepiwo: Internally we already discussed that the code structure could potentially be optimized but come to an agreement that its good enough for now

apps/aecore/lib/aecore/txs/pool/worker.ex
```diff
@@ -78,4 +115,83 @@ defmodule Aecore.Txs.Pool.Worker do
...
+  @doc """
+  A function that adds a merkle proof for every single transaction
+  """
+
+  @spec add_proof_to_txs(list()) :: list()
+  def add_proof_to_txs(user_txs) do
+    for tx <- user_txs do
+      block = get_block_by_txs_hash(tx.txs_hash)
+      tree  = BlockValidation.build_merkle_tree(block.txs)
+      key   = :erlang.term_to_binary(
```
cytadela8: `:erlang.term_to_binary` should also be part `TxData.hash_tx()`.  If you would run dialyzer you would see that the spec of `TxData.hash_tx()` (which is a spec for correct implementation) is different then the actual implementation.

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -153,11 +162,11 @@ defmodule Aecore.Chain.BlockValidation do
     block.txs
     |> Enum.map(
          fn tx -> cond do
-                    SignedTx.is_coinbase(tx) -> tx.data.value
-                    true -> 0
+             SignedTx.is_coinbase(tx) -> tx.data.value
+             true -> 0
                   end
```
cytadela8: This `end` is unaligned now

apps/aecore/lib/aecore/structures/tx_data.ex
```diff
@@ -11,7 +11,7 @@ defmodule Aecore.Structures.TxData do
   Definition of Aecore TxData structure
 
   ## Parameters
-     - nonce: A random integer generated on initialisation of a transaction.Must be unique
+  - nonce: A random integer generated on initialisation of a transaction.Must be unique
```
cytadela8: this is unaligned now

apps/aecore/lib/aecore/txs/pool/worker.ex
```diff
@@ -78,4 +115,83 @@ defmodule Aecore.Txs.Pool.Worker do
...
   end
 
+  @doc """
+  A function that adds a merkle proof for every single transaction
+  """
+
+  @spec add_proof_to_txs(list()) :: list()
+  def add_proof_to_txs(user_txs) do
+    for tx <- user_txs do
+      block = get_block_by_txs_hash(tx.txs_hash)
```
cytadela8: I just noticed you are using `get_block_by_txs_hash(tx.txs_hash)` when you can use the much faster `Chain.get_block(tx.block_hash)`

apps/aecore/test/aecore_get_txs_for_address.exs
```diff
@@ -0,0 +1,63 @@
...
+        |> Map.delete(:signature)
+        |> TxData.new()
+      )
+      key = TxData.hash_tx(transaction)
+      tx_block = Pool.get_block_by_txs_hash(user_tx_with_proof.txs_hash)
+      merkle_tree = BlockValidation.build_merkle_tree(tx_block.txs)
+      merkle_proof = :gb_merkle_trees.merkle_proof(key, merkle_tree)
+
+      assert {:ok, :verified} =
+        :gb_merkle_trees.verify_merkle_proof(key,
```
cytadela8: To properly test the `Pool.add_proof_to_txs` function you should do the same as potential user so:
for every tx in `user_txs_with_proof`:
1. calculate the transaction hash (now you have to also `Map.delete(:proof)`)
2. get the txs_hash from block header using the `:block_hash` key in tx
3. call verify_merkle_proof(hash, tx, txs_hash, tx.proof)

apps/aehttpserver/lib/aehttpserver/web/controllers/user_tx_controller.ex
```diff
@@ -0,0 +1,39 @@
...
+        case params["include_proof"]  do
+          "true" ->
+            proof = Pool.add_proof_to_txs(user_txs)
+            json(conn, Enum.map(proof, fn(tx) ->
+                  %{tx |
+                    from_acc: Serialization.hex_binary(tx.from_acc, :serialize),
+                    to_acc: Serialization.hex_binary(tx.to_acc, :serialize),
+                    txs_hash: Serialization.hex_binary(tx.txs_hash, :serialize),
+                    block_hash: Serialization.hex_binary(tx.block_hash, :serialize),
+                    signature: Serialization.hex_binary(tx.signature, :serialize)
```
cytadela8: Is the proof get serialized nicely?

HristianHristov: Yes, it is serialized in Pool.add_proof_to_txs

apps/aecore/lib/aecore/txs/pool/worker.ex
HristianHristov: The Chain.get_block() function works only with the prev_hash, so it wouldn't work well in this case.

HristianHristov:
@cytadela8 Can you review?

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -25,23 +25,23 @@ defmodule Aecore.Chain.BlockValidation do
       !(is_genesis || check_prev_hash(new_block, previous_block)) ->
         throw({:error, "Incorrect previous hash"})
 
-      # do not check previous block height for genesis block, there is none
-      !(is_genesis || check_correct_height(new_block, previous_block)) ->
+        # do not check previous block height for genesis block, there is none
```
cytadela8: Cases in this `cond do` are unaligned now.

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -57,10 +57,10 @@ defmodule Aecore.Chain.BlockValidation do
       !(validate_block_transactions(block) |> Enum.all?()) ->
         throw({:error, "One or more transactions not valid"})
 
-      coinbase_transactions_sum > Miner.coinbase_transaction_value() + total_fees ->
+        coinbase_transactions_sum > Miner.coinbase_transaction_value() + total_fees ->
```
cytadela8: In this `cond do` also.

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -107,7 +107,7 @@ defmodule Aecore.Chain.BlockValidation do
       chain_state_has_account ->
         tx.data.nonce > Map.get(chain_state, tx.data.from_acc).nonce
 
-      true ->
+        true ->
```
cytadela8: Hey, what is going on with your editor? Now you broke formatting here...

cytadela8:
Ohh, that's merged. Then we can ignore the minor formatting problem, but @HristianHristov please find the issue with your editor.

Issue #133, opened 2017-11-20
implement nakamoto consensus, follow the latest valid chain


thepiwo:
- [x] follow the longest valid chain
- [x] if there are 2 valid blocks at the same height, wait for the following block
- [x] if there is a chain discovered that can get validated, replace the current one with it

can be started after #97

cytadela8:
>  if there are 2 valid blocks at the same height, wait for the following block
Shouldn't we maybe compare them by hash? (lower is better?) We have to choose something if we would like to mine.

thepiwo:
no, this would be a "most work" kind of consensus, as the miner we choose the one we received earlier to maximize possible profit. We will research other consensus mechanisms in the future, if you have any ideas we can discuss them and build a prototype to validate.

cytadela8:
ok, I understand now.

thepiwo:
closed by #162

PR #162, opened 2017-12-14
[GH-133] Nakamoto consensus (longest chain consensus)

cytadela8:
This PR implements the longest chain consensus. To achieve that:
- I refactored `Aecore.Chain.Worker`
- we keep all the chain states for all of the blocks.
- we keep track of the highest block
- I changed relevant behaviour of other sync


cytadela8:
Trigger Travis

PR #174, opened 2017-12-18
[GH-#173] fixed issue with getting /info not as node

### LINKED ISSUE #173: GET /info API function does not work

thepiwo:
closes #173

### DIFF
--- a/apps/aehttpserver/lib/aehttpserver/web/controllers/info_controller.ex
+++ b/apps/aehttpserver/lib/aehttpserver/web/controllers/info_controller.ex
@@ -26,15 +26,21 @@ defmodule Aehttpserver.Web.InfoController do
     pubkey = Base.encode16(pubkey)
 
     #Add whoever's getting our info
-    peer_ip = conn.peer |> elem(0) |> Tuple.to_list |> Enum.join(".")
-    peer_port = Plug.Conn.get_req_header(conn, "peer_port") |> Enum.at(0) |> to_string()
-    peer_port = ":" <> peer_port
-    peer_nonce = Plug.Conn.get_req_header(conn, "nonce") |> Enum.at(0) |> String.to_integer()
-    peer = peer_ip <> peer_port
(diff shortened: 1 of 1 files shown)


apps/aehttpserver/lib/aehttpserver/web/controllers/info_controller.ex
```diff
@@ -26,16 +26,22 @@ defmodule Aehttpserver.Web.InfoController do
...
-
-    if(!(peer_nonce == own_nonce)) do
-      Peers.schedule_add_peer(peer, peer_nonce)
+    peer_port = Plug.Conn.get_req_header(conn, "peer_port")
+    peer_nonce = Plug.Conn.get_req_header(conn, "nonce")
+    if !Enum.empty?(peer_port) && !Enum.empty?(peer_nonce) do
+      peer_ip = conn.peer |> elem(0) |> Tuple.to_list |> Enum.join(".")
+      peer_port = peer_port   |> Enum.at(0) |> to_string()
+      peer_port = ":" <> peer_port
+      peer_nonce = peer_nonce |> Enum.at(0) |> String.to_integer()
```
untra: when `peer_port` was first declared, it was typed as a `[binary]`. see: https://hexdocs.pm/plug/Plug.Conn.html#get_req_header/2
When you reassign it here, it becomes an `integer`. This is how bugs get made unintentionally.

In general, we should try to avoid reassigning variables unless we have to, and even then they should only be reassigned if they are going to maintain the original datatype. Remember Elixir is dynamically typed, so any mistakes made like this will not be caught by the compiler. Theres no penalty for declaring a new local variable.

untra: same goes for `peer_port`, originally a `[binary]`, now a `string`.

thepiwo: yes, I knew about it and it was kind of unintentionally as the code was like this prior to me changing it, but I guess we should be more strict about this in general, thank you!

PR #175, opened 2017-12-18
[GH-165] Better chain_state validation

### LINKED ISSUE #165: Nonce validation
Currently the requirement of nonce always increasing isn't validated. It's only checked in pool.

cytadela8:
Instead of calculating chain state of a block and then combining it with
old chain state, we take old chain_state and apply transactions to it.
While applying transactions we do all required checks. One of them is
missing previously nonce check.

Overall the code is simplified and redundant calculations are removed.

This fixes #165


cytadela8:
@thepiwo How should be coinbase transaction validation handled? I think we should validate the lock time.

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -119,66 +113,36 @@ defmodule Aecore.Chain.ChainState do
...
-        !Map.has_key?(block_state, account) ->
-          Map.put(block_state, account, %{balance: 0, nonce: 0, locked: []})
+  @spec update_chain_state!(map(), binary(), integer(), integer(), integer(), boolean()) :: map()
+  defp update_chain_state!(chain_state, account, value, nonce, lock_time_block, add_to_locked) do
+    account_state = Map.get(chain_state, account, %{balance: 0, nonce: 0, locked: []})
 
-        true ->
-          block_state
-      end
+    if account_state.nonce >= nonce and nonce != -1 do
```
untra: When would the nonce specifically equal `-1`?

thepiwo:
@untra do you know what the elixir default is for type specifications? Right now we are mixing `%Block{}` and `Block.block()`, is maybe `Block.t()` the best way?

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -76,17 +76,26 @@ defmodule Aecore.Chain.BlockValidation do
...
   @spec validate_block_transactions(Block.block()) :: list()
   def validate_block_transactions(block) do
-    block.txs |> Enum.map(fn tx -> SignedTx.is_coinbase(tx) ||  SignedTx.is_valid(tx) end)
+    block.txs
+    |> Enum.map(
+         fn tx -> cond do
+                    SignedTx.is_coinbase(tx) ->
+                      true
+                    true ->
+                      KeyManager.verify(tx.data, tx.signature, tx.data.from_acc)
```
thepiwo: why did you revert this to the old validation check?

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -1,50 +1,50 @@
...
+  def apply_transaction_on_state!(transaction, chain_state, block_height) do
+    add_to_locked = block_height <= transaction.data.lock_time_block
+    cond do
+      SignedTx.is_coinbase(transaction) ->
+        update_chain_state!(chain_state, transaction.data.to_acc, 
+                            transaction.data.value, -1,
+                            transaction.data.lock_time_block,
+                            add_to_locked)
+      transaction.data.from_acc != nil ->
+        if transaction.data.value < 0 do
```
thepiwo: we should have this in the transaction validation

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -1,50 +1,50 @@
...
+          throw {:error, "Negative transaction value"}
+        end 
+        if transaction.data.fee < 0 do #TODO: enforce minimal fee
+          throw {:error, "Nefative transaction fee"}
```
thepiwo: we should have this in the transaction validation

apps/aecore/lib/aecore/chain/block_validation.ex
thepiwo: we should use `SignedTx.is_valid(tx)` which should include non negative value and non negative fee checks

thepiwo:
will this still work for the following case?

account `A` has 100 tokens
spends 80 to `B`
spends 60 to `C`

the expected result would be that either one of the transactions is included (higher fee/earlier, or something) in the block candidate and is actually mined, the other one should be discarded.

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -1,50 +1,50 @@
...
-
-        update_block_state(updated_block_state, transaction.data.to_acc, transaction.data.value,
-                           0, transaction.data.lock_time_block, add_to_locked)
-      end
-
-    reduce_map_list(block_state)
+  @spec calculate_and_validate_chain_state!(list(), map(), integer()) :: map()
+  def calculate_and_validate_chain_state!(txs, chain_state, block_height) do
+    Enum.reduce(txs, chain_state, fn(transaction, chain_state) ->
+      apply_transaction_on_state!(transaction, chain_state, block_height)
```
d-velev: Shouldn't we try-catch here as well?

untra:
@thepiwo 
> @untra do you know what the elixir default is for type specifications? Right now we are mixing `%Block{}` and `Block.block()`, is maybe `Block.t()` the best way?

Sorry, just saw this. There's no default way to name your types but Indeed, `Block.t()` is the most common accepted pattern, usually written as just `Block.t`. This follows the pattern for other elixir types, such as `Enum.t`, `Range.t`, etc.

You can name your types anything, technically. But then you have to refer to it as `Block.anythingT`
```
@type anythingT :: %Block{...}
```
see https://elixirschool.com/en/lessons/advanced/typespec/#defining-custom-type
https://hexdocs.pm/elixir/typespecs.html#types-and-their-syntax

thepiwo:
Thank you very much! I guess we'll make another task to refactor this.

thepiwo:
@cheezus1 @nedelcho-delchev-tues @meivantodorov can you review this as well?

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -1,50 +1,50 @@
...
-  to the current state, returns a map with the
-  accounts as key and their balance as value.
-  """
-  @spec calculate_chain_state(map(), map()) :: map()
-  def calculate_chain_state(block_state, chain_state) do
-    merge_states(block_state, chain_state)
+  @spec apply_transaction_on_state!(SignedTx.signed_tx(), map(), integer()) :: map()
+  def apply_transaction_on_state!(transaction, chain_state, block_height) do
+    add_to_locked = block_height <= transaction.data.lock_time_block
+    cond do
```
d-velev: if instead of cond seems more appropriate

apps/aecore/lib/aecore/chain/chain_state.ex
cytadela8: no, any exception in `apply_transaction_on_state!` should be handled by function calling `calculate_and_validate_chain_state!`

apps/aecore/lib/aecore/chain/chain_state.ex
cytadela8: There are 3 cases and in future we will add another for every new transaction type so I would leave `cond`.

cytadela8:
> will this still work for the following case?
> 
> account A has 100 tokens
> spends 80 to B
> spends 60 to C
> 
> the expected result would be that either one of the transactions is included (higher fee/earlier, or something) in the block candidate and is actually mined, the other one should be discarded.

@thepiwo: Yes, the `filter_invalid_transactions_chainstate` function is used by miner to filter invalid transactions. It uses supplied chainstate (top block chainstate) and applies to it all of the transactions it gets in order. If a transaction fails to apply (in your example the second transaction of A will fail with "balance < 0") the transaction is filtered-out.

cytadela8:
> will this still work for the following case?
> 
> account A has 100 tokens
> spends 80 to B
> spends 60 to C
> 
> the expected result would be that either one of the transactions is included (higher fee/earlier, or something) in the block candidate and is actually mined, the other one should be discarded.

@thepiwo: Yes, the `filter_invalid_transactions_chainstate` function is used by miner to filter invalid transactions. It uses supplied chainstate (top block chainstate) and applies to it all of the transactions it gets in order. If a transaction fails to apply (in your example the second transaction of A will fail with "balance < 0") the transaction is filtered-out.

thepiwo:
@meivantodorov @cheezus1 @nedelcho-delchev-tues can you review this?

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -8,7 +8,7 @@ defmodule Aecore.Chain.BlockValidation do
   alias Aecore.Chain.ChainState
   alias Aecore.Chain.Difficulty
 
-  @spec calculate_and_validate_block!(Block.t(), Block.t(), map(), list(Block.t())) :: {:error, term()} | :ok
+  @spec calculate_and_validate_block!(Block.block(), Block.block(), map(), list(Block.block())) :: {:error, term()} | :ok
```
thepiwo: why? we prefer `.t()` notation

cytadela8: The `.t()` notation (the missing `-type` in `Aecore.Structures.Block`) is introduced by an unmerged PR (#182). We can merge that one first and then revert this commit.

Issue #171, opened 2017-12-18
Dev environment releases not working 


nedelcho-delchev-tues:
The dev releases are not working right now due to major changes. You can use **multi_node_screen.sh** for now

I will wait for all the major PRs to be merged before fixing it.

untra:
I was also having trouble getting the the dev build to work over the weekend. My problems were around rustler and compiling rox, however. Is that currently the issue?

The `multi_node_screen.sh` worked to make an effective release, but I wasn't sure how to use it further after staring up the release with the `console` command.

nedelcho-delchev-tues:
Yes, the issue is about rox not compiling and the application can't start properly I suppose. I will fix it in the coming days.

PR #176, opened 2017-12-19
[GH-144] Fixing Cuckoo

### LINKED ISSUE #144: fix Cuckoo Cycle crashes

meivantodorov:
Closes #144 

- NiF was replaced with port.
- Major changes on the miner worker.
- Some unit tests were refactored.


untra:
> * NiF was replaced with port.
I think that is a smart call, as Erlang/otp tends to play better with data interacting through ports rather than calling native functions in different languages.

So then does this mean the cuckoo cycle runs as a separate service? If I have a cuda enabled gpu, will I be able to run the [cuda-enabled cuckoo cycle c code](https://github.com/tromp/cuckoo/blob/master/src/cuda_miner.cu) instead of the current .cpp NiF?

meivantodorov:
@untra 
We start separately the miner from the vm as os process and we communicate with it via port.
As for the using gpu - at some point yes, for now we have not implemented it.

thepiwo:
can you fix this in this branch? https://github.com/aeternity/elixir-research/issues/179 `{:error, _}` should do it

thepiwo:
@cheezus1 @nedelcho-delchev-tues @d-velev @cytadela8 can you review this as well?

thepiwo:
@meivantodorov ok merge this now?

nedelcho-delchev-tues:
fixes #171

meivantodorov:
@thepiwo I think it is okey to be merged, if something come up we can mange it later on.

thepiwo:
ok @meivantodorov please merge :tada:

PR #177, opened 2017-12-19
[GH-#173] don't reuse variable names, addressed @untra comment

thepiwo:
addresses #174 suggestion from @untra


untra:
This PR definitely helps those small issues. Its a start. 👍  If you want to prevent variable rebinding for good, I would recommend enabling this credo check and then fixing all the errors:
https://github.com/aeternity/elixir-research/blob/master/.credo.exs#L129

Issue #160, opened 2017-12-07
Dialyzer


cytadela8:
- [ ] add `@spec` to functions missing it
- [ ] fix dialyzer errors
- [ ] ~~add dialyzer to Travis~~ dialyzer for elixir is not ready to be used as part of CI.
- [x] add dialyzer to mix

untra:
>  add dialyzer to Travis

I would actually advice against this at the moment; unfortunately dialyzer and elixir are not fully copacetic, and there are a lot of reported false positives (`The created fun has no local return`). Dializer is also currently convinced certain functions consuming `SignedTx.t()` should instead have success typings consuming `atom()` instead.

Dialyzer is still a good tool to identify serious mistakes, and even more mistakes are prevented when public methods get `@spec`'d. But I think there is more work involved before Dialyzer is ready to run as part of CI. :confused:

cytadela8:
@untra Ok, I removed that from checklist.

Issue #161, opened 2017-12-07
Fix credo


cytadela8:
There are some credo warnings. We should fix them.

PR #180, opened 2017-12-29
[GH-160] Dialyxir and Credo fixes

untra:
* enforces a `@type t` typespec for `Block`, `Header`, `TxData`, etc. ensuring consistancy
* removes all parens from types (so `integer()` becomes just `integer`, etc.)
* fixes more credo errors
* brings in latest changes from master


untra:
:exclamation: **Note:** These changes are being added to the Aeternity GH-160 branch. It doesn't complete #160 or #161, but I wanted to continue the work that was started on that front.

apps/aecore/lib/aecore/structures/block.ex
```diff
@@ -5,15 +5,15 @@ defmodule Aecore.Structures.Block do
   alias Aecore.Structures.Block
   alias Aecore.Structures.Header
 
-  @type block :: %Block{}
+  @type t :: %Block{}
```
untra: **Note:** that there is no _correct_ way to typespec elixir structs: `%Block{}` is as acceptable as `Block.t`, its just a matter of team preference and what gets the job done. I can easily revert this change if desired.

:question: Which style does the team prefer; `%Block{}` or `Block.t`?

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -100,34 +99,31 @@ defmodule Aecore.Chain.BlockValidation do
...
+      to_acc_new_state = %{balance: tx.data.value, nonce: 0, locked: []}
+      chain_state_changes = %{tx.data.from_acc => from_acc_new_state, tx.data.to_acc => to_acc_new_state}
+      updated_chain_state = ChainState.calculate_chain_state(chain_state_changes, chain_state)
+      {true, updated_chain_state}
+    else
+      {false, chain_state}
     end
   end
 
-  @spec calculate_root_hash(list()) :: binary()
```
untra: **Note:** erlang requires parens for its types, but elixir has them optional. I opted for the elixir style typenames without parens, but this should be a team choice. I can easily revert this change if desired.

:question: Which style does the team prefer; erlang style with parens (eg `binary()`) or elixir style without parens (eg `binary`)?

apps/aecore/lib/aecore/structures/block.ex
d-velev: %Block{} looks better imo

apps/aecore/lib/aecore/chain/block_validation.ex
d-velev: We've used the erlang style so far so let's keep it like that

apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -4,9 +4,13 @@ defmodule Aecore.Structures.SignedTx do
   """
 
   alias Aecore.Keys.Worker, as: Keys
+  alias Aecore.Structures.TxData
   alias Aecore.Structures.SignedTx
 
-  @type signed_tx() :: %SignedTx{}
+  @type t :: %SignedTx{} | %SignedTx{
+    data: %TxData{},
+    signature: binary()
```
untra: The signature is supposed to be the "Signed %TxData{} with the private key of the sender". Whats the correct typespec for that?

thepiwo:
Thank you! Will you continue work on this or should we take over?

thepiwo:
I guess additionally we should refactor our `list()` types to more specific ones, e.g. `list(Block.t())`

apps/aecore/lib/aecore/structures/signed_tx.ex
thepiwo: Ideally it should be

```
@type t :: %SignedTx{
    data: TxData.t(),
    signature: binary()
}
```

correct?

thepiwo: The `SignedTx` represents the envelop for the `TxData` including the binary signature blob

untra:
@thepiwo I'll keep working on it. Theres a few more things I want to get typespec'd and refactored, and I'll look to make the list types more specific, as mentioned above. When it's getting really ready I'll make a proper PR against the master branch

apps/aehttpclient/lib/client.ex
```diff
@@ -115,16 +130,11 @@ defmodule Aehttpclient.Client do
...
-  end
-
   defp send_to_peer(data, uri) do
-    HTTPoison.post uri, Poison.encode!(data),
-      [{"Content-Type", "application/json"}]
+    HTTPoison.post(uri, Poison.encode!(data), [{"Content-Type", "application/json"}])
   end
 
+  # TODO: what is this function even doing?
   defp get_local_port() do
```
untra: What is going on in this function? The code here seems ridiculous, and either needs a comment explaining its operations, or can be refactored to pattern match against array internals

thepiwo: Its a little weird, we need to get the port we are listening on

PR #184, opened 2018-01-08
[GH-#181] Adding peers leads to timeout with big blockchain

### LINKED ISSUE #181: adding peers leads to timeout with big blockchain
adding a peer leads to various functions in the peers and sync worker to timeout if nodes will not answer fast or the downloading of blocks takes a longer time, in case the blockchain already has a lot of blocks to sync

d-velev:
closes #181, #193


apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -32,14 +32,20 @@ defmodule Aecore.Peers.Sync do
...
-    GenServer.call(__MODULE__, {:ask_peers_for_unknown_blocks, peers})
+    GenServer.cast(__MODULE__, {:ask_peers_for_unknown_blocks, peers})
   end
 
-  @spec add_valid_peer_blocks_to_chain() :: :ok
   def add_valid_peer_blocks_to_chain() do
     GenServer.call(__MODULE__, :add_valid_peer_blocks_to_chain)
   end
 
+  def add_valid_peer_blocks_to_chain(state) do
```
thepiwo: the spec is missing now

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -32,14 +32,20 @@ defmodule Aecore.Peers.Sync do
 
   @spec ask_peers_for_unknown_blocks(map()) :: :ok
   def ask_peers_for_unknown_blocks(peers) do
-    GenServer.call(__MODULE__, {:ask_peers_for_unknown_blocks, peers})
+    GenServer.cast(__MODULE__, {:ask_peers_for_unknown_blocks, peers})
   end
 
-  @spec add_valid_peer_blocks_to_chain() :: :ok
```
thepiwo: the spec is missing now

apps/aecore/lib/aecore/pow/cuckoo.ex
```diff
@@ -135,7 +135,7 @@ defmodule Aecore.Pow.Cuckoo do
         handle_raw_data(process, buff)
       any ->
         Logger.error("[Cuckoo] Unexpeted error : #{inspect(any)}")
-        exit(:kill)
+        wait_for_result(process, buff)
```
thepiwo: why is this changed?

apps/aecore/lib/aecore/peers/scheduler.ex
```diff
@@ -19,9 +19,11 @@ defmodule Aecore.Peers.Scheduler do
     Peers.check_peers()
     Sync.introduce_variety()
     Sync.refill()
-    Sync.ask_peers_for_unknown_blocks(Peers.all_peers())
-    Sync.add_valid_peer_blocks_to_chain()
+    if(Peers.is_chain_synced?()) do
```
thepiwo: why the if?

d-velev: This way we don't queue any unnecessary casts (only called if we have finished syncing)

apps/aecore/lib/aecore/pow/cuckoo.ex
d-velev: Ivan suggested this as a workaround, we seem to get an unexpected message here.

thepiwo: ok great

apps/aecore/lib/aecore/peers/scheduler.ex
thepiwo: Ok, I didn't notice

cytadela8: `Peers.is_chain_synced?()` returns false if our top block height is lower then the one reported by peers. `ask_peers_for_unknown_blocks` download's the blocks we don't know. So now we will never get our chain synced as `ask_peers_for_unknown_blocks` that's exactly what `ask_peers_for_unknown_blocks` does.

d-velev: You are right, the thing I had in mind was to check if a sync is in progress, maybe we can have a status variable for the sync module and check that instead.

thepiwo:
@meivantodorov @d-velev are there still issues with this?

@cytadela8 can you re-review?

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -15,7 +15,17 @@ defmodule Aecore.Chain.BlockValidation do
...
+    server = self()
+    work   = fn() -> Cuckoo.verify(new_block.header) end
+    spawn(fn() ->
+      send(server, {:worker_reply, self(), work.()})
+    end)
+
+    is_difficulty_target_met =
+      receive do
+      {:worker_reply, _from, verified?} -> verified?
+    end
```
cytadela8: What's the point of adding this worker? We just execute Cuckoo.verify and immediately go to "receive do" where we wait for it to finish? How is this better then before?

cytadela8:
@thepiwo For me it looks like my previous comment/"major problem" wasn't fixed. @meivantodorov, @d-velev if you need any help or have any questions don't hesitate to write to me.

thepiwo:
@cytadela8 can you describe the major problem again? I don't understand

cytadela8:
@thepiwo `Sync.ask_peers_for_unknown_blocks(Peers.all_peers())` is the function responsible for downloading new blocks and adding them to chain. `Peers.is_chain_synced?()` returns **false** when other peers report having higher height of their top block, so they have blocks we don't have. *This is not important, but previously there where two functions responsible for the block sync process `Sync.ask_peers_for_unknown_blocks(Peers.all_peers())` and `Sync.add_valid_peer_blocks_to_chain()` this PR merges them.*

By changing the code to:
```
if(Peers.is_chain_synced?()) do
    Sync.ask_peers_for_unknown_blocks(Peers.all_peers())
end
```
We only sync blocks when there is nothing to sync. So basically we don't sync blocks. There might be some "corner case" when `Peers.is_chain_synced?()` returns true despite there being nothing to sync, but that's not the general case. So the new code works like if the `Sync.ask_peers_for_unknown_blocks(Peers.all_peers())` was completely removed.

It is important to note (and I forgot about that previously). That it might **look like** everything is working, because we have one more mechanism of block exchange. We have the "gossip" mechanism of broadcasting new found block to other peers.

P.S. Sorry for late response.

apps/aecore/lib/aecore/chain/block_validation.ex
cytadela8: @d-velev please respond to my comment.

d-velev: @meivantodorov worked on this, he should be able to explain better

d-velev:
@cytadela8, @thepiwo can you review? I did #193 as well, the nodes that we have running have old code so I can't test with a long chain but we should be receiving new blocks while downloading (the Sync GenSever isn't busy with syncing anymore)

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -184,63 +281,31 @@ defmodule Aecore.Peers.Sync do
...
     has_parent_in_chain = Chain.has_block?(block.header.prev_hash)
-    cond do
-      has_parent_block_in_state ->
-        build_chain(state, state[block.header.prev_hash], [block | chain])
-      has_parent_in_chain ->
-        [block | chain]
-      true ->
-        []
+    block_header_hash = BlockValidation.block_header_hash(block.header)
+    if(!Chain.has_block?(block_header_hash)) do
```
cytadela8: Please remove parenthesis around if. (To find such errors in future you can use `mix credo -a` it should point this out.)

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -1,44 +1,121 @@
...
+            Enum.each(blocks, fn(block) ->
+              try do
+                BlockValidation.single_validate_block!(block)
+                peer_block_hash = BlockValidation.block_header_hash(block.header)
+                if !Chain.has_block?(peer_block_hash) do
+                  add_block_to_state(peer_block_hash, block)
+                end
+              catch
+                {:error, message} ->
+                  Logger.error(fn -> message end)
```
cytadela8: When a sync with some peer fails and `Logger.error` is called `remove_running_task(peer_uri)` won't get called. This will cause the peer to be excluded from every future block sync, because `Map.has_key?(get_running_tasks(), uri)` will always be true.

apps/aecore/lib/aecore/chain/block_validation.ex
thepiwo: @meivantodorov can you explain?

d-velev:
I think that's the right place for it, the task doesn't stop when we reach an invalid block and it shouldn't, if we reach an invalid one we don't add it to our state and continue.

cytadela8:
@d-velev My bad, you are correct.

thepiwo:
@cytadela8 so are we good to go?

cytadela8:
@thepiwo We are good to go, but I would like to get the response from @meivantodorov. I will try reaching him on Slack.

apps/aecore/lib/aecore/chain/block_validation.ex
meivantodorov: Our cuckoo has own receive block and must be started as a worker. Otherwise it receives the parent's messages.

PR #188, opened 2018-01-12
[GH-#143] Chain state persistence to disk using rocksdb

### LINKED ISSUE #143: use rocksdb to read/write chainstate by address/pubkey

meivantodorov:
### What this PR does: 
- Write to disk each account state on `RockDB`separately in order to write on chunks instead of writing
the whole chain state every time.
- Provided write options
- Batch write.
- Write to disk block info in order to get easily `top_hash` and `top_height` on start-up stage.
- Retrieving all blocks as map.
- Retrieving all accounts states as map.
- Retrieving latest block info
- Loading previous persisted states on start-up.

Closes #143


apps/aecore/lib/aecore/persistence/worker.ex
```diff
@@ -30,27 +31,56 @@ defmodule Aecore.Persistence.Worker do
...
   end
 
   def handle_call({:write_block_by_hash, {hash, block}}, _from,
     %{db: _db, blocks_family: blocks_family} = state) do
     {:reply, Rox.put(blocks_family, hash, block), state}
   end
 
+  def handle_call({:write_chain_state_by_pubkey, {pubkey, chain_state}}, _from,
+    %{db: _db, chain_state_family: chain_state_family} = state) do
+    {:reply, Rox.put(chain_state_family, pubkey, chain_state), state}
```
thepiwo: does this write every public key as seperate entry in the db?

meivantodorov: yes

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -161,8 +161,13 @@ defmodule Aecore.Chain.Worker do
     Logger.info(fn ->
       "Added block ##{new_block.header.height} with hash #{Base.encode16(new_block_hash)}, total tokens: #{inspect(total_tokens)}"
     end)
+
     ## Store new block to disk
     Persistence.write_block_by_hash(new_block)
+
+    ## Store new chain_state to disk
+    Persistence.write_chain_state_by_pubkey(updated_chain_states)
```
thepiwo: it would be more efficient to update the changed chainstates (calculated earlier) in the database, instead of a full rewrite everytime

thepiwo:
Is there a way to have a typed map() in elixir?

apps/aecore/lib/aecore/persistence/worker.ex
thepiwo: great!

cytadela8:
@thepiwo 
According to https://hexdocs.pm/elixir/typespecs.html, (I think) it should be done with `%{optional(key_type) => value_type}   # map with optional pairs of key_type and value_type`

apps/aecore/lib/aecore/chain/worker.ex
thepiwo: right now you write the full chainstate for each block to disk, we should find a way to only write the changes affected by this block

cytadela8: We would have to probably use some special database with some kind of versioning.

thepiwo: we don't need versioning for our blockchain implementation. Additional indexes for better APIs should be implemented in external programms

cytadela8: We could use SQL probably and save changes as new rows, but this will be probably slow unless configured properly and configuring can be impossible.

cytadela8: We can also Implement our own versioning by generating saving only differences. Then to read data back one would have to apply the changes since the first block.

cytadela8: hmm...

thepiwo: It would be very slow and inefficient, the plan is to use rocksdb for the current state and think about versioning later.

cytadela8: So I think there are two separate problems:
1. We are storing all states (for all blocks) in memory currently. This can be solved in future by just deleting data in memory.
2. When we want to store information in rox we have to write the whole state not just changes.

cytadela8: So the question to @meivantodorov is: How can we split the chain_state for rox so only the updates are written to disk.

thepiwo: Currently we have:
 - all blocks in memory
 - the latest chainstate in memory
 - the blocks downloaded by sync in memory

Ideas to persist or swap:
 - save everything instead of last 10/100 (difficulty adjustment range) as `hash -> block` to rocksdb
 - save each account state `account -> state` to rocksdb and update if this account changes in one block
 - download all blocks for sync to disk, load them from 1 to top to validate

meivantodorov:
@thepiwo, @cytadela8, @nedelcho-delchev-tues  can you review this PR ? Thanks.

apps/aecore/lib/aecore/persistence/worker.ex
```diff
@@ -7,50 +7,124 @@ defmodule Aecore.Persistence.Worker do
...
+  @spec add_block_by_hash(Block.t()) :: :ok | {:error, reason :: term()}
+  def add_block_by_hash(%{header: header} = block) do
     hash = BlockValidation.block_header_hash(header)
-    GenServer.call(__MODULE__, {:write_block_by_hash, {hash, block}})
+    GenServer.call(__MODULE__, {:add_block_by_hash, {hash, block}})
+  end
+  def add_block_by_hash(_block), do: {:error, "bad block structure"}
+
+  @spec get_block_by_hash(String.t()) ::
+  {:ok, block :: Block.t()} | :not_found | {:error, reason :: term()}
```
thepiwo: can you put the spec in one line, this is like this multiple times

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -189,6 +200,32 @@ defmodule Aecore.Chain.Worker do
     {:reply, txs_index, state}
   end
 
+  def handle_info(:timeout, state) do
```
thepiwo: so this will timeout with a bigger blockchain, as everything is loaded in memory, but we can keep it for now and improve with a later task

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -164,8 +168,15 @@ defmodule Aecore.Chain.Worker do
       "Added block ##{new_block.header.height} with hash #{Base.encode16(new_block_hash)}, total tokens: #{inspect(total_tokens)}"
     end)
 
-    ## Store new block to disk
-    Persistence.write_block_by_hash(new_block)
+    ## Store new block to RocksDB
```
thepiwo: can you combine all persistence to happen for a added block in one function inside the persistence worker? 
inside you can have the L171 - L178 as it is now.

thepiwo:
@meivantodorov I propose a later task to fetch blocks lazy and only keep the needed data in RAM.

apps/aecore/lib/aecore/persistence/worker.ex
```diff
@@ -7,50 +7,124 @@ defmodule Aecore.Persistence.Worker do
...
+            chain_state_family: chain_state_family}}
   end
 
-  def handle_call({:write_block_by_hash, {hash, block}}, _from,
-    %{db: _db, blocks_family: blocks_family} = state) do
-    {:reply, Rox.put(blocks_family, hash, block), state}
+  def handle_call({:add_block_by_hash, {hash, block}}, _from,
+    %{db: _db, latest_block_family: latest_block_family,
+      blocks_family: blocks_family} = state) do
+    :ok = Rox.put(latest_block_family, @latest_block_key,
```
cytadela8: I'm not sure about the purpose of the "letest_block", I guess the idea is to store the top block separately so we can easily find it (without checking all blocks). If so then this put should be corrected (height check should be added), because theoretically we can learn about some shorter fork after learning the current top block.

apps/aecore/lib/aecore/persistence/worker.ex
meivantodorov: I'm trying to keep my lines no longer than 80 columns.

meivantodorov:
@thepiwo According this [FAQs](https://github.com/facebook/rocksdb/wiki/RocksDB-FAQ) I can enable WAL and sync write in order to prevent data loses. The elixir library looks like support those writing options

apps/aecore/lib/aecore/persistence/worker.ex
meivantodorov: Thank you, for your comment. I will consider this.

apps/aecore/lib/aecore/chain/worker.ex
meivantodorov: I will consider this, and will let you know, thanks

thepiwo:
Can you try writing everything concerning one block in one transactions, so either everything is written to disk, otherwise the whole block changes, including chainstate are not persisted

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -164,12 +168,17 @@ defmodule Aecore.Chain.Worker do
...
                               chain_states: updated_chain_states,
                               txs_index: new_txs_index}
     if top_height < new_block.header.height do
+      ## Store to RocksDB latest chain_state, latest block and latest block info
+      Persistence.batch_write(%{:chain_state => new_chain_state,
+                                :block => %{new_block_hash => new_block},
+                                :latest_block_info => %{"top_hash" => new_block_hash,
+                                                        "top_height" => new_block.header.height}})
+
+
```
cytadela8: You should also write blocks to disk when `top_height >= new_block.header.height` (in the "else"), because this means we are just adding some fork and it might be longer, but the current block is "lower" then our top.

cytadela8: fixed

meivantodorov: @thepiwo @cytadela8 is there anything else you want to add, that I'm missing ? 
@nedelcho-delchev-tues @d-velev can you review as well ? Thanks.

apps/aecore/lib/aecore/persistence/worker.ex
```diff
@@ -6,51 +6,177 @@ defmodule Aecore.Persistence.Worker do
...
 
   ## Server side
 
   def init(_) do
-    ## We are creating `blocks_by_hash` family for the blocks
+    ## We are ensuring that families for the blocks and chain state
+    ## are created. More about them -
     ## https://github.com/facebook/rocksdb/wiki/Column-Families
-    {:ok, db, %{"blocks_by_hash" => blocks_family}} =
+    {:ok, db, %{"blocks_family"       => blocks_family,
```
nedelcho-delchev-tues: Maybe we should start aligning these(L88-L90) like so: 
```
    {:ok, db, %{"blocks_family"           => blocks_family,
                "latest_block_info_family" => latest_block_info_family,
                "chain_state_family"         => chain_state_family}} =
```

apps/aecore/lib/aecore/persistence/worker.ex
```diff
@@ -6,51 +6,177 @@ defmodule Aecore.Persistence.Worker do
 
   use GenServer
 
+  alias Rox.Batch
   alias Aecore.Chain.BlockValidation
+  alias Aecore.Keys.Worker, as: Keys
 
   require Logger
 
+  @latest_block_key "latest_block_info"
```
d-velev: What is this used for?

apps/aecore/lib/aecore/persistence/worker.ex
```diff
@@ -6,51 +6,177 @@ defmodule Aecore.Persistence.Worker do
...
+            fn({key, val}, batch_acc_) ->
+              Batch.put(batch_acc_, family, key, val)
+            end)
+        end)
+    Batch.write(batch, db)
+    {:reply, :ok, state}
+  end
+
+  def handle_call({:add_block_by_hash, {hash, block}}, _from,
     %{db: _db, blocks_family: blocks_family} = state) do
```
d-velev: Matching db shouldn't be needed if the value isn't used after that

apps/aecore/lib/aecore/persistence/worker.ex
meivantodorov: Not used, thanks.

thepiwo:
@cytadela8 can you re-review?

apps/aecore/test/aecore_persistence_test.exs
```diff
@@ -3,30 +3,63 @@ defmodule PersistenceTest do
...
 
-  setup do
+  setup persistance_state do
     Persistence.start_link([])
     Miner.start_link([])
-    []
+    Miner.mine_sync_block_to_chain
+    path = Application.get_env(:aecore, :persistence)[:path]
+    on_exit fn ->
+      File.rm_rf(path)
```
cytadela8: Hmm, that seems dangerous. I'm not sure what we can do, but some error in `:path` env variable can be catastrophic. Maybe we could check on previous line if the folder is empty and fail the test if it's not?

cytadela8: I would do: `assert false == File.exists?(path)`

thepiwo: I'd check if the path is kind of what we expect

cytadela8: My suggestion doesn't work. The path already exists when this test is invoked. If you won't have any idea how to fix this we can merge as is.

thepiwo: @meivantodorov any ideas?

meivantodorov: I will simply add `if do` and we can finish this

thepiwo:
@meivantodorov so, let's merge this?

PR #192, opened 2018-01-17
[GH-#186] Download blocks in batches

### LINKED ISSUE #186: download blocks in batches
Allow the sync to download blocks in batches
 - similar to `/blocks` endpoint but with the raw block structure, maybe as `/raw_blocks`
 - e.g. download 1000 consecutive blocks with one request

cheezus1:
Sync now downloads blocks in batches
closes #186


apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -208,30 +208,38 @@ defmodule Aecore.Peers.Sync do
...
   # Gets all unknown blocks, starting from the given one
-  defp check_peer_block(peer_uri, block_hash, state) do
-    case Chain.has_block?(block_hash) do
-      false ->
-        case(HttpClient.get_block({peer_uri, block_hash})) do
-          {:ok, deserialized_block} ->
+  defp check_peer_blocks(peer_uri, from_block_hash, state) do
+    case HttpClient.get_raw_blocks({peer_uri, from_block_hash, Chain.top_block_hash()}) do
+      {:ok, blocks} ->
+        if(Enum.count(blocks) > 0) do
```
d-velev: !Enum.empty?(blocks) should be better

apps/aehttpclient/lib/client.ex
```diff
@@ -28,6 +28,28 @@ defmodule Aehttpclient.Client do
...
   end
 
+  @spec get_raw_blocks({term(), binary(), binary()}) :: {:ok, term()} | {:error, binary()}
+  def get_raw_blocks({uri, from_block_hash, to_block_hash}) do
+    from_block_hash = Base.encode16(from_block_hash)
+    to_block_hash = Base.encode16(to_block_hash)
+    uri = uri <> "/raw_blocks?" <>
+            "from_block=" <> from_block_hash <>
+            "&to_block=" <> to_block_hash
+    case(get(uri, :raw_blocks)) do
```
d-velev: We can move the deserialization to the get function and call it here without making a case for what it returns, it does the error handling anyway.

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -212,17 +217,21 @@ defmodule Aecore.Chain.Worker do
...
         case(GenServer.call(__MODULE__, {:get_block, next_block_hash})) do
           {:error, _} -> blocks_acc
           block ->
-            updated_block_acc = [block | blocks_acc]
+            updated_blocks_acc = [block | blocks_acc]
             prev_block_hash = block.header.prev_hash
             next_size = size - 1
 
-            get_blocks(updated_block_acc, prev_block_hash, next_size)
+            if(next_block_hash != final_block_hash) do
```
cytadela8: Style: remove parenthesis inside `if ... do`
Suggestion: you can use `mix credo` to find such mistakes.

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -212,17 +217,21 @@ defmodule Aecore.Chain.Worker do
       end)
   end
 
-  defp get_blocks(blocks_acc, next_block_hash, size) do
+  defp get_blocks(blocks_acc, next_block_hash, final_block_hash, size) do
```
cytadela8: Style: you are calling the same number (max number of blocks to return) first `count` then (inside this function) `size`. Please make the naming consistent. *Renaming from `count` to `size` looks easier.*

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -208,30 +208,38 @@ defmodule Aecore.Peers.Sync do
...
               {:error, _} ->
-                state
+                state_acc
             end
-          :error ->
+          end)
+
+          earliest_block = Enum.at(blocks, Enum.count(blocks) - 1)
+          earliest_block_hash = BlockValidation.block_header_hash(earliest_block.header)
+          if(earliest_block_hash != Chain.top_block_hash()) do
```
cytadela8: What if we have `earliest_block_hash`, but it's not the top_block, because a fork happened?
You should use `Chain.has_block?` to do such checks.

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -208,30 +208,38 @@ defmodule Aecore.Peers.Sync do
   end
 
   # Gets all unknown blocks, starting from the given one
-  defp check_peer_block(peer_uri, block_hash, state) do
-    case Chain.has_block?(block_hash) do
-      false ->
-        case(HttpClient.get_block({peer_uri, block_hash})) do
-          {:ok, deserialized_block} ->
+  defp check_peer_blocks(peer_uri, from_block_hash, state) do
```
cytadela8: We should preserve the `Chain.has_block?(block_hash)` to be sure we don't download any block twice.

cheezus1:
I included the `to_block` in the response so that we make one less request (in some cases). If I exclude it, there can be a case where we can have an empty response. If you want, I will exclude it.

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -208,31 +208,39 @@ defmodule Aecore.Peers.Sync do
...
-              end
-            catch
-              {:error, _} ->
-                state
+        case HttpClient.get_raw_blocks({peer_uri, from_block_hash, Chain.top_block_hash()}) do
+          {:ok, blocks} ->
+            if !Enum.empty?(blocks) do
+              state = Enum.reduce(blocks, state, fn(block, state_acc) ->
+                try do
+                  BlockValidation.single_validate_block(block)
```
thepiwo: this is a complex function, can you make it less deep, by extracting another function?

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -208,31 +208,43 @@ defmodule Aecore.Peers.Sync do
...
+              state
             end
-          :error ->
+          {:error, _} ->
             state
         end
       true ->
         state
-    end
+      end
```
cytadela8: Style: misaligned `end`

cytadela8:
@meivantodorov, @andonov, @nedelcho-delchev-tues Could you review, please?

PR #196, opened 2018-01-20
erlang 20.2 and elixir 1.6.0 support

vans163:
This makes compilation work on erlang 20.2 (with enif_select support and C NIF API update) and elixir 1.6.0, I am not sure if other issues can arise.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -5,7 +5,7 @@ defmodule EpochElixir.Mixfile do
     [app: :epoch_elixir,
      apps_path: "apps",
      version: "0.1.0",
-     elixir: "~> 1.5.1",
+     elixir: "> 1.5.1",
      build_embedded: Mix.env == :prod,
      start_permanent: Mix.env == :prod,
      deps: deps(),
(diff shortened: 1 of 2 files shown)


mix.exs
```diff
@@ -5,7 +5,7 @@ defmodule EpochElixir.Mixfile do
     [app: :epoch_elixir,
      apps_path: "apps",
      version: "0.1.0",
-     elixir: "~> 1.5.1",
+     elixir: "> 1.5.1",
```
untra: This is not Elixir `1.6.0` support. My understanding is this says "only use a version of elixir greater than `1.5.1`". If you have 1.5.2 installed, it will use that. See: https://hexdocs.pm/elixir/Version.html

vans163: Yes correct it should be >=

cytadela8:
I think we should move to Elixir 1.6, but I'm not sure if moving now wouldn't be too fast. We can also apply this patch theoretically, but this would mean some people would be using 1.5 and some 1.6. I'm not sure how big of a risk of running into version dependent problems that is.

d-velev:
HTTPoison doesn't compile with 1.6.0

cytadela8:
I also remember experiencing problems with HTTPoison recenty when I tried compiling "elixir-research" under Elixir 1.6 recently.

thepiwo:
Let's stick with 1.5 for now, once our dependencies compile with 1.6 we switch, by explicit setting it

PR #198, opened 2018-01-24
[GH-#187] Limit block size

### LINKED ISSUE #187: limit block size
limit the size of blocks to 500kb by limiting the amount of transactions included from the pool

cheezus1:
closes #187


apps/aecore/lib/aecore/miner/worker.ex
```diff
@@ -299,6 +280,50 @@ defmodule Aecore.Miner.Worker do
     end)
   end
 
+  defp filter_transactions_by_block_size(txs, new_block_size_bytes) do
+    {valid_txs_by_block_size, _} =
+      List.foldl(txs, {[], false}, fn(tx, {selected_txs, limit_reached}) ->
+        if !limit_reached do
+          new_selected_txs = [tx | selected_txs]
+          new_selected_txs_size_bits =
+            new_selected_txs |> :erlang.term_to_binary() |> Bits.extract() |> Enum.count()
```
cytadela8: You should use `byte_size/1` function instead of `Bits.extract() |> Enum.count()` it should be much faster.

apps/aecore/lib/aecore/miner/worker.ex
```diff
@@ -240,31 +226,26 @@ defmodule Aecore.Miner.Worker do
...
-      %Block{header: unmined_header, txs: valid_txs}
+
+      new_block = create_block(top_block, chain_state, difficulty, [])
+      new_block_size_bits = new_block |> :erlang.term_to_binary() |> Bits.extract() |> Enum.count()
+      new_block_size_bytes = new_block_size_bits / 8
+
+      valid_txs_by_block_size = filter_transactions_by_block_size(valid_txs, new_block_size_bytes)
+
+      total_fees = calculate_total_fees(valid_txs_by_block_size)
+      valid_txs =
```
cytadela8: Why this replace?
`Filter_transactions_by_block_size` should preserve (and it looks like it preserves) order and remove transactions from the end, so the coinbase tx would never get removed as that's the first one.

apps/aecore/lib/aecore/miner/worker.ex
```diff
@@ -240,31 +226,26 @@ defmodule Aecore.Miner.Worker do
...
-          chain_state_hash,
-          difficulty,
-          0,
-          #start from nonce 0, will be incremented in mining
-          Block.current_block_version()
-        )
-      %Block{header: unmined_header, txs: valid_txs}
+
+      new_block = create_block(top_block, chain_state, difficulty, [])
+      new_block_size_bits = new_block |> :erlang.term_to_binary() |> Bits.extract() |> Enum.count()
```
cytadela8: `byte_size` here also

apps/aecore/lib/aecore/miner/worker.ex
```diff
@@ -24,6 +24,7 @@ defmodule Aecore.Miner.Worker do
 
   @mersenne_prime 2147483647
   @coinbase_transaction_value 100
+  @max_block_size_bytes 500_000
```
cytadela8: Please make max_block_size_bytes and Environment variable, so we can keep it separate for testing and "production"

apps/aecore/lib/aecore/miner/worker.ex
cheezus1: The replace is because we recalculate `total_fees`, based on the filtered transactions by block size.

apps/aecore/lib/aecore/miner/worker.ex
```diff
@@ -290,15 +268,56 @@ defmodule Aecore.Miner.Worker do
...
   end
 
+  defp filter_transactions_by_block_size(txs, new_block_size_bytes) do
+    {valid_txs_by_block_size, _} =
+      List.foldl(txs, {[], false}, fn(tx, {selected_txs, limit_reached}) ->
+        if !limit_reached do
+          new_selected_txs = [tx | selected_txs]
+          new_selected_txs_size_bytes =
+            new_selected_txs |> :erlang.term_to_binary() |> :erlang.byte_size()
+          if new_block_size_bytes + new_selected_txs_size_bytes <=
```
d-velev: why is the limit_reached false when the size equals the limit?

apps/aecore/lib/aecore/miner/worker.ex
```diff
@@ -290,15 +268,56 @@ defmodule Aecore.Miner.Worker do
...
-        |> Enum.count()
-      tx_size_bytes = tx_size_bits / 8
+      tx_size_bytes = tx |> :erlang.term_to_binary() |> :erlang.byte_size()
       tx.data.fee >= Float.floor(tx_size_bytes / miners_fee_bytes_per_token)
     end)
   end
 
+  defp filter_transactions_by_block_size(txs, new_block_size_bytes) do
+    {valid_txs_by_block_size, _} =
+      List.foldl(txs, {[], false}, fn(tx, {selected_txs, limit_reached}) ->
```
d-velev: We will go through every transaction even if we've reached the limit, a recursive function that stops when the limit is reached should be better here.

cytadela8:
@meivantodorov, @andonov, @nedelcho-delchev-tues Could you review, please?

apps/aecore/lib/aecore/miner/worker.ex
```diff
@@ -290,15 +270,64 @@ defmodule Aecore.Miner.Worker do
...
+
+  defp filter_transactions_by_block_size(txs, current_tx_index, txs_count, filtered_txs,
+    current_block_size_bytes, next_tx_size_bytes, max_block_size_bytes) do
+
+    current_tx = Enum.at(txs, current_tx_index)
+
+    new_filtered_txs = [current_tx | filtered_txs]
+    next_tx_index = current_tx_index + 1
+    if next_tx_index == txs_count do
+      Enum.reverse(new_filtered_txs)
```
thepiwo: why do we use the reverse here and at L300? This might be slow if done for every transaction

apps/aecore/lib/aecore/miner/worker.ex
```diff
@@ -290,15 +270,64 @@ defmodule Aecore.Miner.Worker do
...
   end
 
+  defp filter_transactions_by_block_size(txs, current_block_size_bytes, max_block_size_bytes) do
+    first_tx_size_bytes = txs |> Enum.at(0) |> Pool.get_tx_size_bytes()
+
+    filter_transactions_by_block_size(txs, 0, Enum.count(txs), [],
+      current_block_size_bytes, first_tx_size_bytes, max_block_size_bytes)
+  end
+
+  defp filter_transactions_by_block_size(txs, current_tx_index, txs_count, filtered_txs,
```
thepiwo: can you add comments in this function to explain how it works

apps/aecore/lib/aecore/miner/worker.ex
cheezus1: it is done at the end, upon finishing the filtering

thepiwo: ok great!

thepiwo:
@meivantodorov @d-velev can you review?

thepiwo:
@cheezus1 whats the progress here?

PR #201, opened 2018-02-01
[GH-#132] Finished integrating the elixir-wallet repository

### LINKED ISSUE #132: replace key worker with elixir-wallet
can be started when https://github.com/aeternity/elixir-wallet/issues/58 is finished

gspasov:
Removed the old Keys module and made a new Wallet module, to handle the elixir-wallet functions. Made a specified path to the `priv dir` inside the `aecore dir`, where the wallets will be stored under `wallets dir`

resolves #132


apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -3,33 +3,54 @@ defmodule Aecore.Structures.SignedTx do
   Aecore structure of a signed transaction.
   """
 
-  alias Aecore.Keys.Worker, as: Keys
   alias Aecore.Structures.TxData
   alias Aecore.Structures.SignedTx
 
+  @typedoc "Structure of the TxData"
+  @type tx() :: %TxData{}
```
thepiwo: this should be `@type t() :: %TxData{}` in tx_data.ex file

apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -3,33 +3,54 @@ defmodule Aecore.Structures.SignedTx do
...
+
+  @typedoc "Structure of the SignedTx module"
   @type t :: %SignedTx{
-    data: TxData.t(),
+    data: tx(),
```
thepiwo: keep this as `TxData.t()`

apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -3,33 +3,54 @@ defmodule Aecore.Structures.SignedTx do
...
+     * data: Aecore %TxData{} structure
+     * signature: Signed %TxData{} with the private key of the sender
   """
   defstruct [:data, :signature]
   use ExConstructor
 
-  @spec is_coinbase?(SignedTx.t()) :: boolean()
-  def is_coinbase?(tx) do
-    tx.data.from_acc == nil && tx.signature == nil
+  @spec is_coinbase(signed_tx()) :: boolean()
```
thepiwo: keep this as `SignedTx.t()`

apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -3,33 +3,54 @@ defmodule Aecore.Structures.SignedTx do
...
+  def is_coinbase(%{data: %{from_acc: key}, signature: signature}) do
+    key == nil && signature == nil
+  end
+
+  @spec is_valid(signed_tx()) :: boolean()
```
thepiwo: keep this as `SignedTx.t()`

apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -3,33 +3,54 @@ defmodule Aecore.Structures.SignedTx do
...
+  def is_valid(%{data: data, signature: signature}) do
+    data.value >= 0 && data.fee >= 0
+      Aewallet.Signing.verify(:erlang.term_to_binary(data), signature, data.from_acc)
```
thepiwo: is there an `&&` missing? otherwise, how does the verification work?

apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -3,33 +3,54 @@ defmodule Aecore.Structures.SignedTx do
...
+  Takes the transaction that needs to be signed
+  and the private key of the sender.
+  Returns a signed tx
+
+  ## Parameters
+     - tx: The transaction data that it's going to be signed
+     - priv_key: The priv key to sign with
+
+  """
+  @spec sign_tx(tx(), binary()) :: {:ok, %SignedTx{}}
```
thepiwo: `@spec sign_tx(TxData.t(), binary()) :: {:ok, SignedTx.t()}`

apps/aecore/lib/aecore/wallet/worker.ex
```diff
@@ -0,0 +1,84 @@
...
+  Module for handling the creation of a Wallet file
+  """
+
+  use GenServer
+
+  @typedoc "Options for network"
+  @type opts :: :mainnet | :testnet
+
+  @aewallet_dir Application.get_env(:aecore, :aewallet)[:path]
+  @aewallet_pass "1234"
```
thepiwo: why do we choose "1234" as password over "" as default?

apps/aecore/lib/aecore/miner/worker.ex
```diff
@@ -12,18 +12,19 @@ defmodule Aecore.Miner.Worker do
...
   alias Aecore.Txs.Pool.Worker, as: Pool
   alias Aeutil.Bits
   alias Aecore.Peers.Worker, as: Peers
+  alias Aecore.Wallet.Worker, as: Wallet
 
   require Logger
 
   @mersenne_prime 2147483647
   @coinbase_transaction_value 100
+  @wallet_default_pass "1234"
```
thepiwo: why do we choose "1234" as password over "" as default?

apps/aecore/lib/aecore/wallet/worker.ex
```diff
@@ -0,0 +1,84 @@
...
+    end
+  end
+
+  defp create_wallet(path) do
+    Aewallet.Wallet.create_wallet(@aewallet_pass, path)
+  end
+
+  defp get_file_name(path) do
+    path
+    |> Path.join("*/")
```
thepiwo: Why is this needed?

apps/aecore/lib/aecore/miner/worker.ex
gspasov: Hmm I just did thought of that, I'll change it

apps/aecore/lib/aecore/structures/signed_tx.ex
gspasov: Yes, my fault. I saw it later, it's fixed

apps/aecore/lib/aecore/wallet/worker.ex
gspasov: As I mentioned up, I'll fix this

apps/aecore/lib/aecore/wallet/worker.ex
gspasov: This is done so that I could search for files I'm not sure if exist inside the aewallet dir. This star could be actually added inside the config path. I'll fix it.

apps/aecore/lib/aecore/wallet/worker.ex
```diff
@@ -0,0 +1,84 @@
...
+    case get_file_name(@aewallet_dir) do
+      []  -> create_wallet(@aewallet_dir)
+      [_] -> :ok
+    end
+  end
+  defp has_wallet(false) do
+    if File.mkdir @aewallet_dir do
+      create_wallet(@aewallet_dir)
+    else
+      throw("Failed creating the wallet dir: #{@aewallet_dir}")
```
d-velev: this should be caught in the init

apps/aecore/lib/aecore/wallet/worker.ex
```diff
@@ -0,0 +1,84 @@
...
+  @typedoc "Options for network"
+  @type opts :: :mainnet | :testnet
+
+  @aewallet_dir Application.get_env(:aecore, :aewallet)[:path]
+  @aewallet_pass "1234"
+
+  ## Client API
+
+  def start_link(_args) do
+    GenServer.start_link(__MODULE__, %{path: @aewallet_dir}, name: __MODULE__)
```
d-velev: Why do we need a GenServer when we don't need to keep state?

apps/aecore/lib/aecore/miner/worker.ex
```diff
@@ -233,7 +234,7 @@ defmodule Aecore.Miner.Worker do
       valid_txs_by_chainstate = BlockValidation.filter_invalid_transactions_chainstate(ordered_txs_list, chain_state, top_block.header.height + 1)
       valid_txs_by_fee = filter_transactions_by_fee(valid_txs_by_chainstate)
 
-      {_, pubkey} = Keys.pubkey()
+      pubkey = Wallet.get_public_key(@wallet_default_pass)
```
d-velev: What if the password is changed?

apps/aecore/lib/aecore/miner/worker.ex
d-velev: If this is used in multiple locations then maybe it should be stored in the config or have it as an attribute in one module and have a getter function for it.

apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -46,9 +47,9 @@ defmodule Aecore.Structures.SignedTx do
      - priv_key: The priv key to sign with
 
   """
-  @spec sign_tx(tx(), binary()) :: {:ok, %SignedTx{}}
+  @spec sign_tx(tx(), binary()) :: {:ok, t()}
```
thepiwo: Is it possible to write `SignedTx.t()` instead of just `t()`?

apps/aecore/lib/aecore/wallet/worker.ex
```diff
@@ -0,0 +1,84 @@
...
+    {:ok, state}
+  end
+
+  @spec get_public_key(String.t()) :: binary()
+  def get_public_key(password) do
+    GenServer.call(__MODULE__, {:get_pub_key, {password, :mainnet}})
+  end
+
+  @spec get_public_key(String.t(), opts()) :: binary()
+  def get_public_key(password, network) do
```
d-velev: Do we need to protect the public key with a password?

apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -3,33 +3,55 @@ defmodule Aecore.Structures.SignedTx do
...
+
+  ## Parameters
+     - tx: The transaction data that it's going to be signed
+     - priv_key: The priv key to sign with
+
+  """
+  @spec sign_tx(tx(), binary()) :: {:ok, t()}
+  def sign_tx(tx, priv_key) do
+    signature = Signing.sign(:erlang.term_to_binary(tx), priv_key)
+    {:ok, %SignedTx{data: tx, signature: signature}}
```
d-velev: Maybe this should just return the SignedTx without :ok since there are no checks if this is the desired result.

apps/aecore/lib/aecore/miner/worker.ex
cheezus1: Maybe store it in the config, as it is used both in the code and the tests.

apps/aecore/lib/aecore/wallet/worker.ex
gspasov: The password is used to decrypt the wallet. Everytime we derive some key, we need to decrypt the wallet.

apps/aecore/lib/aecore/wallet/worker.ex
gspasov: The GenServer was used so that we can start it from the Supervisor because we need to be sure that a wallet is created before starting the miner.

apps/aecore/lib/aecore/structures/signed_tx.ex
gspasov: Well, I'll keep it like this and make some verifications, as well in the TxData module, because now when you mention it I can see that it's needed.

apps/aecore/lib/aecore/miner/worker.ex
gspasov: Yup, on it.

apps/aecore/lib/aecore/miner/worker.ex
gspasov: Same as the above question. I'm on it.

apps/aecore/lib/aecore/wallet/worker.ex
gspasov: Well in the init I check if there is a Wallet file and then call the create_wallet func, but I'll think about this a bit more.

apps/aecore/lib/aecore/wallet/worker.ex
d-velev: You can call the function that does this in the aecore.ex module (the start function) and get rid of the GenServer.

gspasov: Well now when I think about it, the state could store the public key of the wallet, so that the module won't need to derive it everytime it is needed.

thepiwo: @gspasov I think so as well

apps/aecore/lib/aecore/wallet/worker.ex
```diff
@@ -0,0 +1,92 @@
...
+
+  use GenServer
+
+  alias Aewallet.Wallet
+
+  @typedoc "Options for network"
+  @type opts :: :mainnet | :testnet
+
+  @aewallet_dir Application.get_env(:aecore, :aewallet)[:path]
+  @aewallet_pass Application.get_env(:aecore, :aewallet)[:pass]
```
cytadela8: If I'm not mistaken this will get evaluated compile time, not runtime!

apps/aecore/lib/aecore/wallet/worker.ex
```diff
@@ -0,0 +1,92 @@
...
+  end
+
+  @spec get_private_key(String.t(), opts()) :: binary()
+  def get_private_key(password, network) do
+    GenServer.call(__MODULE__, {:get_priv_key, {password, network}})
+  end
+
+  ## Server Callbacks
+
+  def handle_call({:get_pub_key, {password, network}}, _from, %{pubkey: nil} = state) do
```
cytadela8: I think this genserver could also handle "password" and "network" parameters internally. (store them, probably load them from config)

gspasov: Having them stored initialy actually loses the purpose of having them in the first place. In my opinion a password should not be stored anywhere and the network should be able to change on the fly.

apps/aecore/lib/aecore/wallet/worker.ex
gspasov: Will make two function for calling them at run time

thepiwo:

apps/aecore/lib/aecore/wallet/worker.ex
cytadela8: - I think the network should be part of some config, becouse changing the network requires preaty much to restart everything (different database for blocks, different genesis block, different peers). Also even if we would want to change it on the fly it still would be nice if the information would be stored in the GenServer, this way code that wants to sign something doesn't have to pass the information about network every time. 
- By storing password I mean storing in the state of GenServer in RAM.

apps/aecore/lib/aecore/miner/worker.ex
```diff
@@ -233,7 +233,7 @@ defmodule Aecore.Miner.Worker do
       valid_txs_by_chainstate = BlockValidation.filter_invalid_transactions_chainstate(ordered_txs_list, chain_state, top_block.header.height + 1)
       valid_txs_by_fee = filter_transactions_by_fee(valid_txs_by_chainstate)
 
-      {_, pubkey} = Keys.pubkey()
+      pubkey = Wallet.get_public_key(Wallet.get_aewallet_dir())
```
meivantodorov: I think if you are calling internal function to get the path then better do it inside the Wallet.

apps/aecore/lib/aecore/wallet/worker.ex
meivantodorov: May be we can let the user to decide if he/she wants to keep the password into the memory or to type it every time.

thepiwo:
can you add to the readme that `build-essential autoconf autogen libtool libgmp3-dev` are dependencies

thepiwo:
@nedelcho-delchev-tues @cytadela8 @d-velev @meivantodorov @cheezus1 can you review?

d-velev:
Can you replace the :erlang.term_to_binary with Serialization.pack_binary like it was before?

apps/aecore/lib/aecore/wallet/worker.ex
```diff
@@ -0,0 +1,143 @@
...
+    {:reply, key, state}
+  end
+
+  def handle_call({:get_priv_key, {password, network}}, _from, %{path: path} = state) do
+    {:ok, priv_key} =
+      Wallet.get_private_key(get_file_name(path), password, network: network)
+    {:reply, priv_key, state}
+  end
+
+  def handle_call({:bech32_encode, {pub_key, prefix}}, _from, state) do
```
d-velev: No reason for this or the decode function to be calls.

apps/aecore/test/multiple_transactions_test.exs
```diff
@@ -279,172 +301,183 @@ defmodule MultipleTransactionsTest do
...
   end
 
   defp create_signed_tx(from_acc, to_acc, value, nonce, fee, lock_time_block \\ 0) do
     {from_acc_pub_key, from_acc_priv_key} = from_acc
     {to_acc_pub_key, _to_acc_priv_key} = to_acc
     {:ok, tx_data} = SpendTx.create(from_acc_pub_key, to_acc_pub_key, value,
-                                   nonce, fee, lock_time_block)
-    {:ok, signature} = Keys.sign(tx_data, from_acc_priv_key)
+      nonce, fee, lock_time_block)
+    {:ok, %{signature: signature}} = SignedTx.sign_tx(tx_data, from_acc_priv_key)
```
cytadela8: Here you are extracting the signature from SignedTx just to create the signed_tx again. You should instead:
```
    {:ok, signed_tx} = SignedTx.sign_tx(tx_data, from_acc_priv_key)
    signed_tx
end
```

thepiwo:
@cheezus1 @d-velev can you re-review?

thepiwo:
@cheezus1 @d-velev can you re-review?

apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -24,13 +25,33 @@ defmodule Aecore.Structures.SignedTx do
...
+  def is_coinbase?(%{data: %{from_acc: key}, signature: signature}) do
+    key == nil && signature == nil
   end
 
   @spec is_valid?(SignedTx.t()) :: boolean()
-  def is_valid?(tx) do
-    tx.data.value >= 0 && tx.data.fee >= 0 && Keys.verify_tx(tx)
+  def is_valid?(%{data: data, signature: signature}) do
+    data.value >= 0 && data.fee >= 0 &&
+      Signing.verify(Serialization.pack_binary(data), signature, data.from_acc)
```
d-velev: Maybe move pack_binary to Signing.verify, it would be easier to change it and it won't have to be written every time something is signed

apps/aecore/lib/aecore/wallet/worker.ex
```diff
@@ -0,0 +1,155 @@
...
+  @typedoc "Wallet type"
+  @type wallet_type :: :ae | :btc
+
+  @typedoc "Options for network"
+  @type opts :: :mainnet | :testnet
+
+  ## Client API
+
+  def start_link(_args) do
+    GenServer.start_link(__MODULE__, %{path: get_aewallet_dir(), pubkey: nil}, name: __MODULE__)
```
d-velev: I think the path shouldn't be stored here since it's already in the env and there is no function to alter it. In general, I think that functions that change the state should be available if a genserver is used.

apps/aecore/lib/aecore/structures/signed_tx.ex
gspasov: This function is implemented in the `elxir-wallet` repository so I don't think it's a good idea to change it's base functionality for this. It takes binary as the first argument and I'd like to keep it this way. This way the function is more generic and not bouned to some specific implementation.

thepiwo: ok, lets keep it like this

apps/aecore/lib/aecore/wallet/worker.ex
thepiwo: @d-velev can you approve now?

d-velev:
Is it possible to create a new keypair for testing purposes? For example: we already have a wallet file created and it is read by 2 nodes on one machine, if they both have the same public key we would want to create a new one on one of the nodes so that we can see chainstate changes.

thepiwo:
@gspasov can you give an answer here, so we can merge soon?

gspasov:
@thepiwo I'm going to work on the comment/suggestion that @d-velev gave me.
Sorry for the delay

apps/aecore/lib/aecore/wallet/worker.ex
```diff
@@ -47,32 +48,42 @@ defmodule Aecore.Wallet.Worker do
 
   @spec get_public_key() :: binary()
   def get_public_key() do
-    get_public_key(get_aewallet_pass())
+    get_public_key("")
```
thepiwo: why is the default password hardcoded now?

apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -25,13 +26,33 @@ defmodule Aecore.Structures.SignedTx do
...
+  and the private key of the sender.
+  Returns a signed tx
+
+  ## Parameters
+     - tx: The transaction data that it's going to be signed
+     - priv_key: The priv key to sign with
+
+  """
+  @spec sign_tx(SpendTx.t(), binary()) :: {:ok, SignedTx.t()}
+  def sign_tx(%SpendTx{} = tx, priv_key) when byte_size(priv_key) == 32 do
```
thepiwo: is it possible to not pass the priv_key here, but have it in the `Signing.sign` method?

apps/aecore/lib/aecore/wallet/worker.ex
gspasov: No, it's not hardcoded, this is the default `derivation_path`. Now as first argument to the `get_public_key()` I have the `derivation_path`, so the full func is `get_public_key(derivation_path, passoword, network)`

thepiwo: okay, can we make a variable for the default `derivation_path` then?

apps/aecore/lib/aecore/structures/signed_tx.ex
gspasov: No, I need the private key to be passed to the `Aewallet.Signing.sing()` function, otherwise it cannot sign the data

thepiwo: can you fetch it in `sign_tx`?

gspasov: Well, I guess I could call `Wallet.get_private_key()`

thepiwo: ok, would be better than passing the private_key every time

thepiwo: can't this be called in the `Signing.sign` function?

gspasov: This is the function in the `elixir-wallet` repo

thepiwo: ok, thanks, I didn't notice

gspasov:
@thepiwo can you check and merge?

apps/aecore/lib/aecore/structures/signed_tx.ex
thepiwo: @gspasov the private-ley is still passed, could you pass `Wallet.get_private_key()` as mentioned?

PR #212, opened 2018-02-07
[GH-#172] Oracles

### LINKED ISSUE #172: Oracles implementation
Edit: issue description not relevant anymore, to be updated later on.

d-velev:
closes #172


apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -112,7 +113,12 @@ defmodule Aecore.Chain.BlockValidation do
   @spec validate_transaction_chainstate(SignedTx.t(), ChainState.account_chainstate(), integer()) :: {boolean(), map()}
   defp validate_transaction_chainstate(tx, chain_state, block_height) do
     try do
-      {true, ChainState.apply_transaction_on_state!(tx, chain_state, block_height)}
+      case tx do
```
thepiwo: why is here a differentiation? you could just not use the `block_height` inside the function

apps/aecore/lib/aecore/structures/block.ex
```diff
@@ -4,11 +4,11 @@ defmodule Aecore.Structures.Block do
...
   alias Aecore.Structures.Block
   alias Aecore.Structures.Header
-  alias Aecore.Structures.SignedTx
+  alias Aecore.Structures.TxData
 
   @type t :: %Block{
-    header: Header.t,
-    txs: list(SignedTx.t())
+    header: Header.t(),
+    txs: list(TxData.t())
```
thepiwo: this should stay `SignedTx`

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -53,7 +53,7 @@ defmodule Aecore.Peers.Sync do
     GenServer.call(__MODULE__, {:set_chain_sync_status, status})
   end
 
-  @spec add_block_to_state(binary(), term()) :: :ok
+  @spec add_block_to_state(binary, term()) :: :ok
```
thepiwo: `term()` could even get `Block.t()`

apps/aecore/lib/aecore/structures/oracle_response_tx_data.ex
```diff
@@ -0,0 +1,43 @@
...
+                            response: response, fee: fee,
+                            nonce: Chain.lowest_valid_nonce()}
+    else
+      :error
+    end
+  end
+
+  @spec is_oracle_response_tx(map()) :: boolean()
+  def is_oracle_response_tx(tx) do
+    Map.has_key?(tx, "operator") && Map.has_key?(tx, "oracle_hash") &&
```
thepiwo: I'd move this to `SignedTx`

```
@spec is_oracle_response_tx(SignedTx.t()) :: boolean()
def is_oracle_response_tx(tx) do
   case tx.data do
      %OracleResponseTxData{} -> true
      _ -> false
   end
end
```

apps/aecore/lib/aecore/structures/oracle_registration_tx_data.ex
```diff
@@ -0,0 +1,50 @@
...
+                                fee: fee, nonce: Chain.lowest_valid_nonce()}
+    rescue
+      e ->
+       Logger.error("Invalid query or response format definition; " <> e.message)
+       :error
+    end
+  end
+
+  @spec is_oracle_registration_tx(map()) :: boolean()
+  def is_oracle_registration_tx(tx) do
```
thepiwo: I'd move this to `SignedTx` as well

apps/aecore/lib/aecore/structures/oracle_query_tx_data.ex
```diff
@@ -0,0 +1,51 @@
...
+      true ->
+        {:ok, pubkey} = Keys.pubkey()
+        %OracleQueryTxData{sender: pubkey, oracle_hash: oracle_hash,
+                           query_data: query_data, query_fee: query_fee,
+                           fee: fee, nonce: Chain.lowest_valid_nonce()}
+    end
+  end
+
+  @spec is_oracle_query_tx(map()) :: boolean()
+  def is_oracle_query_tx(tx) do
```
thepiwo: I'd move this to `SignedTx` as well

apps/aecore/lib/aecore/miner/worker.ex
```diff
@@ -210,21 +210,6 @@ defmodule Aecore.Miner.Worker do
     top_block_hash = BlockValidation.block_header_hash(top_block.header)
     chain_state = Chain.chain_state(top_block_hash)
 
-    # We take an extra block and then drop one at the head of the list
```
thepiwo: why is this not needed any more?

apps/aecore/lib/aecore/miner/worker.ex
```diff
@@ -241,11 +226,14 @@ defmodule Aecore.Miner.Worker do
...
                       Application.get_env(:aecore, :tx_data)[:lock_time_coinbase]) |
                    valid_txs_by_fee]
-      root_hash = BlockValidation.calculate_root_hash(valid_txs)
 
       new_chain_state = ChainState.calculate_and_validate_chain_state!(valid_txs, chain_state, top_block.header.height + 1)
       chain_state_hash = ChainState.calculate_chain_state_hash(new_chain_state)
 
+      #valid_txs_with_oracle_registration_txs = oracle_registration_txs ++ valid_txs
+
+      root_hash = BlockValidation.calculate_root_hash(valid_txs)
```
thepiwo: doesn't this include the `oracle_registration_txs` ?

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -214,13 +316,56 @@ defmodule Aecore.Chain.Worker do
...
+    if Application.get_env(:aecore, :operator)[:is_node_operator] do
+      oracles = Application.get_env(:aecore, :operator)[:oracles]
+      Enum.each(block.txs, fn(tx) ->
+          if(match?(%OracleQueryTxData{}, tx.data) &&
+             Map.has_key?(oracles, tx.data.oracle_hash)) do
+            encoded_tx_hex_hash =
+              tx
+              |> Serialization.tx(:serialize)
+              |> Poison.encode!()
+            HTTPoison.post(oracles[tx.data.oracle_hash],
```
thepiwo: move to `Aehttpclient`

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -214,13 +316,56 @@ defmodule Aecore.Chain.Worker do
     end
   end
 
-  defp update_txs_index(current_txs_index, new_block_txs_index) do
-    Map.merge(current_txs_index, new_block_txs_index,
+  defp update_txs_index_or_oracle_responses(map1, map2) do
```
thepiwo: can you rename them semantical again?

apps/aeutil/lib/serialization.ex
```diff
@@ -7,24 +7,91 @@ defmodule Aeutil.Serialization do
...
+          signature = hex_binary(tx.signature, direction)
+          {data, signature}
+      end
+    %SignedTx{data: new_data, signature: new_signature}
+  end
+
+  @spec build_tx(map()) :: transaction_types()
+  def build_tx(tx) do
+    cond do
+      TxData.is_tx_data_tx(tx["data"]) ->
```
thepiwo: doesn't `tx.data` work?

apps/aecore/lib/aecore/miner/worker.ex
d-velev: We take the blocks a few lines down, I don't know why that was still there

apps/aecore/lib/aecore/miner/worker.ex
d-velev: It does, I've forgotten to remove the comment

apps/aeutil/lib/serialization.ex
d-velev: It works

apps/aecore/lib/aecore/chain/block_validation.ex
d-velev: I would have to make a really big case, I think it's better this way

apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -43,4 +43,31 @@ defmodule Aecore.Structures.SignedTx do
     end
   end
 
+  @spec is_oracle_query_tx(map()) :: boolean()
+  def is_oracle_query_tx(tx) do
+    Map.has_key?(tx, "sender") && Map.has_key?(tx, "oracle_hash") &&
```
thepiwo: wouldn't pattern match tx.data type be better?

thepiwo: ah I didn't noticed the serialization uses it for both directions, so it has to be like this

apps/aecore/lib/aecore/structures/oracle_response_tx_data.ex
```diff
@@ -20,7 +20,7 @@ defmodule Aecore.Structures.OracleResponseSpendTx do
              :nonce]
   use ExConstructor
 
-  @spec create(binary(), any(), integer()) :: %OracleResponseSpendTx{}
+  @spec create(binary(), any(), integer()) :: t
```
thepiwo: maybe even do OracleResponseSpendTx.t()

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -117,7 +118,12 @@ defmodule Aecore.Chain.BlockValidation do
...
+        %SignedTx{data: %SpendTx{}} ->
+          {true, ChainState.apply_transaction_on_state!(tx, chain_state, block_height)}
+        _oracle_tx ->
```
thepiwo: can you specifically match the oracle tx types?

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -52,21 +63,56 @@ defmodule Aecore.Chain.ChainState do
...
+    operator_address =
+      Chain.registered_oracles[transaction.data.oracle_hash].data.operator
+    sender_state =
+      Map.get(chain_state, transaction.data.sender,
+              %{balance: 0, nonce: 0, locked: []})
+    operator_state =
+      Map.get(chain_state, operator_address,
+              %{balance: 0, nonce: 0, locked: []})
+    cond do
+      sender_state.balance -
```
thepiwo: would it be possible to call `deduct_from_account_state` here?

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -251,13 +343,53 @@ defmodule Aecore.Chain.Worker do
...
+    Map.merge(prev_block_data, new_block_data,
       fn(_, current_list, new_block_list) ->
         current_list ++ new_block_list
       end)
   end
 
+  defp generate_registrated_oracles_map(block) do
+    Enum.reduce(block.txs, %{}, fn(tx, acc) ->
+        if(match?(%OracleRegistrationTxData{}, tx.data)) do
+          Map.put(acc, :crypto.hash(:sha256, :erlang.term_to_binary(tx)), tx)
```
meivantodorov: You can use SignedTx.hash_tx(tx) to hash a signed transaction

thepiwo:
Can you add some documentation in the readme?

apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -26,12 +27,50 @@ defmodule Aecore.Structures.SignedTx do
...
 
   @spec is_valid?(SignedTx.t()) :: boolean()
   def is_valid?(tx) do
-    tx.data.value >= 0 && tx.data.fee >= 0 && Keys.verify_tx(tx)
+    if(match?(%SpendTx{}, tx.data)) do
+      not_negative = tx.data.value >= 0
+      signature_valid = Keys.verify_tx(tx)
+      not_negative && signature_valid
+    else
+      true
```
cytadela8: We have to validate signatures of Orace txs also. _This connects with my comments in Chain.ChainState_

apps/aecore/lib/aecore/txs/pool/worker.ex
```diff
@@ -78,7 +76,9 @@ defmodule Aecore.Txs.Pool.Worker do
           Logger.info("Transaction is already in pool")
         else
           # Broadcasting notifications for new transaction in a pool(per account and every)
-          Notify.broadcast_new_transaction_in_the_pool(tx)
+          if match?(%SpendTx{}, tx.data) do
+            Notify.broadcast_new_transaction_in_the_pool(tx)
```
cytadela8: Why do we only broadcast SpendTx-s?

cytadela8: (This limits the chance any oracle tx will be minded in a big network significantly.)

apps/aecore/oracle.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def do_HEAD(self):
+        self._set_headers()
+
+    def do_POST(self):
+        content_length = int(self.headers['Content-Length'])
+        post_data = json.loads(self.rfile.read(content_length))['data']
+        get_data = json.loads(urllib2.urlopen("https://api.fixer.io/latest?symbols=" + post_data['query_data']['currency']).read())
+        payload = {"oracle_hash":post_data['oracle_hash'], "response":get_data, "fee":10}
+        req = urllib2.Request('http://localhost:4000/oracle_response')
```
cytadela8: I have trouble understanding the purpose of this python http server. Also:
- it has hardcoded node port
- it is open to the word (?)
- it has some wird hardcoded call to "api.fixer.io"

apps/aehttpclient/lib/client.ex
```diff
@@ -92,13 +92,24 @@ defmodule Aehttpclient.Client do
...
+      |> Poison.encode!()
+    HTTPoison.post(uri,
+                   encoded_serialized_tx,
+                   [{"Content-Type", "application/json"}],
+                   [stream_to: self()])
+  end
+
   defp handle_response(:block, body, _headers) do
-    response = Poison.decode!(body, as: %Block{}, keys: :atoms!)
+    response = Poison.decode!(body)
```
cytadela8: Why is this changed? (Is everything decoded correctly still?)

apps/aehttpserver/lib/aehttpserver/web/channels/notify.ex
```diff
@@ -13,14 +14,16 @@ defmodule Aehttpserver.Web.Notify do
 
   def broadcast_new_block_added_to_chain_and_new_mined_tx(block) do
     Enum.each(block.txs, fn(tx) ->
-      Aehttpserver.Web.Endpoint.broadcast!("room:notifications", "new_mined_tx_everyone", %{"body" => Serialization.tx(tx, :serialize)})
-      if tx.data.from_acc != nil do
-        Aehttpserver.Web.Endpoint.broadcast!("room:notifications", "new_mined_tx:" <> Base.encode16(tx.data.from_acc), %{"body" => Serialization.tx(tx, :serialize)})
+      if match?(%SpendTx{}, tx.data) do
```
cytadela8: I think we should also broadcast Oracle txs.

apps/aehttpserver/lib/aehttpserver/web/controllers/oracle_controller.ex
```diff
@@ -0,0 +1,19 @@
+defmodule Aehttpserver.Web.OracleController do
+  use Aehttpserver.Web, :controller
+
+  alias Aecore.OraclePrototype.Oracle
+  alias Aeutil.Bits
+
+  require Logger
+
+  def oracle_response(conn, _params) do
+    body = conn.body_params
```
cytadela8: `/oracle_response` should have some authorization as it's visible to the world. (You can for example check if the response come from localhost, that should be good enough for now)

apps/aehttpserver/lib/aehttpserver/web/router.ex
```diff
@@ -19,6 +19,8 @@ defmodule Aehttpserver.Web.Router do
     get "/peers", PeersController, :info
     resources "/tx", TxController, param: "account", only: [:show]
     post "/new_block", BlockController, :new_block
+    post "/oracle_response", OracleController, :oracle_response
+    get "/peers", PeersController, :info
```
cytadela8: This line is duplicated. (`get "/peers", PeersController, :info`)

apps/aeutil/lib/serialization.ex
```diff
@@ -8,28 +8,97 @@ defmodule Aeutil.Serialization do
...
+        OracleResponseTxData.new(tx_data)
+    end
+  end
+
+  @spec serialize_tx_data(transaction_types(),
+                          :serialize | :deserialize) :: Serialization.transaction_types()
+  def serialize_tx_data(tx_data, direction) do
+    case tx_data do
+      %SpendTx{} ->
+        %{tx_data | from_acc: hex_binary(tx_data.from_acc, direction),
```
cytadela8: I would suggest adding "type" field for easy deserialization. (Such field would be generally useful for anyone using our APIs)

apps/aeutil/test/aeutil_serialization_test.exs
```diff
@@ -49,24 +50,79 @@ defmodule AeutilSerializationTest do
...
+              164, 173, 185, 138, 50, 126, 89, 81, 12, 91, 169, 185, 211, 52, 219>>,
+            value: 100
+          },
+          signature: nil
+        }
+      ]
+    }
   end
 
+  def get_block_map() do
```
cytadela8: What is this function/"constant" used for? The only call doesn't seem to be used for anything.

d-velev: It's used in the test, `assert Serialization.block(block_map, :deserialize) == block`

apps/aeutil/lib/serialization.ex
d-velev: This doesn't seem like it would improve internal work

thepiwo: @cytadela8 yes I thought of this as well, I will prepare a task to do this

apps/aecore/lib/aecore/structures/signed_tx.ex
thepiwo: good catch

apps/aecore/lib/aecore/txs/pool/worker.ex
thepiwo: oh noticed this as well, but forgot to comment

apps/aecore/oracle.py
thepiwo: it is a sample oracle data providing service, for merging, we should remove this @d-velev

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -165,4 +221,16 @@ defmodule Aecore.Chain.ChainState do
                        transaction.data.lock_time_block)
   end
 
+  defp deduct_from_account_state!(chain_state, account, value, nonce) do
```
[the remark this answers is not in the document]
d-velev: Apply to state also applies a lock, I think it wouldn't work

apps/aecore/lib/aecore/txs/pool/worker.ex
d-velev: This only excludes the transaction from websocket notifications, not  peer broadcast

apps/aecore/oracle.py
d-velev: The oracle test uses it

apps/aehttpclient/lib/client.ex
d-velev: All of the transformations are done in the Serialization module

apps/aehttpserver/lib/aehttpserver/web/channels/notify.ex
thepiwo: yes, we should @d-velev

apps/aecore/lib/aecore/chain/chain_state.ex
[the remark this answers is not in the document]
thepiwo: this should work, the lock is handled differently for coinbase and normal spend transactions anyways.

gspasov:
Is this RP going to wait for PR #231 that @rdimitrovbg and I are currently working on?

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -31,11 +38,16 @@ defmodule Aecore.Chain.Worker do
     genesis_chain_state = ChainState.calculate_and_validate_chain_state!(Block.genesis_block().txs, %{}, 0)
     chain_states = %{genesis_block_hash => genesis_chain_state}
     txs_index = calculate_block_acc_txs_info(Block.genesis_block())
+    registered_oracles = generate_registered_oracles_map(Block.genesis_block())
+    oracle_responses = generate_oracle_response_map(Block.genesis_block())
```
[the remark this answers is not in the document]
thepiwo: @cytadela8 if we address forking via a snapshots task in the future, would this be ok, as is?

thepiwo:
@gspasov I think it should, depending on how long you take.

apps/aecore/lib/aecore/chain/chain_state.ex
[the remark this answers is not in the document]
cytadela8: @d-velev It only applies the lock if locktime > current height

apps/aecore/lib/aecore/chain/worker.ex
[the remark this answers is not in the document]
cytadela8: `registered_oracles` and `oracle_responses` would have to be part of snapshots. 

The problem is that `registered_oracles` and `oracle_responses` are part of blockchain state like our `chain_state` structure. It has to be treated equally to `chain_state`.

apps/aecore/lib/aecore/txs/pool/worker.ex
cytadela8: So then this should be fixed if "comment" `apps/aehttpserver/lib/aehttpserver/web/channels/notify.ex` line 17 should be fixed.

cytadela8:
Please let me now when you would like another review.

thepiwo:
adjust to match https://github.com/aeternity/protocol/blob/master/oracles/oracles.md specs, not considered prototype any more

d-velev:
@cytadela8 You can review now, serialization is messy but we can open an issue for that.

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -168,7 +178,7 @@ defmodule Aecore.Chain.Worker do
     add_validated_block(block, new_chain_state)
   end
 
-  @spec add_validated_block(Block.t(), ChainState.account_chainstate()) :: :ok
+  @spec add_validated_block(Block.t(), map()) :: :ok
```
thepiwo: shouldn't this `map()` type be `ChainState.chainstate()`

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -187,7 +197,16 @@ defmodule Aecore.Chain.Worker do
     end
   end
 
-  @spec chain_state() :: ChainState.account_chainstate()
+  @spec registered_oracles() :: map()
```
thepiwo: be more specific about `map()`

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -187,7 +197,16 @@ defmodule Aecore.Chain.Worker do
...
+  def registered_oracles() do
+    GenServer.call(__MODULE__, :registered_oracles)
+  end
+
+  @spec oracle_interaction_objects() :: map()
```
thepiwo: be more specific about `map()`

apps/aecore/lib/aecore/oracle/oracle.ex
```diff
@@ -0,0 +1,272 @@
...
+
+  require Logger
+
+  @type ttl :: %{ttl: non_neg_integer(), type: :relative | :absolute}
+
+  @doc """
+  Registers an oracle with the given requirements for queries and responses,
+  a fee that should be paid by queries and a TTL.
+  """
+  @spec register(map(), map(), non_neg_integer(), non_neg_integer(), ttl()) :: :ok | :error
```
thepiwo: be more specific about `map()` if possible for this case

apps/aecore/lib/aecore/oracle/oracle.ex
```diff
@@ -0,0 +1,272 @@
...
+
+    {:ok, tx} = SignedTx.sign_tx(tx_data, Wallet.get_private_key())
+    Pool.add_transaction(tx)
+  end
+
+  @doc """
+  Creates a query transaction with the given oracle address, data query
+  and a TTL of the query and response.
+  """
+  @spec query(binary(), any(), non_neg_integer(), non_neg_integer(), ttl(), ttl()) :: :ok | :error
```
thepiwo: be more specific about `map()` and `binary()`

apps/aecore/lib/aecore/oracle/oracle.ex
```diff
@@ -0,0 +1,272 @@
...
+
+    {:ok, tx} = SignedTx.sign_tx(tx_data, Wallet.get_private_key())
+    Pool.add_transaction(tx)
+  end
+
+  @doc """
+  Creates an oracle response transaction with the query referenced by its
+  transaction hash and the data of the response.
+  """
+  @spec respond(binary(), any(), non_neg_integer()) :: :ok | :error
```
thepiwo: be more specific about `binary()` and `any()`, type redefinitions for your usecase are fine, to give the parameter types semantic names

apps/aecore/lib/aecore/oracle/oracle.ex
```diff
@@ -0,0 +1,272 @@
...
+                                                                            query_sender:
+                                                                              query_sender,
+                                                                            response: response,
+                                                                            query_height_included:
+                                                                              query_height_included,
+                                                                            response_height_included:
+                                                                              response_height_included
+                                                                          }},
+                                                                         acc ->
+      query_absolute_ttl =
```
thepiwo: maybe you can put the reduce function in a separate `def`

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -54,7 +55,7 @@ defmodule Aecore.Peers.Sync do
     GenServer.call(__MODULE__, {:set_chain_sync_status, status})
   end
 
-  @spec add_block_to_state(binary(), term()) :: :ok
+  @spec add_block_to_state(binary, Block.t()) :: :ok
```
thepiwo: `()` for the type definition is missing

apps/aecore/lib/aecore/structures/oracle_extend_tx_data.ex
```diff
@@ -0,0 +1,123 @@
...
+          OracleExtendTxData.t(),
+          Wallet.pubkey(),
+          ChainState.account(),
+          non_neg_integer(),
+          non_neg_integer(),
+          non_neg_integer(),
+          Oracle.registered_oracles()
+        ) :: :ok | {:error, String.t()}
+  def preprocess_check(tx, sender, account_state, fee, nonce, _block_height, registered_oracles) do
+    cond do
```
thepiwo: shouldn't this fail if an absolute ttl is below height?

apps/aecore/lib/aecore/structures/oracle_registration_tx_data.ex
```diff
@@ -0,0 +1,190 @@
...
+        response_format: response_format,
+        ttl: ttl
+      }) do
+    formats_valid =
+      try do
+        ExJsonSchema.Schema.resolve(query_format)
+        ExJsonSchema.Schema.resolve(response_format)
+        true
+      rescue
+        e ->
```
thepiwo: why rescue instead of catch? use full error name, instead of e

apps/aecore/lib/aecore/txs/pool/worker.ex
thepiwo: this is still a valid comment for @d-velev to address

apps/aehttpserver/lib/aehttpserver/web/controllers/oracle_controller.ex
thepiwo: as far as I can see, this is addressed already

apps/aecore/lib/aecore/structures/oracle_extend_tx_data.ex
d-velev: The TTL in this transaction is just an integer which is added to the oracle %{ttl: value}

apps/aecore/lib/aecore/structures/oracle_registration_tx_data.ex
d-velev: The resolve function raises an error instead of throwing, I also don't know what the possible errors are and they wouldn't be handled in a different way

apps/aecore/lib/aecore/oracle/oracle.ex
d-velev: Maps should be json schemas, not possible

apps/aecore/lib/aecore/structures/oracle_extend_tx_data.ex
thepiwo: okay

apps/aecore/lib/aecore/oracle/oracle.ex
thepiwo: but you could add a `json_schema` type to alias map for some use-cases

apps/aecore/lib/aecore/oracle/oracle.ex
d-velev: I've set the query_data type to any, it's not map anymore

apps/aecore/lib/aecore/oracle/oracle.ex
```diff
@@ -0,0 +1,293 @@
...
+
+  @spec calculate_relative_ttl(%{ttl: non_neg_integer(), type: :absolute}, non_neg_integer()) ::
+          non_neg_integer()
+  def calculate_relative_ttl(%{ttl: ttl, type: :absolute}, block_height) do
+    ttl - block_height
+  end
+
+  @spec tx_ttl_is_valid?(SignedTx.t(), non_neg_integer()) :: boolean
+  def tx_ttl_is_valid?(tx, block_height) do
+    case tx do
```
gspasov: Here you say in the `@spec` that tx is `SignedTx.t()` and then in the case you check whether it is some kind of Oracle tx. Am I not understanding the logic here, or the `@spec` is wrong?

apps/aecore/lib/aecore/oracle/oracle.ex
```diff
@@ -0,0 +1,293 @@
...
+                                                                              response_height_included
+                                                                          }},
+                                                                         acc ->
+      query_absolute_ttl =
+        calculate_absolute_ttl(
+          query.query_ttl,
+          query_height_included
+        )
+
+      query_has_expired = query_absolute_ttl <= block_height && response == nil
```
gspasov: Can you explain this line?

apps/aecore/lib/aecore/structures/spend_tx.ex
```diff
@@ -134,5 +141,25 @@ defmodule Aecore.Structures.SpendTx do
     Map.put(account_state, :balance, new_balance)
   end
 
+  @spec is_minimum_fee_met?(SignedTx.t(), :miner | :pool | :validation) :: boolean()
+  def is_minimum_fee_met?(tx, identifier) do
```
gspasov: What is the function of the identifier. What is the difference between the `:pool` and the `:miner` identifiers

apps/aecore/lib/aecore/txs/pool/worker.ex
```diff
@@ -126,11 +123,38 @@ defmodule Aecore.Txs.Pool.Worker do
...
   def get_tx_size_bytes(tx) do
     tx |> :erlang.term_to_binary() |> :erlang.byte_size()
   end
 
+  @spec is_minimum_fee_met?(SignedTx.t(), :miner | :pool | :validation, non_neg_integer()) ::
+          boolean()
+  def is_minimum_fee_met?(tx, identifier, block_height \\ nil) do
+    case tx.data.payload do
+      %SpendTx{} ->
+        SpendTx.is_minimum_fee_met?(tx, identifier)
```
gspasov: If almost every child tx of DataTx is going to have the `is_minimum_fee_met?` why not have it as a callback function in the Transaction Behaviour and have a generic call to it. They this case would be one liner eg. `tx.data.payload.type.is_minimum_fee_met?(some generic arguments)`. Maybe the args could be: SpendTx and block_height, atm that will cover all the funcs.

apps/aeutil/lib/serialization.ex
```diff
@@ -18,30 +20,59 @@ defmodule Aeutil.Serialization do
...
 
   def tx(tx, :deserialize) do
     tx_data = tx["data"]
+
     data = DataTx.deserialize(tx_data)
+
+    updated_data =
+      case data do
+        %OracleQueryTxData{} ->
+          update_in(data, [:payload, :oracle_address], &Account.base58c_decode(&1))
```
gspasov: Same as the above, but about the docoding func

apps/aeutil/lib/serialization.ex
```diff
@@ -18,30 +20,59 @@ defmodule Aeutil.Serialization do
...
 
-  @spec tx(map(), :deserialize) :: SpendTx.t()
-  def tx(tx, :serialize), do: serialize_value(tx)
+  @spec tx(map(), :serialize | :deserialize) :: SignedTx.t()
+  def tx(tx, :serialize) do
+    case tx.data.payload do
+      %OracleQueryTxData{} ->
+        tx
+        |> serialize_value()
+        |> update_in(["data", "payload", "oracle_address"], &Account.base58c_encode(&1))
```
gspasov: Can't these updates with the encoded versions of the data be done in the serialize_value() func, where all the other encoding is done?

apps/aecore/lib/aecore/oracle/oracle.ex
d-velev: spec is wrong

apps/aecore/lib/aecore/structures/spend_tx.ex
d-velev: Spend tx minimum fee is calculated by a formula in which one of the constants can be set by the miner meaning that the calculation can be different in the 2 cases

apps/aecore/lib/aecore/txs/pool/worker.ex
d-velev: I think the cases differ too much for this to be useful, the SpendTx takes an identifier and none of the others besides the response need it, the SpendTx doesn't require block height and so on

apps/aeutil/lib/serialization.ex
d-velev: What do you mean?

gspasov: See the below comment. I meesed up the orientation

apps/aeutil/lib/serialization.ex
d-velev: sure

apps/aecore/lib/aecore/oracle/oracle.ex
```diff
@@ -0,0 +1,296 @@
...
+    end
+  end
+
+  @spec ttl_is_valid?(ttl()) :: boolean()
+  def ttl_is_valid?(ttl) do
+    case ttl do
+      %{ttl: ttl, type: :absolute} ->
+        ttl > 0
+
+      %{ttl: ttl, type: :relative} ->
```
gspasov: Can't this case and the first one be done in only one case, being that they have the same outcome? eg `%{ttl: ttl, type: _type} -> `

d-velev: If it's done like that you would be able to input something other than :absolute or :relative for types

gspasov: This is just a pattern match, checking whether the type is `:absolute` or `:relative` should be checked when creating/initializing an Oracle type transaction. At this stage it should be sure that the type is either `:abosolute` or `:relative`, right?

thepiwo:
@cytadela8 can you re-review?

apps/aecore/lib/aecore/oracle/oracle.ex
d-velev: This is called in the is_valid? functions

apps/aecore/lib/aecore/structures/oracle_response_tx_data.ex
```diff
@@ -0,0 +1,167 @@
...
+           block_height,
+           oracle_state
+         ) do
+      :ok ->
+        interaction_object = interaction_objects[tx.query_id]
+        query_fee = interaction_object.query.query_fee
+
+        new_senderount_state =
+          Map.get(accounts, sender, Account.empty())
+          |> deduct_fee(fee - query_fee)
```
cytadela8: (optional) I would change to:
 ```
|> Account.transaction_in(query_fee)
|> deduct_fee(fee)
```

File: apps/aecore/lib/aecore/structures/oracle_response_tx_data.ex [hunk 46]
```diff
@@ -0,0 +1,167 @@
...
+
+        updated_accounts_chainstate = Map.put(accounts, sender, new_senderount_state)
+
+        updated_interaction_objects =
+          Map.put(interaction_objects, tx.query_id, %{
+            interaction_object
+            | response: tx,
```
cytadela8: shouldn't it be `| response: tx.response,`?

apps/aecore/lib/aecore/structures/spend_tx.ex
```diff
@@ -44,6 +45,9 @@ defmodule Aecore.Structures.SpendTx do
   defstruct [:receiver, :amount, :version]
   use ExConstructor
 
+  @spec get_chain_state_name() :: Account.chain_state_name()
+  def get_chain_state_name(), do: :accounts
```
cytadela8: (important) This should be `nil` otherwise `process_chainstate` will receive `chainstate.accounts` twice and will supposed to be returning new accounts state twice. This passes tests currently due to the order of updating in `DataTx`.

apps/aecore/lib/aecore/structures/spend_tx.ex
```diff
@@ -134,5 +141,25 @@ defmodule Aecore.Structures.SpendTx do
...
+      bytes_per_token =
+        case identifier do
+          :pool ->
+            Application.get_env(:aecore, :tx_data)[:pool_fee_bytes_per_token]
+
+          :miner ->
+            Application.get_env(:aecore, :tx_data)[:miner_fee_bytes_per_token]
+        end
+
+      tx.data.fee >= Float.floor(tx_size_bytes / bytes_per_token)
```
cytadela8: (very optional) To avoid Floats you can change to `tx.data.fee * bytes_per_token >= tx_size_bytes`

apps/aecore/lib/aecore/txs/pool/worker.ex
cytadela8: I would still consider adjusting to add `is_minimum_fee_met?` to Transaction Behaviour. It would make Oracle transactions more self contained.

apps/aehttpclient/lib/client.ex
```diff
@@ -95,6 +95,20 @@ defmodule Aehttpclient.Client do
     get(uri <> "/tx_pool/#{acc}", :acc_txs)
   end
 
+  def post_query_to_oracle(tx, uri) do
```
cytadela8: Why do we treat query txs specially? Shouldn't oracle react to query when it's included in block?

apps/aehttpserver/lib/aehttpserver/web/router.ex
```diff
@@ -21,6 +21,9 @@ defmodule Aehttpserver.Web.Router do
     post("/new_block", BlockController, :new_block)
     get("/blocks", BlockController, :get_blocks)
     get("/raw_blocks", BlockController, :get_raw_blocks)
+    post("/oracle_query", OracleController, :oracle_query)
+    options("/oracle_query", OracleController, :options)
```
cytadela8: (important) `oracle_query` should be authorized.

README.md
```diff
@@ -221,3 +221,78 @@ the log can be found in the source folder under:`apps/aecore/logs`
...
+  relative - the object is removed when a block with height equal to the TTL + block height in which the transaction was included is mined). A TTL is defined like so - %{ttl: value, type: :relative | :absolute}
+
+**Registering an oracle:**
+  `Aecore.Oracle.Oracle.register(query_format, response_format, query_fee, fee, ttl)`
+  The query and response formats are treated as json schemas with which the queries and responses
+  are validated. The query fee is the minimum fee that will be required for queries made to the
+  oracle. If the oracle responds to the query on time, he will receive that fee as a compensation
+  for the response fee he had to pay.
+
+  example format schema -
```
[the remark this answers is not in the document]
d-velev: I've described what the TTL is at the start of the oracle usage part of the readme

apps/aecore/lib/aecore/structures/oracle_query_tx_data.ex
```diff
@@ -0,0 +1,234 @@
...
+
+  require Logger
+
+  @type tx_type_state :: ChainState.oracles()
+
+  @type id :: binary()
+
+  @type payload :: %{
+          oracle_address: Wallet.pubkey(),
+          query_data: any(),
```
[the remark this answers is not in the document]
d-velev: It can be anything, if the schema says that the data should be integer and not object then just an integer is valid

apps/aecore/lib/aecore/structures/oracle_query_tx_data.ex
```diff
@@ -0,0 +1,234 @@
+defmodule Aecore.Structures.OracleQueryTxData do
```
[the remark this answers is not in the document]
d-velev: Since I did these modules when the SpendTx was still named TxData

apps/aecore/lib/aecore/structures/oracle_query_tx_data.ex
```diff
@@ -0,0 +1,234 @@
...
+      !Map.has_key?(registered_oracles, tx.oracle_address) ->
+        {:error, "No oracle registered with that address"}
+
+      !Oracle.data_valid?(
+        registered_oracles[tx.oracle_address].tx.query_format,
+        tx.query_data
+      ) ->
+        {:error, "Invalid query data"}
+
+      fee < tx.query_fee < registered_oracles[tx.oracle_address].tx.query_fee ->
```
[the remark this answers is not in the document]
d-velev: I'm not sure why this is like this, it wasn't intended, should be just tx.query_fee < registered....

apps/aecore/lib/aecore/structures/oracle_response_tx_data.ex
d-velev: yes that will be better

apps/aehttpclient/lib/client.ex
d-velev: This is old code I forgot to remove

apps/aecore/lib/aecore/structures/spend_tx.ex
thepiwo: Good find, I proposed this to @d-velev but had some additional code to catch this. `:accounts` returned is the semantic correct behavior but it should be filtered.

https://github.com/aeternity/elixir-node/blob/GH-274/apps/aecore/lib/aecore/structures/data_tx.ex#L83-L88

apps/aecore/lib/aecore/structures/oracle_query_tx_data.ex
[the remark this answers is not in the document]
thepiwo: Rename to `Tx` would be good

apps/aecore/lib/aecore/structures/oracle_query_tx_data.ex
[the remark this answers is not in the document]
thepiwo: can't we reuse the json type, you introduced here? (which is any())

apps/aecore/config/prod.exs
```diff
@@ -104,7 +104,13 @@ config :aecore, :peers,
 config :aecore, :miner, resumed_by_default: true
 
 config :aecore, :tx_data,
-  miner_fee_bytes_per_token: 100,
-  pool_fee_bytes_per_token: 100
+  lock_time_coinbase: 10,
```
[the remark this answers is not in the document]
thepiwo: we did, noticed this as well

d-velev:
> or start a new line before fn {query_id, ... 
This is the way the formatter does it, it wont let me change it.  About the is_minimum_fee_met? function - if I want to add is_minimum_fee_met? as a callback I have to pass the whole signedtx to the spendtx one and just the data to the others, that means I have to do a check by type and also pass a redundant parameter when it's a spendtx, also only 2 of the 5 transactions need a identifier so it will be inefficient just for the sake of abstraction

thepiwo:
@cytadela8 what do you think?

apps/aecore/lib/aecore/structures/data_tx.ex
```diff
@@ -72,7 +72,13 @@ defmodule Aecore.Structures.DataTx do
           ChainState.chainstate()
   def process_chainstate!(%DataTx{} = tx, chainstate, block_height) do
     accounts_state = chainstate.accounts
-    tx_type_state = Map.get(chainstate, tx.type.get_chain_state_name(), %{})
+
+    tx_type_state =
+      if(tx.type == SpendTx) do
```
cytadela8: Why not just return `nil` in `SpendTx.get_chain_state_name()`? It would work as expected.

cytadela8: Why not just return `nil` in `SpendTx.get_chain_state_name()`? It would work as expected.

apps/aecore/lib/aecore/structures/oracle_registration_tx.ex
```diff
@@ -81,73 +81,65 @@ defmodule Aecore.Structures.OracleRegistrationTxData do
...
         ) :: :ok | {:error, String.t()}
-  def preprocess_check(tx, sender, account_state, fee, nonce, block_height, registered_oracles) do
+  def preprocess_check!(tx, sender, account_state, fee, block_height, registered_oracles) do
     cond do
       account_state.balance - fee < 0 ->
-        {:error, "Negative balance"}
-
-      account_state.nonce >= nonce ->
-        {:error, "Nonce too small"}
+        throw({:error, "Negative balance"})
```
cytadela8: Why the nonce validation was removed?

apps/aecore/lib/aecore/structures/oracle_response_tx.ex
```diff
@@ -52,92 +52,85 @@ defmodule Aecore.Structures.OracleResponseTxData do
...
         registered_oracles: registered_oracles,
         interaction_objects: interaction_objects
       }) do
     cond do
       account_state.balance - fee < 0 ->
-        {:error, "Negative balance"}
-
-      account_state.nonce >= nonce ->
-        {:error, "Nonce too small"}
+        throw({:error, "Negative balance"})
```
cytadela8: Why nonce validation was removed also here?

cytadela8:
The nonce validation was moved to DataTx, but it's applied inside different types. It's good for merge.

apps/aecore/lib/aecore/structures/oracle_registration_tx.ex
cytadela8: It's in DataTx.

apps/aecore/lib/aecore/structures/oracle_response_tx.ex
cytadela8: It's in DataTx.

PR #214, opened 2018-02-09
[GH-#209] Serialize data with msgpack for binary representations

### LINKED ISSUE #209: serialize data with msgpack for binary representations
serialize data for hashing and signing with msgpack, totally remove `term_to_binary` to ensure compability with other languages

d-velev:
closes #209


apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -79,7 +80,7 @@ defmodule Aecore.Chain.BlockValidation do
 
   @spec block_header_hash(Header.t) :: binary()
   def block_header_hash(%Header{} = header) do
-    block_header_bin = :erlang.term_to_binary(header)
+    block_header_bin = Serialization.term_to_msgpack(header)
```
thepiwo: maybe rename to `pack_binary`

thepiwo:
can you review?
@nedelcho-delchev-tues 
@cheezus1 
@gspasov

apps/aeutil/lib/serialization.ex
```diff
@@ -57,4 +57,19 @@ defmodule Aeutil.Serialization do
...
+        Map.from_struct(%{term | header: Map.from_struct(term.header)})
+      %SignedTx{} ->
+        Map.from_struct(%{term | data: Map.from_struct(term.data)})
+      %{__struct__: _} ->
+        Map.from_struct(term)
+      _ ->
+        term
+    end
+    |> Msgpax.pack!(iodata: false)
+  end
```
cytadela8: Can we get a deserialization function? It might be useful in future.

d-velev: I think this is hard to do right since you can pack almost anything, if we need it in the future we can do that depending on what we need to deserialize.

PR #220, opened 2018-02-14
[GH-#219] update block candidate every 500 nonces

### LINKED ISSUE #219: update block-candidate periodically while mining
goals:
 - have a more current timestamp in the block-header
 - include new transactions

thepiwo:
closes #219

### DIFF
--- a/apps/aecore/lib/aecore/miner/worker.ex
+++ b/apps/aecore/lib/aecore/miner/worker.ex
@@ -25,6 +25,7 @@ defmodule Aecore.Miner.Worker do
 
   @mersenne_prime 2147483647
   @coinbase_transaction_value 100
+  @new_candidate_nonce_count 500
 
   def start_link(_args) do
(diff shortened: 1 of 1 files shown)


apps/aecore/lib/aecore/miner/worker.ex
```diff
@@ -25,6 +25,7 @@ defmodule Aecore.Miner.Worker do
 
   @mersenne_prime 2147483647
   @coinbase_transaction_value 100
+  @new_candidate_nonce_count 500
```
cytadela8: I would put much smaller number here (~10). AFAIK one cuckoo execution is few seconds.

cytadela8: (We might even be able for now to allow ourselves to regenerate every execution without much overhead)

thepiwo: in my testing the 500 was reached really quick ~10sec-20sec in our current configuration, lets keep it at 500 now, can change this later

Issue #210, opened 2018-02-05
represent hashes and accounts in json as bech32 with prefixes


thepiwo:
change hex representation of hashes to bech32 with prefix

this could look like this:
`Bech32.encode("tx", tx_hash_binary |> Aeutil.Bits.extract5)`
using https://github.com/stampery/elixir-bip0173/

`def extract5(str) when is_binary(str), do: for <<x :: 5 <- str >>, do: x`

thepiwo:
wait with this for confirmation if we actually want to use base58check or maybe bech32

nedelcho-delchev-tues:
Why would we want to use bech32, what would we gain from it ?

thepiwo:
It is considered more modern and if we make the effort in changing out serialization, we can just use this.

general motivation: https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#motivation

thepiwo:
I updated the issue to convert to bech32, but we should wait to after the elixir-wallet integration.
I have not found a way yet to convert the 5-bit array [0-31] from bech32 decoding to binary yet

thepiwo:
https://www.youtube.com/watch?v=NqiN9VFE4CU

thepiwo:
closed by https://github.com/aeternity/elixir-research/pull/225

PR #225, opened 2018-02-15
[GH-#210] Represent hashes as Bech32 with a prefix and signatures as Base64

d-velev:
Address serialization to be done in the wallet


thepiwo:
@nedelcho-delchev-tues @cheezus1 @meivantodorov @gspasov can you review?

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -62,7 +62,7 @@ defmodule Aecore.Peers.Worker do
   def genesis_block_header_hash() do
     Block.genesis_block().header
     |> BlockValidation.block_header_hash()
-    |> Base.encode16()
+    |> Header.bech32_encode()
```
cytadela8: For internal use I would leave binary representation.

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -58,7 +60,7 @@ defmodule Aecore.Chain.Worker do
 
   @spec get_block_by_hex_hash(String.t()) :: Block.t()
   def get_block_by_hex_hash(hash) do
```
cytadela8: We should also rename functions and variable names with "hex" in them if they are no longer hex. You can potentially remove some of such functions and variables. I would suggest using binary representation internally and bech for UI (including API and logs).

thepiwo:
@cheezus1 @gspasov can you please review?

apps/aecore/lib/aecore/peers/worker.ex
d-velev: We use it to compare the genesis hash we get from client info, I think this is fine.

thepiwo: better to compare using binary, as we use binary representation internally

d-velev:
As I said, serialization of addresses will be done in the wallet.

thepiwo:
@d-velev can you resolve conflicts and merge?

thepiwo:
I think you can merge the PR as well

PR #227, opened 2018-02-16
[GH-#223] Added timestamp validations

### LINKED ISSUE #223: validate block header timestamps
validate block timestamp
- [ ] not more than 1 hour in the future of current system unixtime
- [ ] not earlier than median timestamp of the last 10 blocks

Artur64:
Closes #223


apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -176,4 +179,28 @@ defmodule Aecore.Chain.BlockValidation do
   defp check_correct_height?(new_block, previous_block) do
     previous_block.header.height + 1 == new_block.header.height
   end
+
+  @spec valid_header_timestamp(Block.t()) :: boolean()
+  defp valid_header_timestamp(%Block{header: new_block_header}) do
+    case new_block_header.timestamp <= :os.system_time() + 3_600_000 do
+      true ->
+        top_block = Chain.top_block()
```
thepiwo: does this work? we had issues with calling the chain for fetching blocks to calculate the next difficulty and passed `blocks_for_difficulty_calculation`

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -176,4 +179,28 @@ defmodule Aecore.Chain.BlockValidation do
   defp check_correct_height?(new_block, previous_block) do
     previous_block.header.height + 1 == new_block.header.height
   end
+
+  @spec valid_header_timestamp(Block.t()) :: boolean()
+  defp valid_header_timestamp(%Block{header: new_block_header}) do
+    case new_block_header.timestamp <= :os.system_time() + 3_600_000 do
```
thepiwo: make `3_600_000` a constant

https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants

thepiwo: use elixir style timestamp `System.system_time(:milliseconds)` instead of erlang `:os.system_time()`

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -176,4 +179,28 @@ defmodule Aecore.Chain.BlockValidation do
...
+      true ->
+        top_block = Chain.top_block()
+        last_blocks = Chain.get_blocks(top_block, 10)
```
thepiwo: make `10` constant as well

https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -176,4 +179,28 @@ defmodule Aecore.Chain.BlockValidation do
...
+
+        last_blocks_timestamps =
+          Enum.reduce(last_blocks, [], fn block, acc -> acc = acc ++ [block.header.timestamp] end)
```
d-velev: Enum.map or a for is better here since we don't care about the value of the accumulator from the last iteration, we just want to get a list of all timestamps. Also, the accumulator for a reduce is the last data returned in the function so there is no need to do `acc = acc ++ ...`, `acc ++ [block.header.timestamp]` does the job.

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -176,4 +179,28 @@ defmodule Aecore.Chain.BlockValidation do
...
+
+        average = Enum.sum(last_blocks) / 10
+
+        if new_block_header.timestamp > average do
```
d-velev: `new_block_header.timestamp > average` returns true or false on its own so this if isn't needed

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -176,4 +179,28 @@ defmodule Aecore.Chain.BlockValidation do
...
+  defp valid_header_timestamp(%Block{header: new_block_header}) do
+    case new_block_header.timestamp <= :os.system_time() + 3_600_000 do
+      true ->
+        top_block = Chain.top_block()
+        last_blocks = Chain.get_blocks(top_block, 10)
+
+        last_blocks_timestamps =
+          Enum.reduce(last_blocks, [], fn block, acc -> acc = acc ++ [block.header.timestamp] end)
+
+        average = Enum.sum(last_blocks) / 10
```
d-velev: Shouldn't the sum be divided by the number of elements in last_blocks since if we have less than 10 blocks in the chain, this wouldn't be the average?

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -176,4 +183,23 @@ defmodule Aecore.Chain.BlockValidation do
   defp check_correct_height?(new_block, previous_block) do
     previous_block.header.height + 1 == new_block.header.height
   end
+
+  @spec valid_header_timestamp(Block.t()) :: boolean()
+  defp valid_header_timestamp(%Block{header: new_block_header}) do
```
cytadela8: As this returns `boolean` add `?` at the end of function name.

apps/aecore/test/aecore_validation_test.exs
```diff
@@ -15,18 +15,13 @@ defmodule AecoreValidationTest do
...
   @tag :validation
-  test "validate new block" do
+  test "validate block header timestamp" do
     new_block = get_new_block()
     prev_block = get_prev_block()
 
     blocks_for_difficulty_calculation = [new_block, prev_block]
-    _ = BlockValidation.calculate_and_validate_block!(
-      new_block, prev_block, get_chain_state(), blocks_for_difficulty_calculation)
-    wrong_height_block = %Block{new_block | header: %Header{new_block.header | height: 3}}
```
thepiwo: can you add the test, testing "Incorrect height" back

cytadela8:
Fix CI.

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -45,6 +45,9 @@ defmodule Aecore.Chain.BlockValidation do
       !(is_genesis || check_correct_height?(new_block, previous_block)) ->
         throw({:error, "Incorrect height"})
 
+      !valid_header_timestamp?(new_block) ->
```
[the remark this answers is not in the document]
Artur64: Because hash check will always fail if i will change timestamp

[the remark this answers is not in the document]
thepiwo: but this seems to be an issue with your test, no reason to change the code. I propose you write the opposite test case as well and make both cases work, either one fails first, if one is incorrect.

thepiwo:
@Artur64 if you can't get the tests to work as imagined, we can just merge as is.

thepiwo:
@Artur64 are you still working on this? Otherwise we will merge...

thepiwo:
@Artur64 can you finish this today, otherwise we will merge, as is

PR #229, opened 2018-02-16
[GH-#216] Lazy read from rocksdb

### LINKED ISSUE #216: lazy read from rocksdb what blocks are needed
only keep the last 100 blocks in memory

meivantodorov:
### What this PR does:

- Keep only 100 blocks in memory;
- On start-up load 100 blocks from rocksdb;
- On batch blocks requests first we check in memory if the block is not found then we check in rocksdb;

Closes #216

### DIFF
--- a/apps/aecore/lib/aecore/chain/worker.ex
+++ b/apps/aecore/lib/aecore/chain/worker.ex
@@ -273,14 +273,23 @@ defmodule Aecore.Chain.Worker do
 
   defp get_blocks(blocks_acc, next_block_hash, final_block_hash, count) do
     if next_block_hash != final_block_hash && count > 0 do
-      case(GenServer.call(__MODULE__, {:get_block, next_block_hash})) do
-        {:error, _} -> blocks_acc
-        block ->
-          updated_blocks_acc = [block | blocks_acc]
-          prev_block_hash = block.header.prev_hash
-          next_count = count - 1
+      next_attempt = fn(block) -> get_blocks([block | blocks_acc],
+                                             block.header.prev_hash,
+                                             final_block_hash, count - 1) end
 
-          get_blocks(updated_blocks_acc, prev_block_hash, final_block_hash, next_count)
+      ## At first we are making attempt to get the block from the chain state.
+      ## If there is no such block then we check into the db.
+      case (GenServer.call(__MODULE__, {:get_block, next_block_hash})) do
+        {:error, _} ->
+          case Persistence.get_block_by_hash(next_block_hash) do
+            {:ok, block} ->
+              next_attempt.(block)
+            _ ->
+              blocks_acc
+          end
+
+        block ->
+          next_attempt.(block)
       end
     else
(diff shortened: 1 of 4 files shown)


apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -165,15 +165,28 @@ defmodule Aecore.Chain.Worker do
     Enum.each(new_block.txs, fn(tx) -> Pool.remove_transaction(tx) end)
     new_block_hash = BlockValidation.block_header_hash(new_block.header)
 
-    updated_blocks_map = Map.put(blocks_map, new_block_hash, new_block)
+    updated_blocks_map  = Map.put(blocks_map, new_block_hash, new_block)
+    hundred_blocks_map  =
+    if Kernel.map_size(updated_blocks_map) > 100 do
```
thepiwo: make the number configurable 

https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants

thepiwo:
can you keep the genesis block always in memory? Each `GET /info` call would try to fetch it from disk

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -165,15 +165,28 @@ defmodule Aecore.Chain.Worker do
...
     new_block_hash = BlockValidation.block_header_hash(new_block.header)
 
-    updated_blocks_map = Map.put(blocks_map, new_block_hash, new_block)
+    updated_blocks_map  = Map.put(blocks_map, new_block_hash, new_block)
+    hundred_blocks_map  =
+    if Kernel.map_size(updated_blocks_map) > number_of_blocks_in_memory() do
+      [{_,b} | sorted_blocks] =
+        Enum.sort(updated_blocks_map,
+          fn({_,b1}, {_,b2}) ->
+            b1.header.timestamp < b2.header.timestamp
```
thepiwo: sort by height, timestamp is not a trustable metric for sorting

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -165,15 +165,28 @@ defmodule Aecore.Chain.Worker do
...
+          end)
+      Logger.info("Block ##{b.header.height} has been removed from memory")
+      Enum.into(sorted_blocks, %{})
```
thepiwo: looks like the maximum of 1 block is removed from memory, maybe it would be better to sort and then use `Enum.drop` or similar

meivantodorov: I'm dropping the first one from the sorted list by using pattern matching.

thepiwo: okay, is good for now

meivantodorov: Actually I'm dropping the second element now in order to keep the genesis block as well.

PR #236, opened 2018-02-21
[GH-234] Added fast get_block_by_height to Chain

### LINKED ISSUE #234: Fast get block by height
Currently accessing a block by height requires traversing possibly whole chain form top block. This can be optimized by storing references to previous blocks in each block. It's possible to compromise between storing all references and long access time to achieve O(log n) references and O(log n) access time.

cytadela8:
- Added refs for n-th prev blocks for n beeing a power of 2
- Refactored Chain.Worker
- Added function to get block by height using the refs

Work-In-Progress

closes #234


thepiwo:
this seems to add a lot of complexity to existing code, what do we need `block_by_height` for at all?

cytadela8:
It can be used for block sync so it can be done from bottom to the top. (This has several advantages.)
I also used that in my Voting implementation to reference old balances of accounts.
As for the complexity: a lot of changes are refactoring and aside from those I add one calculation code to add_validated_block and the block_by_height function.

thepiwo:
Okay, we will refactor the sync to do headers first, so we solve this problem in the future.

cytadela8:
Ok, so we might close this for now and only merge if it will be useful in the future.

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -82,6 +98,29 @@ defmodule Aecore.Chain.Worker do
...
+  @spec longest_blocks_chain() :: list(Block.t())
+  def longest_blocks_chain() do
+    get_blocks(top_block_hash(), top_height() + 1)
+  end
+ 
+  def get_chain_state_by_height(height, chain_hash \\ nil) do
+    get_block_info_by_height(height, chain_hash).chain_state
+  end
+
+  @spec chain_state(binary()) :: ChainState.account_chainstate()
```
thepiwo: what does this function give? Chain-State at height? If so, how is it computed efficiently?

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -158,25 +179,39 @@ defmodule Aecore.Chain.Worker do
...
+                    top_height: top_height} = state) do
     new_block_txs_index = calculate_block_acc_txs_info(new_block)
     new_txs_index = update_txs_index(txs_index, new_block_txs_index)
-    Enum.each(new_block.txs, fn(tx) -> Pool.remove_transaction(tx) end)
     new_block_hash = BlockValidation.block_header_hash(new_block.header)
-
-    updated_blocks_map = Map.put(blocks_map, new_block_hash, new_block)
-    updated_chain_states = Map.put(chain_states, new_block_hash, new_chain_state)
-
+    # refs_list is generated so it contains n-th prev blocks for n-s beeing a power of two. So for chain A<-B<-C<-D<-E<-F<-G<-H. H refs will be [G,F,D,A]. This allows for log n findning of block with given height.
```
thepiwo: why is this change needed? the previous code was way simpler

thepiwo:
Okay, lets try to optimize this PR and then merge it. We can include an endpoint for the explorer. But I think the best approach to get block by height is to keep an index `%{height => hash}`

apps/aecore/lib/aecore/chain/worker.ex
cytadela8: Chainstate at hash. It's not a new function.

apps/aecore/lib/aecore/chain/worker.ex
cytadela8: I add the `refs` for every block to enable me to traverse chain backwards very fast.

cytadela8:
@thepiwo The problem is how will we be able to distinguish between forks?

I did following changes:
- wrap chainstates and blocks in one (unnamed) structure - block_info, this way asociating new data with block is easy and we have only one map.
- add the `refs` table to every block. This table contains hashes of prev_block, 2nd prev, 4th prev, 8th prev and so on... (for every n-th such that n is a power of 2)
- using the `refs` we can easily traverse chain backwards. (this is what the `get_block_info_by_height` does) Let's say the top block of the longest chain (or of the fork we would like to traverse) has height 11011011 (represented in binary). And we would like to get the block with height 10010101. To do that we do the following:
   - get 2nd prev of top block, we get block with height 11011001
   - get 4th prev of block with height 11011001, we get a block with height 11010101
   - get 64th prev of block with height 11010101, we get the target block with height 10010101
One can _easily_ see we will do at most log2 (n) steps. So this is very fast :)

cytadela8:
What one could argue is that we could recalculate the %{height => hash} map every time some forking happens, but this solution is more stable and can explore forks without any problems.

cytadela8:
(This solution works for any tree and can be for example used to calculate Lowest Common Ancestor - that is the place when the chain forked. The reason I coded this solution is that it was known to me previously, It's also available in detail on Wikipedia, but for some reason only on polish version of the article, English version contains other solutions.)

apps/aecore/lib/aecore/chain/worker.ex
d-velev: Can you explain what's going on here? The usage of arrays looks weird, if working with indexes is really needed, wouldn't an indexed map do the job?

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -17,33 +18,36 @@ defmodule Aecore.Chain.Worker do
...
   def start_link(_args) do
     GenServer.start_link(__MODULE__, {}, name: __MODULE__)
   end
 
   def init(_) do
     genesis_block_hash = BlockValidation.block_header_hash(Block.genesis_block().header)
-    genesis_block_map = %{genesis_block_hash => Block.genesis_block()}
     genesis_chain_state = ChainState.calculate_and_validate_chain_state!(Block.genesis_block().txs, %{}, 0)
-    chain_states = %{genesis_block_hash => genesis_chain_state}
+    blocks_map = %{genesis_block_hash =>
```
thepiwo: the name `blocks_map` is not fitting

apps/aecore/lib/aecore/chain/worker.ex
thepiwo: I didn't notice we already had this...

thepiwo:
is it necessary to use `:array` erlang-style functions?

thepiwo:
I propose to merge this as it is soon, but we should remove the chain_state at hash functions and only keep snapshots, e.g. each 10000 blocks to ease recalculation when forking

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -273,4 +309,31 @@ defmodule Aecore.Chain.Worker do
...
+    else
+      blocks_map[get_nth_prev_hash(n, 0, begin_hash, blocks_map)]
+    end
+  end
+  
+  defp get_nth_prev_hash(0, _i, hash, blocks_map) do
+    hash
+  end
+
+  defp get_nth_prev_hash(n, i, hash, blocks_map) do
```
thepiwo: write documentation to explain how this function works, or rewrite to be readable (no bit-magic)

apps/aecore/lib/aecore/chain/worker.ex
cytadela8: What do you suggest: "block_info_map", "block_data_map"?

apps/aecore/lib/aecore/chain/worker.ex
cytadela8: ok

cytadela8:
> is it necessary to use `:array` erlang-style functions?

I can use Enum.get and lists. In theory that will be slower, becouse it requires going through the whole list (of 30 elements so that's not much) and arrays are constant time access.

I suppose the price for this simplification is low so I will change that.

cytadela8:
As for the snapshots: that would be out of scope of this PR. I can work on that, but that's a separate change.

thepiwo:
yes sure, snapshot will be a separate task

apps/aecore/lib/aecore/chain/worker.ex
thepiwo: will you still work on this?

cytadela8:
@thepiwo It should be ready now. Because I wanted to save the refs in Rox and load them on start, I added a new rox "family" that can be used also for storing headers. Separating headers from blocks and loading them on start should be simple now and should only require changes to `Aecore.Chain.Worker`.
I also experienced problems with loading Persistance data on start. I have to call `Kernel.send(Process.whereis(Aecore.Chain.Worker),:timeout)` to get the data loaded. I will investigate.

thepiwo:
wait for #263 to merge

thepiwo:
@cytadela8 ready to merge?

cytadela8:
@thepiwo Ready for merge. I changed few things in the way blocks are removed from memory and how such blocks are handled later so:
- there shouldn't be any errors when invoking any functions on removed blocks
- it works now as I intended.

cytadela8:
Let me go through the changes again and fix some small mistakes made in merging.

thepiwo:
feel free to merge, whenever

Issue #233, opened 2018-02-21
Address types (wallets, contracts, votings...)


cytadela8:
I would like to propose changes to chain_state to make it possible to treat transactions differently based on the `to_acc` address. This will make it possible to have just one transaction type. When an transaction is executed `to_acc` address controls the flow of execution. 
If `to_acc` address is:
- a contract - function for handling transactions to contracts is executed
- a wallet (normal account) - current function for adding balance is executed
- a special (constant/fixed/predefind) string - a special blockchain operation can be executed. for example: burning tokens, storing data, creation of a contract, creation of a voting...

_This is an completely opposite approach to adding many different transaction types. Instead we add different address types._

thepiwo:
can you explain the advantages further? What is the difference to https://github.com/aeternity/elixir-research/issues/203?

cytadela8:
So this is a different approach by instead of creating different transaction types, we use different object types in `chain_state`. One could have different transaction structure for transactions to contracts, voting, creating contracts, creating voting etc., but there is no need for that. Deciding what to do with a transaction can be done only depending on the `to_acc` address. My reasoning is: _Let's call all of the contracts, votings, accounts and other such structures BlockchainObjects._
- We already have to have different Elixir structures for different types of BlockchainObjects.
- We have to store BlockchainObjects somewhere. We could store more maps like "chain_state", but I consider it simpler to store everything in already existing `chain_state` map.
- When everything is stored in `chain_state` accessing information about different types of BlockchainObjects can be done the same way. There only needs to exist a function that returns the structure of BlockchainObject with a given address. (see #194 implementation, you access both accounts and votings with: Chain.chain_state()[address])
- Different "kinds" of transactions can be handled just by attaching different data to them - votings will expect different fields in the "data" part of transaction, contracts different etc. The advantages of this I can think of are:
    - any interface for creating transactions can make it possible to create any transaction by just leave giving free access to the data field to the user. 
    - we won't need multiple functions for creation different transaction objects, serializing them or any other interaction with them. 
    - this should make it easier for contracts to create different transaction types and (for example) for multi sig wallets to handle all kinds of transactions without understanding their "kinds".
- one problem with this idea is multi signature transactions. This can be handled by changing the current structure of transaction to have a list in `from_acc` and `signature` fields.

gspasov:
[...]

Issue #194, opened 2018-01-18
Extendable PoS Voting


cytadela8:
# Extendable PoS Voting

## Idea

We would like accounts to "vote", the weight of vote should (generally) depend on the balance. There are many different scenarios when that is useful (All kinds of online secure pools, blockchain Governance, DAOs). That's why it's important to find as extendable way of implementing voting. [Below](#solutions) I propose few systems with customisable formulas. The idea behind this prototype is to implement most of the scenarios natively for better performance then an implementation in contracts would achieve.

## Plan

From below set of solutions I'm going to implement OnChainFormula, but with predefined formulas implemented in Elixir. (Instead of specifying Formula code in Question, Question will specify which of implemented by me formulas to use)

- [x] Make it possible to create new types of txs (probably use some of the code from GH-172)
- [x] Add Question transaction that adds Voting related data to chainstate
- [ ] Create interfaces (including web) for pulling data from Votings
- [x] Add Vote transaction
- [x] Add different Formulas

## Possible addictions

- [ ] Find a way to delegate votes. (Probably user could be allowed to set some custom fields in his account state, but how to change formula? Should we move signature validation to formula? Maybe we could store in account-state not only info about balance, but also "voting\_power"?)

## Solutions

Below solutions can be implemented both natively (on blockchain level) or with a contract (if our contract language will allow everything required).

### OnChainFormula

Definitions:
- Voting = a process of supplying a global Answer to Requester after TimeFrame using Votes that affect VotingState according to Formula. Voting is started with Question
- TimeFrame = start block, end block pair. All blocks from block after start block to end block are considered "in"
- Requester = an Account, Contract or the Blockchain
- VotingState = a tuple-like structure representing current state of an ongoing voting.
- Answer = VotingState after TimeFrame end
- Vote = A transaction with tuple-like data that that alters VotingState with Formula. Only one vote per account is allowed. When Voting fee is taken (as normally) from current state of account, but to Formula the state after (at) start block is supplied.
- Formula = A non recursive function in a functional language to be deffined. It has 3 parameters: VotingState, Vote, AccountState. Returns: VotingState. Vote fee has to cover price of Formula calculation.
- Question = A transaction created by Requester specifying TimeFrame, Formula and initial VotingState. The fee has to cover VotingState storage cost.

### OffChainFormula

[...]

PR #248, opened 2018-03-01
[GH-194] Extendable Voting

cytadela8:
My solution to Voting for review. 

This implements `OnChainFormula` as described in #194.

This uses changes from #234 and #233.

Please refer to `README.md` for how-to.

**DO NOT MERGE**


cytadela8:
This experimental implementation of voting is abandoned; There is no plan to merge this implementation.

Following changes to this PR were proposed:
- Instead of applying vote to voting state when it's included in block, all votes should be saved and voting result should be calculated on `end_height`.
Current implementation requires access to old ChainStates.
- Transaction data should be an Elixir Struct.
Current implementation limits tx.data field validation.
- HTTP endpoints should be added and other UX related code.

PR #251, opened 2018-03-02
[GH-#230] Change account identifier in api to bech32 encoding

### LINKED ISSUE #230: change account identifier in api to bech32 encoding
wait for #201 and #210 to be merged

internally the compressed public key should be used (33 bytes)
and externally the bech32 encoded version of this.

d-velev:
closes #230


apps/aecore/lib/aecore/wallet/worker.ex
```diff
@@ -120,18 +120,29 @@ defmodule Aecore.Wallet.Worker do
   ## Server Callbacks
 
   def handle_call({:get_pub_key, {derivation_path, password, network}}, _from, %{pubkey: nil} = state) do
-    {:ok, pub_key} =
-      get_aewallet_dir()
-      |> get_file_name()
-      |> Wallet.get_public_key(password, network: network)
+    pub_key = if derivation_path == "" do
```
gspasov: I think that for this project it's already decided to put the `if statement` on a new line if we are going to pass its value to a varialbe. This occurs more than once in the following code.

thepiwo:
@cheezus1 @nedelcho-delchev-tues can you review this as well?

PR #263, opened 2018-03-12
[GH-#203] Transaction building structure refactor

### LINKED ISSUE #203: transaction data refactoring
# The scope of this issue changed

will follow this refactoring proposal: https://github.com/aeternity/elixir-research/issues/203#issuecomment-367045472

.

.

.

.

.

# Deprecated description

## plain data transaction

- custom signed_tx
- validate only signature, fees and nonce

```
%SignedTx{
  data: %DataTx{
              type: String.t(),
              payload: any type,
              from_acc: binary(),
              fee: integer(),
              nonce: non_neg_integer(),
            },
  signature: binary()
}
```

## proof of existence prototype

- type: ProofOfExistenceTx
- payload: hash of any file/data, base-encoded

------

 - in the future this can be used for other aepps transactions that need to be persisted in the blockchain
 - the plan is to merge this if ready, so #197 can depend on it

gspasov:
resolves #203 

# What is the idea of this Major Refactor:
The idea is to make a generic approach/functionality for all the transaction and make it easy for adding new types of transaction to the blockchain without making much changes around the code. 
Moreover to have all the specific functionality/validation for each new transaction in its own module and not all around the code. 
This will make the code more clean and tidy and adding a new transaction type will become easier.

### Working on `SignedTx `
- `is_valid?()` now is more generalized and makes for an easier validation of the different types of child transactions inside the `DataTx`

### Working on the different Transaction modules
- Made the `DataTx` module that is going to serve as a wrapper transaction module for all the transactions
- It's structure is the following
```elixir

	@typedoc "Transaction module name eg. Elixir.Aecore.Structures.SpendTx"
	@type tx_type :: SpendTx | atom()
	
	@typedoc "The structure of the diffent types of transaction that could be added to the blockchain"
	@type payload :: %SpendTx{} | map()

	%DataTx{type: tx_type(),
		payload: map(),
		from_acc: binary(),
		fee: non_neg_int(),
		nonce: non_neg_int()}
```
  - This structure allows all the transactions that will be created in the future to be under this generic `DataTx` module;

### Created a Behaviour for all the child transactions of the `DataTx` module:
  - This module states all the necessary callback functions that any child transaction should implement so that it works correctly on the blockchain

```elixir
   @typedoc "Arbitrary map holding all the specific elements required                
   by the specified transaction type"                                                
   @type payload :: map()                                                            
                                                                                     
   @typedoc "Structure of a custom transaction"                                      
   @type tx_types :: SpendTx.t()                                                     
                                                                                     
   @typedoc "Reason for the error"                                                   
   @type reason :: String.t()                                                        
                                                                                     
   @typedoc "Structure that holds specific transaction info in the chainstate"       
   @type tx_type_state() :: map()                                                    
```                                                                                     
- `Callbacks`:

- This callback us used as a builder to the specific transaction                                                                       
```elixir                                                                                     
   @callback init(payload()) :: tx_types()                                                                                                                     
```  

- This callback is used to check whether the data given to this transaction is valid
```elixir 
@callback is_valid?(tx_types()) :: boolean()                                      
```

- This callback is used to process the transaction (update the account_state and tx_type_state if there is for this transaction, for example SpendTx doesn't have a specific type state)
```elixir                                                                         
   @callback process_chainstate!(tx_types(),                                         
                                 Wallet.pubkey(),                                    
                                 fee :: non_neg_integer(),                           
                                 nonce :: non_neg_integer(),                         
                                 block_height :: non_neg_integer(),                  
                                 Account.t(),                                        
                                 tx_type_state()) :: {Account.t(), tx_type_state()}  

```

- This callback is used to do all the preprocess work and checks before the transaction is executed and the chainstate is changed
```elixir
 @callback preprocess_check(ChainState.account(),                             
                            fee :: non_neg_integer(),                         
                            nonce :: non_neg_integer(),                       
                            block_height :: non_neg_integer(),                
                            tx_type_state :: map()) :: :ok | {:error, reason} 
```

- This callback is used to deduct the fee of the transaction from the `from_acc` (initializer of the Tx)
```elixir                                                                              
 @callback deduct_fee(ChainState.account(),                                   
                      fee :: non_neg_integer(),                               
                      nonce :: non_neg_integer()) :: ChainState.account()     
```

### Changed `SpendTx` module to be a child module of the `DataTx` (as all other new transactions will be):
  - Now it uses the Transaction Behaviour and overrides all the callbacks needed for a transaction
  - All the related/specific functionality for this type of transaction has been moved to its own module (initialization, making a spend tx, etc.)

- Structure:
```elixir
%SpendTx{                  
   to_acc: binary(),            
   value: non_neg_integer(),           
   lock_time_block: non_neg_integer()  
 }                                     

```

### Working on the `Chain.Worker` modules:
- Changed the `chainstate` so that now it holds the accounts in seperate key `:accounts`
- Also it will be possible to add new keys with data specific for the different transaction modules where the key should be the module name eg: `Elixir.Aecore.Structures.Oracles`

- Structure:
```elixir
%{:accounts => %{a_pub_key => %Account{balance: 3,                                
                                              nonce: 100,                                
                                              locked: []},    
                 b_pub_key => %Account{balance: 5,                                
                                              nonce: 1,                                  
                                              locked: []},
                 c_pub_key => %Account{balance: 4,                                
                                              nonce: 1,                                  
                                              locked: []}}
```

And it could possibly change to:
```elixir
%{:accounts => %{a_pub_key => %Account{balance: 3,                                
                                              nonce: 100,                                
                                              locked: []},    
                 b_pub_key => %Account{balance: 5,                                
                                              nonce: 1,                                  
                                              locked: []},
                 c_pub_key => %Account{balance: 4,                                
                                              nonce: 1,                                  
                                              locked: []}
   Elixir.Aecore.Structures.Oracles => %{:devieces => [],
                                                                :trusted => [],
                                                                :comands => %{}}}
```

### Working on the `BlockValidation` and `ChainState` modules:
- Made the code be more module specific, so we changed the place of some of the functions where it made more sense and tried to utilize them as best as possible
- Removed some repeated code/functionality between the two modules


apps/aecore/config/config.exs
```diff
@@ -53,7 +53,7 @@ config :aecore, :aewallet,
   path: Path.absname(aewallet_path)
 
 config :aecore, :persistence,
-  path: Path.absname(persistence_path),
+  path: persistence_path |> Path.absname() |> Path.join("//"),
```
thepiwo: According to the Elixir documentation "Note that any trailing slash is removed when joining."

apps/aecore/lib/aecore/keys/worker.ex
```diff
@@ -0,0 +1,359 @@
+defmodule Aecore.Keys.Worker do
```
thepiwo: why is this added back?

thepiwo: @gspasov shouldn't the wallet be used?

apps/aecore/config/config.exs
gspasov: This is done so that the path is created correctly. With 2 slashes we get the desired result.

apps/aecore/lib/aecore/keys/worker.ex
gspasov: It has slipped through us for some reason

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -5,120 +5,129 @@ defmodule Aecore.Chain.ChainState do
...
+  @typedoc "State of an account"
+  @type acc_state() :: %{balance: integer(),
+                         locked: [%{amount: integer(), block: integer()}],
+                         nonce: integer()}
+
+  @typedoc "Structure of the accounts"
+  @type accounts() :: %{pubkey() => acc_state()}
+
+  @typedoc "Structure of the chainstate"
+  @type chainstate() :: map()
```
thepiwo: this can be more specific than `map()` e.g. 
`@type chain_state_types() :: :accounts`
`@type chainstate() :: %{chain_state_types() => accounts()}`
(not sure if correct example)

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -28,7 +30,9 @@ defmodule Aecore.Chain.Worker do
   def init(_) do
     genesis_block_hash = BlockValidation.block_header_hash(Block.genesis_block().header)
     genesis_block_map = %{genesis_block_hash => Block.genesis_block()}
-    genesis_chain_state = ChainState.calculate_and_validate_chain_state!(Block.genesis_block().txs, %{}, 0)
+    genesis_chain_state =
+      ChainState.calculate_and_validate_chain_state!(Block.genesis_block().txs, %{accounts: %{}}, 0)
```
thepiwo: maybe make a `ChainState.empty()` funtion to return a base structure `%{accounts: %{}}` to reduce possible errors by changing the structure

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -143,6 +149,11 @@ defmodule Aecore.Chain.Worker do
 
   ## Server side
 
+  def handle_call(:clear_state, _from, state) do
+    {:ok, new_state, _} = init("")
```
thepiwo: looks weird with `""` I know it doesn't matter but maybe choose something else

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -233,7 +244,7 @@ defmodule Aecore.Chain.Worker do
     chain_states =
       case Persistence.get_all_accounts_chain_states() do
         chain_states when chain_states == %{} -> state.chain_states
-        chain_states -> %{top_hash => chain_states}
+        chain_states -> %{top_hash => chain_states["chain_state"]}
```
thepiwo: looks weird, can you explain why this call is needed? Is `chain_states` the correct semantical name still? If so why does it have a `"chain_state"` item?

apps/aecore/lib/aecore/structures/account.ex
```diff
@@ -31,63 +34,25 @@ defmodule Aecore.Structures.Account do
...
-      %Account{account | locked: new_locked}
-    else
-      new_balance = account.balance + tx.value
-      if new_balance < 0 do
-        throw {:error, "Negative balance"}
-      end
-      %Account{account | balance: new_balance}
-    end
-  end
+  @spec update_locked(Account.t(), integer()) :: Account.t()
```
thepiwo: I preferred having the `tx_in!` and `tx_out!` functions here, otherwise this file makes no sense and is a good abstraction to not have all code in the chain state calculation. Should be easy to adjust with your implementation.

apps/aecore/lib/aecore/structures/data_tx.ex
```diff
@@ -0,0 +1,88 @@
...
+  def process_chainstate(%__MODULE__{} = tx, block_height, chainstate) do
+    account_state = chainstate.accounts
+    subdomain_chainstate = Map.get(chainstate, tx.type, %{})
+
+    new_accounts_state =
+      tx.payload
+      |> tx.type.init()
+      |> tx.type.process_chainstate!(tx.from_acc, tx.fee, tx.nonce, block_height,
+                                account_state, subdomain_chainstate)
+      Map.put(chainstate, :accounts, new_accounts_state) ## TODO return the subdomain_chainstate as well
```
thepiwo: is this `TODO` still open?

apps/aecore/lib/aecore/structures/data_tx.ex
```diff
@@ -0,0 +1,88 @@
...
+
+  alias Aecore.Keys.Worker, as: Keys
+  alias Aeutil.Serialization
+  alias Aeutil.Parser
+
+  @typedoc "Reason for the error"
+  @type reason :: String.t()
+
+  @typedoc "Structure of the main transaction wrapper"
+  @type t :: %__MODULE__{
```
thepiwo: why is `%__MODULE__` syntax used here?

thepiwo: multiple times in this file

apps/aecore/lib/aecore/structures/spend_tx.ex
```diff
@@ -3,43 +3,129 @@ defmodule Aecore.Structures.SpendTx do
...
+  @type payload :: %__MODULE__{} | map()
+
+  @typedoc "Reason for the error"
+  @type reason :: String.t()
+
+  @typedoc "Public key of the account"
+  @type pub_key() :: binary()
+
+  @typedoc "Structure that holds specific transaction info in the chainstate"
+  @type subdomain_chainstate() :: map()
```
thepiwo: describe `map()` more specific

apps/aecore/lib/aecore/structures/spend_tx.ex
```diff
@@ -3,43 +3,129 @@ defmodule Aecore.Structures.SpendTx do
...
+  alias Aecore.Structures.Account
+
+  @typedoc "Arbitrary structure data of a transaction"
+  @type payload :: %__MODULE__{} | map()
+
+  @typedoc "Reason for the error"
+  @type reason :: String.t()
+
+  @typedoc "Public key of the account"
+  @type pub_key() :: binary()
```
thepiwo: `@type pub_key()` is used in multiple places with different names `@type pubkey()` sometimes. Move it to the Wallet worker.

apps/aecore/lib/aecore/structures/transaction.ex
```diff
@@ -0,0 +1,76 @@
...
+  @type payload :: map()
+
+  @typedoc "Structure of a custom transaction"
+  @type tx_struct :: map()
+
+  @typedoc "Reason for the error"
+  @type reason :: String.t()
+
+  @typedoc "The public key of the account"
+  @type pub_key :: binary()
```
thepiwo: `@type pub_key()` is used in multiple places with different names `@type pubkey()` sometimes. Move it to the Wallet worker.

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -5,120 +5,129 @@ defmodule Aecore.Chain.ChainState do
...
+  alias Aecore.Structures.DataTx
   alias Aecore.Structures.Account
   alias Aeutil.Serialization
   alias Aeutil.Bits
 
   require Logger
 
-  @type account_chainstate() :: %{binary() => map()}
+  @typedoc "Public key representing an account"
+  @type pubkey() :: binary()
```
thepiwo: `@type pubkey()` is used in multiple places with different names `@type pub_key()` sometimes. Move it to the Wallet worker.

apps/aecore/lib/aecore/structures/transaction.ex
```diff
@@ -0,0 +1,76 @@
+defmodule Aecore.Structures.Transaction do
```
thepiwo: does this function as interface?

gspasov: Yes, it's functionality is as to an interface

apps/aecore/lib/aecore/structures/spend_tx.ex
gspasov: We typed it map(), because it could be any structure that the chainstate may hold, for a specific transaction. For example it can hold the chainstate of an oracle or of a votingTx and we don't know what structure they might need. That's why we described it as a generic map()

apps/aecore/lib/aecore/structures/data_tx.ex
gspasov: Because we are in the specified module and we can use this insted of an alias to itself. I think it's less confusing, if you think otherwise I can change it.

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -5,120 +5,129 @@ defmodule Aecore.Chain.ChainState do
...
 
-  @spec calculate_and_validate_chain_state!(list(), account_chainstate(), integer()) :: account_chainstate()
-  def calculate_and_validate_chain_state!(txs, chain_state, block_height) do
+  @typedoc "State of an account"
+  @type acc_state() :: %{balance: integer(),
```
d-velev: There already is an account structure, this shouldn't be needed.

apps/aecore/lib/aecore/structures/data_tx.ex
```diff
@@ -0,0 +1,88 @@
...
+  - type: The type of transaction that may be added to the blockchain
+  - payload: The strcuture of the specified transaction type
+  - from_acc: The public address of the account originating the transaction
+  - fee: The amount of tokens given to the miner
+  - nonce: A random integer generated on initialisation of a transaction (must be unique!)
+  """
+  defstruct [:type, :payload, :from_acc, :fee, :nonce]
+  use ExConstructor
+
+  @spec init(atom(), map(), binary(), integer(), integer()) :: {:ok, DataTx.t()}
```
thepiwo: is the return type correct?

apps/aecore/lib/aecore/structures/data_tx.ex
```diff
@@ -0,0 +1,88 @@
...
+    if fee >= 0 do
+      payload
+      |> type.init()
+      |> type.is_valid()
+    else
+      {:error, "Fee not enough"}
+    end
+  end
+
+  @spec process_chainstate(DataTx.t(), non_neg_integer(), map()) :: map()
```
thepiwo: can you be more specific of the `map()` type

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -5,120 +5,129 @@ defmodule Aecore.Chain.ChainState do
...
-  @spec update_chain_state_locked(account_chainstate(), Header.t()) :: map()
-  def update_chain_state_locked(chain_state, header) do
-    Enum.reduce(chain_state, %{}, fn({address, object}, acc) ->
-      case object do
-        account = %Account{} ->
-          Map.put(acc, address, Account.update_locked(account, header))
-        other ->
-          Map.put(acc, address, other)
-      end
+  @spec validate_tx(SignedTx.t(), chainstate(), integer()) :: {boolean(), map()}
```
thepiwo: can you be more specific of the `map()` type

apps/aecore/lib/aecore/structures/account.ex
gspasov: The tx_in and tx_out functionality is moved to it's corresponding transaction module SpendTx. In my opinion it's the most correct place for them. Each transaction module should handle it's work on the blockchain. The accounts module on the other hand is usefull for reasons of abstraction and even if at the moment it doesn't have much functionality that may change in the future.

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -5,120 +5,129 @@ defmodule Aecore.Chain.ChainState do
...
+          locked_sum + amount
+        end)
+      new_total_tokens = total_tokens + state.balance + locked_tokens
+      new_total_unlocked_tokens = total_unlocked_tokens + state.balance
+      new_total_locked_tokens = total_locked_tokens + locked_tokens
+      {new_total_tokens, new_total_unlocked_tokens, new_total_locked_tokens}
     end)
   end
 
+  @spec update_chain_state_locked(chainstate(), Header.t()) :: map()
```
thepiwo: can you be more specific of the `map()` type

apps/aecore/lib/aecore/pow/hashcash.ex
```diff
@@ -28,7 +28,7 @@ defmodule Aecore.Pow.Hashcash do
   @doc """
   Find a nonce
   """
-  @spec generate(map(), integer()) :: {:ok, Header.t} | {:error, term()}
+  @spec generate(map(), integer()) :: {:ok, Header.t()} | {:error, term()}
```
thepiwo: can you be more specific of the `map()` type

apps/aecore/lib/aecore/structures/data_tx.ex
thepiwo: can you be more specific of the `map()` type

apps/aecore/lib/aecore/structures/data_tx.ex
```diff
@@ -0,0 +1,88 @@
...
+
+    new_accounts_state =
+      tx.payload
+      |> tx.type.init()
+      |> tx.type.process_chainstate!(tx.from_acc, tx.fee, tx.nonce, block_height,
+                                account_state, subdomain_chainstate)
+      Map.put(chainstate, :accounts, new_accounts_state) ## TODO return the subdomain_chainstate as well
+  end
+
+  @spec serialize(DataTx.t()) :: map()
```
thepiwo: can you be more specific of the `map()` type

apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -1,19 +1,21 @@
...
   alias Aeutil.Serialization
   alias Aeutil.Bits
 
-  @typedoc "Structure of the SignedTx module"
-  @type t :: %SignedTx{
-          data: SpendTx.t(),
-          signature: binary()
-        }
+  @typedoc "Structure that holds the account info"
+  @type account_chainstate :: map()
```
thepiwo: can you be more specific of the `map()` type

thepiwo: Isn't this type defined somewhere else?

apps/aecore/lib/aecore/structures/spend_tx.ex
```diff
@@ -3,43 +3,129 @@ defmodule Aecore.Structures.SpendTx do
...
+  @behaviour Aecore.Structures.Transaction
 
-  @type t :: %SpendTx{
-    from_acc: binary(),
+  alias Aeutil.Serialization
+  alias Aeutil.Parser
+  alias Aecore.Structures.Account
+
+  @typedoc "Arbitrary structure data of a transaction"
+  @type payload :: %__MODULE__{} | map()
```
thepiwo: can you be more specific of the `map()` type and general about what can be "arbitrary" exactly

apps/aecore/lib/aecore/chain/worker.ex
gspasov: This is because before the chain_state was actually only holding the accounts, but now we want the chainstate to hold other {key, value} pairs as well. We typed it this way so that we save the whole chain state in the db, not only the accounts: key values, but all the other pairs that will come in the future as well. And then to get them from the db, you'll only need to create a extraction function in the persistence module

apps/aecore/lib/aecore/chain/worker.ex
gspasov: I'll look into this

apps/aecore/lib/aecore/structures/spend_tx.ex
thepiwo: yes but this should be described in the type, so everybody (including dyalizer) knows this

apps/aecore/lib/aecore/structures/transaction.ex
```diff
@@ -0,0 +1,76 @@
...
+
+  alias Aecore.Structures.DataTx
+  alias Aecore.Structures.Account
+  alias Aecore.Chain.ChainState
+
+  @typedoc "Arbitrary structure data of a transaction"
+  @type payload :: map()
```
thepiwo: describe map() more specific

apps/aecore/lib/aecore/structures/transaction.ex
```diff
@@ -0,0 +1,76 @@
...
+
+  @typedoc "Structure of a custom transaction"
+  @type tx_struct :: map()
```
thepiwo: describe map() more specific

apps/aecore/lib/aecore/structures/transaction.ex
```diff
@@ -0,0 +1,76 @@
...
+  @type reason :: String.t()
+
+  @typedoc "The public key of the account"
+  @type pub_key :: binary()
+
+  @typedoc "Structure that holds the account info"
+  @type account_state :: %{pub_key() => Account.t()}
+
+  @typedoc "Structure that holds specific transaction info in the chainstate"
+  @type subdomain_chainstate() :: map()
```
thepiwo: describe map() more specific

apps/aecore/lib/aecore/structures/transaction.ex
```diff
@@ -0,0 +1,76 @@
...
+
+          true ->
+           :ok
+      end
+  """
+  @callback preprocess_check(account_state(),
+                             fee :: non_neg_integer(),
+                             nonce :: non_neg_integer(),
+                             block_height :: non_neg_integer(),
+                             additional_variables :: map()) :: :ok | {:error, reason}
```
thepiwo: describe map() more specific

apps/aecore/lib/aecore/structures/spend_tx.ex
gspasov: Ok, I see your point now, will change this

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -5,120 +5,129 @@ defmodule Aecore.Chain.ChainState do
...
   """
-  @spec calculate_chain_state_hash(account_chainstate()) :: binary()
-  def calculate_chain_state_hash(chain_state) do
+  @spec calculate_chain_state_hash(chainstate()) :: binary()
+  def calculate_chain_state_hash(chainstate) do
     merkle_tree_data =
-      for {account, data} <- chain_state do
-        {account, Serialization.pack_binary(data)}
-      end
+    for {accounts, data} <- chainstate.accounts do
```
d-velev: should be {account, data} like before

apps/aecore/lib/aecore/structures/data_tx.ex
gspasov: Maybe not, will look into this and make the map() type more specific

apps/aecore/lib/aecore/structures/account.ex
thepiwo: It should be in account, as it affects the accounts state, otherwise the lock functionality makes no sense here as well. In the account you can call the function in the corresponding transaction module .

apps/aecore/lib/aecore/structures/data_tx.ex
thepiwo: I'd prefer to be specific about it, so by reading this line I know what happens, independently from the module

apps/aecore/lib/aecore/structures/data_tx.ex
```diff
@@ -0,0 +1,88 @@
...
+  alias Aecore.Keys.Worker, as: Keys
+  alias Aeutil.Serialization
+  alias Aeutil.Parser
+
+  @typedoc "Reason for the error"
+  @type reason :: String.t()
+
+  @typedoc "Structure of the main transaction wrapper"
+  @type t :: %__MODULE__{
+    type: atom(),
```
thepiwo: be more specific about `atom()`

apps/aecore/lib/aecore/structures/data_tx.ex
thepiwo: be more specific about `atom()` as well

thepiwo:
can you add some documentation in the Readme as well, how to make a transaction for example, it got really complicated with this (payload -> datatx -> get priv -> sign -> to pool) maybe add some utility to easy craft a spend_tx to some account from the current one (`create_spend_tx_to_pool(amount, to_account)`)

apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -30,12 +32,13 @@ defmodule Aecore.Structures.SignedTx do
...
-  @spec validate(SignedTx.t()) :: boolean()
-  def validate(%{data: data, signature: signature}) do
-    case data do
-      %SpendTx{} ->
-        SpendTx.validate(data) &&
-          Signing.verify(Serialization.pack_binary(data), signature, data.from_acc)
+  @spec is_valid?(SignedTx.t()) :: boolean()
+  def is_valid?(%SignedTx{data: data} = tx) do
+    if Signing.verify(Serialization.pack_binary(data), tx.signature, data.from_acc) do
+      case DataTx.is_valid(data) do
```
d-velev: Why don't the is_valid functions for each type return a boolean so that there is no need for a case, the reason message isn't used anyway. If we want to tell the user the input is invalid, we can Log.error(reason) and return false after that.

apps/aecore/lib/aecore/structures/spend_tx.ex
```diff
@@ -3,43 +3,129 @@ defmodule Aecore.Structures.SpendTx do
...
+          |> transaction_out(block_height, tx.value * -1, nonce, -1)
+        new_accounts = Map.put(accounts, from_acc, new_from_account_state)
+
+        to_acc = Map.get(accounts, tx.to_acc, Account.empty())
+        new_to_account_state =
+          transaction_in(to_acc, block_height, tx.value, tx.lock_time_block)
+        Map.put(new_accounts, tx.to_acc, new_to_account_state)
+
+      {:error, reason} = err ->
+        throw err
```
d-velev: Is this caught anywhere?

apps/aecore/lib/aecore/structures/account.ex
gspasov: So you mean that from the SpendTx module where i need to make the tx_in and tx_out I'll call them from the Account module like so: `Account.tx_in!()` and `Account.tx_out!()`

apps/aecore/lib/aecore/structures/data_tx.ex
gspasov: Yes, it'll be fixed as well

apps/aecore/lib/aecore/structures/data_tx.ex
```diff
@@ -3,17 +3,28 @@ defmodule  Aecore.Structures.DataTx do
...
   alias Aecore.Keys.Worker, as: Keys
+  alias Aecore.Chain.ChainState
+  alias Aecore.Structures.SpendTx
   alias Aeutil.Serialization
   alias Aeutil.Parser
 
+  require Logger
+
+  @typedoc "Name of the specified transaction module"
+  @type txType :: SpendTx | atom()
```
thepiwo: why do we need the `| atom()`, should be removed or specified as type

apps/aecore/lib/aecore/structures/data_tx.ex
```diff
@@ -3,17 +3,28 @@ defmodule  Aecore.Structures.DataTx do
...
+
+  @typedoc "Structure of a transaction that may be added to be blockchain"
+  @type payload :: SpendTx.t() | map()
```
thepiwo: why do we need the `| map()`, should be removed or specified as type

apps/aecore/lib/aecore/structures/data_tx.ex
thepiwo: the naming shouldn't be camelCase, we use underscore_lowercase in general

gspasov: Because we want to say that the txType will not only be SpendTx, but can be something else as well, if we create a new Transaction module

apps/aecore/lib/aecore/structures/data_tx.ex
```diff
@@ -32,49 +43,65 @@ defmodule  Aecore.Structures.DataTx do
...
                 payload: type.init(payload),
                 from_acc: from_acc,
                 fee: fee,
                 nonce: nonce}
   end
 
-  @spec is_valid(DataTx.t()) :: :ok | {:error, reason()}
-  def is_valid(%__MODULE__{type: type, payload: payload, fee: fee}) do
+  @spec is_valid?(DataTx.t()) :: boolean()
+  def is_valid?(%DataTx{type: type, payload: payload, fee: fee}) do
```
thepiwo: why do we need to check if it is bigger than 0 before passing to the type specific check? shouldn't the `type.is_valid?()` eliminate all 0 fees anyways?

apps/aecore/lib/aecore/structures/data_tx.ex
```diff
@@ -32,49 +43,65 @@ defmodule  Aecore.Structures.DataTx do
...
-      |> tx.type.init()
-      |> tx.type.process_chainstate!(tx.from_acc, tx.fee, tx.nonce, block_height,
-                                account_state, subdomain_chainstate)
-      Map.put(chainstate, :accounts, new_accounts_state) ## TODO return the subdomain_chainstate as well
+  @spec process_chainstate(DataTx.t(), non_neg_integer(),
+    ChainState.chainstate()) :: ChainState.chainstate()
+  def process_chainstate(%DataTx{} = tx, block_height, chainstate) do
+    try do
+      account_state = chainstate.accounts
+      subdomain_cs = Map.get(chainstate, tx.type, %{})
```
thepiwo: don't use `cs` as short name

apps/aecore/lib/aecore/structures/data_tx.ex
gspasov: Same goes for the payload, we don't want to specificaly say that the payload can only be SpendTx

apps/aecore/lib/aecore/structures/data_tx.ex
gspasov: Because then we can end the function earlier if the fee is 0 and don't go deeper into the code

apps/aecore/lib/aecore/structures/data_tx.ex
thepiwo: what is this something else? should be specified in the type

apps/aecore/lib/aecore/structures/data_tx.ex
thepiwo: what is this something else? should be specified in the type, if it for now is only `SpendTx`, just allow `SpendTx`

apps/aecore/lib/aecore/structures/data_tx.ex
thepiwo: what if there is a transaction that doesn't include a fee as rule? e.g. the coinbase transaction?

apps/aecore/lib/aecore/chain/worker.ex
thepiwo: what is included in chain_states?

apps/aecore/lib/aecore/structures/data_tx.ex
gspasov: Okay

apps/aecore/lib/aecore/structures/data_tx.ex
gspasov: The coinbase transaction sets fee: 0, so it will be catched with fee >= 0

thepiwo: oh yes, it's greater-equals, but I think its not necessary, just increases complexity

thepiwo: I'd agree to keep it if we turn the check around, make it fail if it is smaller 0 in semantics

gspasov: Not necessarily, in my opinion it gives separation of the validation checks, and it makes the checks in steps. That way we reduce unecessary computing (eg check if the specific transaction is valid before checking if the data transaction fields are valid beforehand). The fee is a type inside the DataTx, therefore it should be checked in DataTx module, the rest checks are inside the `type` in this case `:value`

thepiwo: yes I agree, but make the case you check correct in a semantic mannor, so that by reading the code  this allows ANY transaction to be fee > 0

apps/aecore/lib/aecore/structures/transaction.ex
```diff
@@ -0,0 +1,72 @@
...
+  alias Aecore.Structures.Account
+  alias Aecore.Chain.ChainState
+  alias Aecore.Wallet.Worker, as: Wallet
+
+  @typedoc "Arbitrary map holding all the specific elements required
+  by the specified transaction type"
+  @type payload :: map()
+
+  @typedoc "Structure of a custom transaction"
+  @type tx_struct :: SpendTx.t()
```
thepiwo: maybe rename to `tx_types`

thepiwo:
I think `subdomain_cs` is a completely incorrect name for what is describes

apps/aecore/lib/aecore/structures/account.ex
```diff
@@ -56,6 +61,32 @@ defmodule Aecore.Structures.Account do
                        locked: updated_locked}
   end
 
+  @doc """
+  Builds a SpendTx where the miners public key is used as a sender (from_acc)
+  """
+  @spec spend(Wallet.pubkey(), non_neg_integer(),
+              non_neg_integer()) :: {:ok, SignedTx.t()}
+  def spend(to_acc, amount, fee) do
```
thepiwo: Pretty cool!

gspasov:
@thepiwo @d-velev @meivantodorov @nedelcho-delchev-tues @cheezus1 @cytadela8 
Can you re-review? 
Only the README part is not here, will be added shortly.

apps/aecore/lib/aecore/chain/worker.ex
thepiwo: @gspasov ?

apps/aecore/lib/aecore/structures/data_tx.ex
```diff
@@ -0,0 +1,123 @@
...
+  alias Aecore.Keys.Worker, as: Keys
+  alias Aecore.Chain.ChainState
+  alias Aecore.Structures.SpendTx
+  alias Aeutil.Serialization
+  alias Aeutil.Parser
+
+  require Logger
+
+  @typedoc "Name of the specified transaction module"
+  @type tx_type :: SpendTx
```
thepiwo: I'd choose a plural name, to indicate more possible types

apps/aecore/lib/aecore/structures/spend_tx.ex
```diff
@@ -3,43 +3,132 @@ defmodule Aecore.Structures.SpendTx do
...
+    value: non_neg_integer(),
+    lock_time_block: non_neg_integer()
+  }
+
+  @typedoc "Reason for the error"
+  @type reason :: String.t()
+
+  @typedoc "Structure that holds specific transaction info in the chainstate.
+  In the case of SpendTx we don't have a subdomain chainstate."
+  @type tx_type_state() :: %{}
```
thepiwo: Is this always empty? otherwise make specific

apps/aecore/lib/aecore/structures/spend_tx.ex
```diff
@@ -3,43 +3,132 @@ defmodule Aecore.Structures.SpendTx do
...
-  @spec create(binary(), binary(), non_neg_integer(), non_neg_integer(), non_neg_integer(), non_neg_integer(), map()) :: {:ok, TxData.t()}
-  def create(from_acc, to_acc, value, nonce, fee, lock_time_block \\ 0, data \\ %{}) do
-    {:ok, %SpendTx{from_acc: from_acc,
-                  to_acc: to_acc,
-                  value: value,
-                  nonce: nonce,
-                  fee: fee,
-                  lock_time_block: lock_time_block,
-                  data: data}}
+  # Callbacks
```
thepiwo: why have this callbacks comment?

apps/aecore/lib/aecore/structures/data_tx.ex
thepiwo: in `Aecore.Structures.Transaction` the name is different

apps/aecore/lib/aecore/structures/spend_tx.ex
gspasov: In the case of SpendTx, yes it is always empty

apps/aecore/lib/aecore/structures/spend_tx.ex
gspasov: Maybe to indicate that they are not just some functions created in the SpendTx but come as callbacks from the Behavior, if you think it's unecessary I'll remove them

apps/aecore/lib/aecore/structures/data_tx.ex
gspasov: But then if you read some specification let say for example -> @spec init(tx_types(), ...) this would indicate to me that the first argument requires a list of types or something similar because of he plural form

apps/aecore/lib/aecore/chain/worker.ex
gspasov: At the moment the chain_state holds only the ket `:accounts` with its map of values of all the accounts, but later it will hold all the specific transaction states as well in there.

apps/aecore/lib/aecore/structures/spend_tx.ex
thepiwo: okay, just asking

apps/aecore/lib/aecore/structures/spend_tx.ex
thepiwo: okay, good!

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -192,26 +203,26 @@ defmodule Aecore.Chain.Worker do
...
     end)
 
-    state_update1 = %{state | blocks_map: hundred_blocks_map,
+    state_update = %{state | blocks_map: hundred_blocks_map,
                               chain_states: updated_chain_states,
                               txs_index: new_txs_index}
+
     if top_height < new_block.header.height do
-      Persistence.batch_write(%{:chain_state => new_chain_state,
+      Persistence.batch_write(%{:chain_state => %{:chain_state => new_chain_state},
```
d-velev: Should be just %{chain_state: %{chain_state: new_chain_state}, and so on

apps/aecore/lib/aecore/persistence/worker.ex
```diff
@@ -187,6 +190,38 @@ defmodule Aecore.Persistence.Worker do
     {:reply, all_blocks, state}
   end
 
+  def handle_call(:delete_all_blocks, _from,
+    %{blocks_family: blocks_family} = state) do
+    for {key, _} <- Rox.stream(blocks_family), do: Rox.delete(blocks_family, key)
```
d-velev: Use Enum.each instead

apps/aecore/lib/aecore/persistence/worker.ex
```diff
@@ -187,6 +190,38 @@ defmodule Aecore.Persistence.Worker do
...
+    {:reply, :ok, state}
+  end
+
+  def handle_call(:delete_chainstate, _from,
+    %{chain_state_family: chain_state_family} = state) do
+    for {key, _} <- Rox.stream(chain_state_family), do: Rox.delete(chain_state_family, key)
```
d-velev: Same as the comment above

apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -30,12 +29,10 @@ defmodule Aecore.Structures.SignedTx do
...
 
-  @spec validate(SignedTx.t()) :: boolean()
-  def validate(%{data: data, signature: signature}) do
-    case data do
-      %SpendTx{} ->
-        SpendTx.validate(data) &&
-          Signing.verify(Serialization.pack_binary(data), signature, data.from_acc)
+  @spec is_valid?(SignedTx.t()) :: boolean()
+  def is_valid?(%SignedTx{data: data} = tx) do
+    if Signing.verify(Serialization.pack_binary(data), tx.signature, data.from_acc) do
```
d-velev: there should be an else returning false in case the signature isn't valid

apps/aecore/lib/aecore/persistence/worker.ex
gspasov: Why should I use Enum.each instead, is there a benefit to it? Give reason behind your suggestion

apps/aecore/lib/aecore/chain/worker.ex
gspasov: This is done for readability purpose

apps/aeutil/lib/serialization.ex
```diff
@@ -97,26 +75,99 @@ defmodule Aeutil.Serialization do
     if is_tuple(head) do
       merkle_proof(Tuple.to_list(head), acc)
     else
-      acc = [bech32_binary(head, :account, :serialize) | acc]
+      acc = [serialize_value(head, :account) | acc]
       merkle_proof(tail, acc)
     end
   end
 
   @spec pack_binary(term()) :: map()
```
d-velev: This looks a bit confusing, I think it's better to keep it the way it was before (have a case depending on the data which can be extended), it seems hard to handle each possible case in advance and have it done with code that looks good. Also, the map and list packs look wrong, the list one returns a list in the end when it should be a msgpack binary and the map one packs each value and then packs the whole thing again which gives an incorrect result.

apps/aecore/lib/aecore/persistence/worker.ex
d-velev: Because for returns a list of the results for each iteration and each returns an :ok

apps/aeutil/lib/serialization.ex
gspasov: I'm looking into this

apps/aecore/lib/aecore/persistence/worker.ex
gspasov: I don't think this would matter in this case, having in mind that we don't use the result of the for/Enum.each result. We just delete each element

thepiwo:
@gspasov can you merge with master again?

apps/aecore/lib/aecore/persistence/worker.ex
d-velev: Yes, we don't use the result, that's why each should be used, using for will store each result which is an unneeded operation

thepiwo:
@gspasov merging master and fixing the compiler warnings should eliminate the ci failure

apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -1,19 +1,19 @@
 defmodule Aecore.Structures.SignedTx do
   @moduledoc """
-  Aecore structure of a signed transaction.
```
thepiwo: why is the moduledoc text removed?

apps/aecore/test/aecore_chain_test.exs
```diff
@@ -15,12 +16,16 @@ defmodule AecoreChainTest do
 
   setup do
     Chain.start_link([])
+    on_exit fn ->
+      Persistence.delete_all_blocks()
```
thepiwo: shouldn't the state be deleted as well?

apps/aecore/test/aecore_get_txs_for_address.exs
```diff
@@ -10,8 +11,17 @@ defmodule GetTxsForAddressTest do
   alias Aecore.Txs.Pool.Worker, as: Pool
   alias Aeutil.Serialization
 
+
+  setup do
+    on_exit fn ->
+      Persistence.delete_all_blocks()
```
thepiwo: shouldn't the state be deleted as well?

apps/aecore/test/aecore_miner_test.exs
```diff
@@ -1,13 +1,21 @@
...
   use ExUnit.Case
 
+  alias Aecore.Persistence.Worker, as: Persistence
   alias Aecore.Structures.SignedTx
   alias Aecore.Chain.Worker, as: Chain
   alias Aecore.Miner.Worker, as: Miner
 
+  setup do
+    on_exit fn ->
+      Persistence.delete_all_blocks()
```
thepiwo: shouldn't the state be deleted as well?

apps/aecore/test/aecore_persistence_test.exs
```diff
@@ -11,12 +11,16 @@ defmodule PersistenceTest do
...
     Miner.mine_sync_block_to_chain
-    path = Application.get_env(:aecore, :persistence)[:path]
+    Miner.mine_sync_block_to_chain
+
     on_exit fn ->
-      if File.exists?(path) do
-        File.rm_rf(path)
-      end
+      Persistence.delete_all_blocks()
+      Chain.clear_state()
```
thepiwo: shouldn't the state in the persistence be deleted as well?

apps/aecore/test/aecore_pow_cuckoo_test.exs
```diff
@@ -2,25 +2,35 @@ defmodule AecoreCuckooTest do
...
   alias Aecore.Structures.SpendTx
 
   @moduledoc """
   Unit tests for the cuckoo module
   """
-   @tag timeout: 10_000
+
+  setup do
+    on_exit fn ->
+      Persistence.delete_all_blocks()
```
thepiwo: shouldn't the state be deleted as well?

apps/aecore/lib/aecore/structures/signed_tx.ex
gspasov: Strange, it shouldn't be

apps/aecore/test/aecore_chain_test.exs
thepiwo: (same in other tests, I deleted my duplicated comments)

Issue #278, opened 2018-03-16
remove unused functionality and logging from phoenix endpoint


thepiwo:

thepiwo:
closed with #279

PR #279, opened 2018-03-16
[GH-#278] removed unused plug from endpoint, log to debug

thepiwo:


thepiwo:
@DanielaIvanova @Artur64 may you review this?

PR #290, opened 2018-03-16
[GH-#256] Adjusted block, header encodings

### LINKED ISSUE #255: adjust header / block structure fields
- [x] rename `chain_state_hash` to `root_hash`
- [x] rename `difficulty_target` to `target`
- [x] rename `timestamp` to `time`

### LINKED ISSUE #256: adjust block / header field encoding
- [ ] encode `prev_hash` using base58c
- [ ] encode `root_hash` using base58c
- [ ] encode `txs_hash` using base58c
- [ ] encode `txs` using base58c for each tx over the binary(msgpack) representation

Artur64:
Resolves #255, resolves #256


thepiwo:
to merge, first format your code mix format using elixir 1.6.2 and merge master afterwards.

thepiwo:
@Artur64 can you use the default merge commit message in the future?

apps/aecore/lib/aecore/wallet/worker.ex
```diff
@@ -105,11 +106,11 @@ defmodule Aecore.Wallet.Worker do
   """
   @spec encode(binary(), wallet_type()) :: String.t()
   def encode(pub_key, :ae) do
-    Encoding.encode(pub_key, :ae)
+    Bits.encode58("ak$", pub_key)
   end
 
   def encode(pub_key, :btc) do
-    Encoding.encode(pub_key, :btc)
+    Bits.encode58("btc", pub_key)
```
thepiwo: can you remove this function overall

apps/aeutil/lib/bits.ex
```diff
@@ -1,14 +1,20 @@
 # Taken from http://minhajuddin.com/2016/11/01/how-to-extract-bits-from-a-binary-in-elixir/
 # License: CC BY-SA 3.0
 defmodule Aeutil.Bits do
-  def bech32_encode(prefix, bin) do
-    SegwitAddr.encode(prefix, 0, :binary.bin_to_list(bin))
+  @prefix_list ["ak$", "bh$", "bs$", "tx$", "bx$", "ok$", "cs$", "tr$"]
+
+  def encode58(prefix, data) when is_binary(data) do
```
thepiwo: can you make the prefix be passed as atom here `:account_hash` , `:block_hash` ...
based on this choose the prefix

apps/aeutil/lib/bits.ex
```diff
@@ -1,14 +1,20 @@
...
+    prefix <> Kernel.to_string(:base58.binary_to_base58(data))
   end
 
-  def bech32_decode(bech32) do
-    case SegwitAddr.decode(bech32) do
-      {:ok, {_, _, bin_list}} -> :binary.list_to_bin(bin_list)
-      {:error, _} = error -> error
+  def decode58(data) when is_binary(data) do
```
thepiwo: can you pass the expected prefix as atom here as well, fail if the one in data doesn't match the expected prefix

apps/aeutil/lib/bits.ex
thepiwo: rename to `base58c_encode`

apps/aeutil/lib/bits.ex
thepiwo: rename to `base58c_decode`

mix.exs
```diff
@@ -52,7 +52,8 @@ defmodule EpochElixir.Mixfile do
       {:distillery, "~> 1.5.2", runtime: false},
       {:aewallet, github: "aeternity/elixir-wallet"},
       {:msgpax, "~> 2.1.1"},
-      {:bip0173, "~> 0.1.2"}
+      {:bip0173, "~> 0.1.2"},
+      {:erl_base58, "~> 0.0.1"}
```
thepiwo: does this library actually do base58check?

mix.lock
```diff
@@ -1,43 +1,160 @@
 %{
```
thepiwo: this file should not be formatted, this is why it is not included in `.formatter.exs`

mix.exs
Artur64: It encodes/decodes binaries/integers to base58 and vice versa

thepiwo: as requested in #256, we need `base58c`

apps/aehttpserver/lib/aehttpserver/web/controllers/balance_controller.ex
```diff
@@ -2,14 +2,14 @@ defmodule Aehttpserver.Web.BalanceController do
   use Aehttpserver.Web, :controller
 
   alias Aecore.Chain.Worker, as: Chain
-  alias Aewallet.Encoding
+  alias Aeutil.Bits
 
   def show(conn, params) do
-    case Encoding.decode(params["account"]) do
+    case Bits.decode58c(params["account"]) do
```
thepiwo: Use the `Account.base58_decode` function to decode account hashes

apps/aehttpserver/lib/aehttpserver/web/controllers/block_controller.ex
```diff
@@ -77,8 +75,7 @@ defmodule Aehttpserver.Web.BlockController do
           BlockValidation.block_header_hash(Block.genesis_block().header)
 
         hash ->
-          hash_bin = Bits.bech32_decode(hash)
-          hash_bin
+          Bits.decode58c(hash)
```
thepiwo: Use the `Block.base58_decode` function to decode block hashes

apps/aeutil/lib/serialization.ex
```diff
@@ -129,25 +130,25 @@ defmodule Aeutil.Serialization do
...
       :to_acc ->
-        Aewallet.Encoding.encode(value, :ae)
+        Account.base58_encode(Kernel.to_string(value))
 
       :signature ->
         base64_binary(value, :serialize)
 
       :proof ->
-        Aewallet.Encoding.encode(value, :ae)
+        Account.base58_encode(value)
```
thepiwo: why does the proof use the this?

apps/aeutil/lib/serialization.ex
```diff
@@ -180,9 +181,9 @@ defmodule Aeutil.Serialization do
   end
 
   def deserialize_value(value) when is_binary(value) do
-    case Bits.bech32_decode(value) do
+    case Bits.decode58c(value) do
```
thepiwo: remove this function and use the separate decoder at each position

apps/aeutil/lib/bits.ex
```diff
@@ -1,22 +1,69 @@
 # Taken from http://minhajuddin.com/2016/11/01/how-to-extract-bits-from-a-binary-in-elixir/
 # License: CC BY-SA 3.0
 defmodule Aeutil.Bits do
-  def bech32_encode(prefix, bin) do
-    SegwitAddr.encode(prefix, 0, :binary.bin_to_list(bin))
+  @prefix_list ["ak$", "bh$", "bs$", "tx$", "bx$", "ok$", "cs$", "tr$"]
+
+  def encode58c(prefix, payload) when is_binary(payload) do
+    case prefix do
+      :account_pubkey ->
```
thepiwo: separate the prefix to atom conversion, to remove duplicate code

apps/aeutil/lib/bits.ex
thepiwo: that about this?

apps/aeutil/lib/bits.ex
thepiwo: what about renaming this?

apps/aeutil/lib/serialization.ex
gspasov: Because this is how the proof was being encoded [before](https://github.com/aeternity/elixir-node/pull/263/files#diff-92c5a5701af6ae4117cba6975aba699aL88)

apps/aehttpserver/lib/aehttpserver/web/controllers/block_controller.ex
Artur64: There is no such function

apps/aehttpserver/lib/aehttpserver/web/controllers/balance_controller.ex
Artur64: There is no such function

apps/aeutil/lib/serialization.ex
thepiwo: then I didn't catch it earlier, what is the reasoning for this, do we need a specific encoding for this at all? [integer()] shouldn't, right?

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -81,10 +80,10 @@ defmodule Aecore.Chain.Worker do
...
   end
 
-  @spec get_block_by_bech32_hash(String.t()) :: Block.t() | {:error, binary()}
-  def get_block_by_bech32_hash(hash) do
-    decoded_hash = Bits.bech32_decode(hash)
-    get_block(decoded_hash)
+  @spec get_block_by_base58_hash(String.t()) :: Block.t()
+  def get_block_by_base58_hash(hash) do
+    decoded_hash = Header.base58c_decode(hash)
+    GenServer.call(__MODULE__, {:get_block_info_from_memory_unsafe, decoded_hash})
```
thepiwo: why doesn't this use the save call any more?

apps/aeutil/lib/serialization.ex
```diff
@@ -128,26 +126,26 @@ defmodule Aeutil.Serialization do
...
       :to_acc ->
-        Aewallet.Encoding.encode(value, :ae)
+        Account.base58c_encode(value)
 
       :signature ->
         base64_binary(value, :serialize)
 
       :proof ->
-        Aewallet.Encoding.encode(value, :ae)
+        Account.base58c_encode(value)
```
thepiwo: I still don't get why this should be correct

apps/aeutil/lib/bits.ex
```diff
@@ -1,15 +1,45 @@
 # Taken from http://minhajuddin.com/2016/11/01/how-to-extract-bits-from-a-binary-in-elixir/
 # License: CC BY-SA 3.0
 defmodule Aeutil.Bits do
-  def bech32_encode(prefix, bin) do
-    SegwitAddr.encode(prefix, 0, :binary.bin_to_list(bin))
+  def encode58c(type, payload) when is_binary(payload) do
+    prefix =
+      case type do
```
thepiwo: my imagination was to use the atoms in both ways, like this it makes no sense, you can pass the prefix from the calling function

gspasov: I thought we use the atom as just to clarify the prefix that we are using. After all we have to get the string value of the prefix. Yes, we could just pass it in the calling function, but maybe this way is more readable. That was our understanding of the your idea. Can u elaborate a bit futher on this, maybe we are missing the point for some reason.

apps/aeutil/lib/serialization.ex
gspasov: This is how the `proof` was encoded before. If there is a another prefix for the `proof` encoding we will change it.

apps/aeutil/lib/bits.ex
thepiwo: for decoding you pattern match the prefix - which is fine, my idea was to pass the expected prefix as atom

gspasov: Oh, and then convert it to string ?

thepiwo: then pattern match both ways to/from string. But your approach is fine, just pass the prefix as string instead of atom, to match the decoding.

apps/aecore/lib/aecore/chain/worker.ex
gspasov: I have no idea why did is changed, we'll fix this

mix.exs
thepiwo: does this library actually do base58check?

gspasov: No, we do it manually in the code

apps/aeutil/lib/serialization.ex
thepiwo: And I commented on that before, lets use base64 without prefix for now

mix.exs
thepiwo: we have the `quanterall/base58check`, why not use it?

gspasov: Because the prefix there is fixed to 4 bytes, we use prefix with 3 in our case

thepiwo: `quanterall/base58check` can be used without prefix as well

gspasov: Yes, we can pass it like encoded = Base58Check.encode58check("", value_to_encode) and it will work, but when we decode -> Base58Check.decode58check(encoded) we get a return value:

thepiwo: okay, lets keep it like this...

thepiwo: yes, lets keep it, as is

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -49,8 +49,8 @@ defmodule Aecore.Chain.ChainState do
   Builds a merkle tree from the passed chain state and
   returns the root hash of the tree.
   """
-  @spec calculate_chain_state_hash(chainstate()) :: binary()
-  def calculate_chain_state_hash(chainstate) do
+  @spec calculate_txs_hash(chainstate()) :: binary()
```
d-velev: This name doesn't make sense, call it calculate_root_hash

apps/aeutil/lib/bits.ex
```diff
@@ -1,15 +1,35 @@
...
-      {:ok, {_, _, bin_list}} -> :binary.list_to_bin(bin_list)
-      {:error, _} = error -> error
-    end
+  defp encode58(payload) do
+    checksum = generate_checksum(payload)
+
+    payload
+    |> Kernel.<>(checksum)
+    |> :base58.binary_to_base58()
+    |> Kernel.to_string()
```
d-velev: Don't need to use Kernel here

apps/aeutil/lib/bits.ex
```diff
@@ -1,15 +1,35 @@
...
+    checksum
+  end
+
+  def decode58(payload) do
+    decoded_payload =
+      payload
+      |> String.to_charlist()
+      |> :base58.base58_to_binary()
+
+    bsize = Kernel.byte_size(decoded_payload) - 4
```
d-velev: Don't need to use Kernel here

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -49,8 +49,8 @@ defmodule Aecore.Chain.ChainState do
   Builds a merkle tree from the passed chain state and
   returns the root hash of the tree.
   """
-  @spec calculate_txs_hash(chainstate()) :: binary()
-  def calculate_txs_hash(chainstate) do
+  @spec calculate_root_hash(chainstate()) :: binary()
```
thepiwo: this function does `calculate_root_hash`

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -123,13 +123,13 @@ defmodule Aecore.Chain.BlockValidation do
     end)
   end
 
-  @spec calculate_txs_hash(list(SignedTx.t())) :: binary()
-  def calculate_txs_hash(txs) when txs == [] do
+  @spec calculate_root_hash(list(SignedTx.t())) :: binary()
```
thepiwo: this function does `calculate_txs_hash`

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -123,13 +123,13 @@ defmodule Aecore.Chain.BlockValidation do
...
+  def calculate_root_hash(txs) when txs == [] do
     <<0::256>>
   end
 
-  @spec calculate_txs_hash(list(SignedTx.t())) :: binary()
-  def calculate_txs_hash(txs) do
+  @spec calculate_root_hash(list(SignedTx.t())) :: binary()
+  def calculate_root_hash(txs) do
```
thepiwo: this function does `calculate_txs_hash`

thepiwo: the txs_hash is the merkle roota hash of the transactions tree included within one block

apps/aecore/lib/aecore/chain/chain_state.ex
thepiwo: the root_hash (or (chain) state root hash) is the merkle root hash of the state (tree) at the height of this block

Issue #280, opened 2018-03-16
rename transaction fields


thepiwo:
- [x] from_acc => sender
- [x] to_acc => receiver
- [x] value => amount

PR #292, opened 2018-03-19
[GH-#281] Added version field in the SpendTx module structure

### LINKED ISSUE #281: include version number in specific transactions
- [x] use hardcoded value, 1 for now
- [x] add version to spend_tx

gspasov:
resolves #281 
Will be merged to #280 which will merge both the issues (#280 and #281) with the master branch

### DIFF
--- a/apps/aecore/config/config.exs
+++ b/apps/aecore/config/config.exs
@@ -37,12 +37,6 @@ persistence_path =
     env -> env
   end
 
-spend_tx_version =
-  case System.get_env("SPEND_TX_VERSION") do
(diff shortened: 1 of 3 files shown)


apps/aecore/config/config.exs
```diff
@@ -37,6 +37,12 @@ persistence_path =
     env -> env
   end
 
+spend_tx_version =
+  case System.get_env("SPEND_TX_VERSION") do
```
thepiwo: We don't need it via environment variable

PR #293, opened 2018-03-20
[GH-#265] Making credo more strict and fixing the issues

### LINKED ISSUE #265: make credo config more strict and fix issues

gspasov:
resolves #265


cytadela8:
The priority of checks is mostly: "high" and "higher". Shouldn't we use the whole spectrum of priorities? Or is it due to how it interacts with "Trevis"?

gspasov:
This is just a `tag` to set how importang this check is for you. If you have a `high` or `higher` check Travis won't accept the pull request. It's made this way so that a sloppy code doesn't pass through Travis.

cytadela8:
Ok, so it's for Travis.

thepiwo:
yes it is to have a mandatory strictness to the code - for now.

thepiwo:
I do prefer methods without arguments, to use `()`, what do you thin?

gspasov:
Well, I guess that's how Elixir style wants them. Thats why when we do `mix format` if we have a anonymous function it removes the brackets, e.g. 
```elixir 
Enum.reduce(%{}, %{}, fn a, b -> c end)
```

cytadela8:
I also like having parenthesis always when there is a function declaration or call involved, although general Elixir style rules may state otherwise.

gspasov:
As far as I've seen they prefer no parenthesis on declaration, but having them when the func is called

apps/aecore/lib/aecore/chain/difficulty.ex
```diff
@@ -19,7 +23,9 @@ defmodule Aecore.Chain.Difficulty do
       distance = calculate_distance(list)
 
       next_difficulty =
-        (latest_block.header.difficulty_target * (@target_distance / distance))
```
d-velev: Does credo fail if it is done this way? @target_distance / distance can be separated in a variable to avoid using Kernel

gspasov: Credo requires when pipe operator is used to start with a single variable. Is using Kernel a problem here?

d-velev: Never mind, I thought it was failing because of the arithmetic expression.

thepiwo: this look really weird

gspasov: Well, I could seperate it in another variable as @d-velev suggested

thepiwo: yes okay

untra:
I have to say, you guys have set up some quality guard rails and have whipped this repo into shape.
This is the first time I've seen an Elixir repo crank up the credo strictness levels this high. Everyone on the Aeternity team should be proud; most Elixir projects never get themselves setup so well. This repo has been whipped into shape.

Good work 👍

PR #302, opened 2018-03-22
[GH-#299] Fixes a bug where an invalid transaction would be mined

### LINKED ISSUE #299: transactions resulting in invalid chainstate get mined
will provide a test case in code

gspasov:
resolves #299

### DIFF
--- a/apps/aecore/test/aecore_tx_test.exs
+++ b/apps/aecore/test/aecore_tx_test.exs
@@ -82,8 +82,8 @@ defmodule AecoreTxTest do
 
   test "invalid spend transaction", tx do
     from_acc = Wallet.get_public_key()
-    value = 500
-    fee = 300
+    value = 200
+    fee = 50
 
     :ok = Miner.mine_sync_block_to_chain()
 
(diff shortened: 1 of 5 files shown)


apps/aecore/test/aecore_tx_test.exs
```diff
@@ -73,4 +79,29 @@ defmodule AecoreTxTest do
...
+    priv_key = Wallet.get_private_key()
+    {:ok, signed_tx} = SignedTx.sign_tx(tx_data, priv_key)
+
+    :ok = Pool.add_transaction(signed_tx)
+
+    :ok = Miner.mine_sync_block_to_chain()
+
+    # We should have only made two coinbase transactions
+    assert Enum.count(Chain.chain_state().accounts) == 1
+    assert Chain.chain_state().accounts[Wallet.get_public_key()].balance == 200
```
thepiwo: can you reduce value to 200 and fee to 50, mine one more block, it should be included by then

gspasov: @thepiwo I should mine 2 more blocks, because the miner will need 300 tokens so that the transaction is valid, so the transaction will be mined on the 4th mined block

thepiwo: yes realized that, good!

PR #315, opened 2018-03-27
[GH-313] Fixed HTTP server endpoint bug

### LINKED ISSUE #313: fix HTTP server endpoints to work with the new block serialization
Example - in BlockController:
       `Enum.map(blocks, fn block ->
        hash = BlockValidation.block_header_hash(block.header)
        %{
          "hash" => Header.base58c_encode(hash),
          "header" => Serialization.block(block, :serialize).header,
          "tx_count" => Enum.count(block.txs)
        }
      end)`

`Serialization.block(block, :serialize).header` gives a key :header not found since the header field isn't present in the serialized block structure anymore. I'm not sure if there are other cases where this happens

Artur64:
Resolves #313


thepiwo:
@Artur64 no other places where this is called?

PR #316, opened 2018-03-28
[GH-#296] Fixing the transasction pattern matching problem in chain_state

### LINKED ISSUE #296: fix chain transaction type matching
https://github.com/aeternity/elixir-node/blob/master/apps/aecore/lib/aecore/chain/worker.ex#L383

The pattern-matching here can't work correct

gspasov:
resolves #296 

Also renamed the chain_sync key to be chain_sync_status as it was before.

### DIFF
--- a/apps/aecore/lib/aecore/chain/worker.ex
+++ b/apps/aecore/lib/aecore/chain/worker.ex
@@ -7,8 +7,6 @@ defmodule Aecore.Chain.Worker do
   use Bitwise
 
   alias Aecore.Structures.Block
-  alias Aecore.Structures.SpendTx
-  alias Aecore.Structures.DataTx
   alias Aecore.Structures.Header
   alias Aecore.Chain.ChainState
   alias Aecore.Txs.Pool.Worker, as: Pool
@@ -22,6 +20,7 @@ defmodule Aecore.Chain.Worker do
   require Logger
 
   @type txs_index :: %{binary() => [{binary(), binary()}]}
+
   # upper limit for number of blocks is 2^max_refs
   @max_refs 30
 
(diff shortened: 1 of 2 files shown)


apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -379,39 +378,23 @@ defmodule Aecore.Chain.Worker do
...
-
-          %DataTx{} ->
-            tx.data.sender
-        end
-      end
-
-    accounts_unique = accounts |> List.flatten() |> Enum.uniq() |> List.delete(nil)
+    accounts_unique =
+      block.txs
+      |> Enum.map(&[&1.data.sender, &1.data.payload.receiver])
```
thepiwo: by removing the pattern match, we make it more difficult to use this with transactions that have no receiver.

thepiwo: by removing the pattern match, we make it more difficult to use this with transactions that have no receiver.

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -20,7 +20,7 @@ defmodule Aecore.Peers.Sync do
   def start_link(_args) do
     GenServer.start_link(
       __MODULE__,
-      %{peer_blocks: %{}, peer_block_tasks: %{}, chain_sync_running: false},
+      %{peer_blocks: %{}, peer_block_tasks: %{}, chain_sync_status: false},
```
thepiwo: why is this renamed? status true|false is semantic incorrect, running true|false was correct

gspasov: I renamed it to running in a previous issue, and now I thought that it was maybe a bit confusing. I'll change it.

thepiwo: if you want to use status as name, use atoms as identifier

gspasov: No, I'll rename it as you suggested. The semantic name does make more sense.

thepiwo: I'm just saying status could be semantical if it is not boolean, but atom

gspasov: Yes, but then in the places where the status is checked with `if statement` it should be changed with `case`, and there is no reason for that, if I can just return the naming.

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -380,10 +380,20 @@ defmodule Aecore.Chain.Worker do
 
     accounts_unique =
       block.txs
-      |> Enum.map(&[&1.data.sender, &1.data.payload.receiver])
+      |> Enum.map(fn tx ->
+        case tx.data.type do
```
thepiwo: aren't we back at the patternmatch that can't work?

gspasov: You are correct, idk where my thinking was. Fixing it now

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -379,39 +379,31 @@ defmodule Aecore.Chain.Worker do
...
-          tx_bin = Serialization.pack_binary(tx)
-          :crypto.hash(:sha256, tx_bin)
-        end)
-
+      # txs associated with the given account
       tx_tuples =
-        Enum.map(tx_hashes, fn hash ->
+        block.txs
+        |> Enum.filter(fn tx ->
+          tx.data.sender == account || tx.data.payload.receiver == account
```
thepiwo: don't we have the same problem here? https://github.com/aeternity/elixir-node/pull/316#discussion_r177672949

gspasov: I don't think so, because the filter will only take the transactions where the sender or the receiver is matching the account. Here we don't care if we have one or the other. Maybe it won't work if we neither have sender nor receiver, but I don't think we are going to have such scenario.

PR #317, opened 2018-03-29
[GH-#252] Extract state structure handling

### LINKED ISSUE #252: extract state structure handling
To goal is to have each kind of state in a separate structure inside the chainstate. 

```
{
 accounts: account_state,
 oracle: oracle_state,
 voting: voting_state
}
```

This task should make the following refactoring:
 - make chain state structure with type, represent top structure in structures folder 
 - move applying functions from chain_state to the structure (I think that are all functions for now, but keep the chain_state as wrapper, call should be e.g. `chain worker` -> `chain_state` -> `state structure`, we will need this as later tickets refactor the state structure to tree representation but the api to current functionality should stay the same)

rdimitrovbg:
resolves #252


thepiwo:
what is the state on resolving conflicts?

DanielaIvanova:
@thepiwo We are almost ready. Have to review everything and test again.

apps/aecore/lib/aecore/structures/spend_tx.ex
```diff
@@ -78,29 +79,36 @@ defmodule Aecore.Structures.SpendTx do
...
 
-    preprocess_check!(tx, sender, sender_account_state, fee, block_height, %{})
+  defp process_chainstate!(%SpendTx{} = tx, sender, fee, nonce, accounts, %{}) do
+    sender_account_state = Account.get_account_state(accounts, sender)
 
-    new_sender_account_state =
-      sender_account_state
-      |> deduct_fee(fee)
-      |> Account.transaction_out(tx.amount * -1, nonce)
+    case preprocess_check!(tx, sender_account_state, fee) do
```
gspasov: Shouldn't you call the full function of preprocess_check!/7 here ?

DanielaIvanova: We think that is more readable and clean. You don't have to pass unused arguments.

thepiwo: I would adjust this, to keep it the same as the transaction interface

apps/aecore/lib/aecore/structures/account.ex
```diff
@@ -30,14 +33,22 @@ defmodule Aecore.Structures.Account do
 
   def empty, do: %Account{balance: 0, nonce: 0}
 
+  @spec new(account_payload()) :: Account.t()
```
d-velev: Doesn't the ExConstructor do exactly this?

apps/aecore/lib/aecore/structures/account.ex
```diff
@@ -30,14 +33,22 @@ defmodule Aecore.Structures.Account do
...
+
   @doc """
   Builds a SpendTx where the miners public key is used as a sender (sender)
   """
   @spec spend(Wallet.pubkey(), non_neg_integer(), non_neg_integer()) :: {:ok, SignedTx.t()}
   def spend(receiver, amount, fee) do
     sender = Wallet.get_public_key()
     sender_priv_key = Wallet.get_private_key()
-    nonce = Map.get(Chain.chain_state().accounts, sender, %{nonce: 0}).nonce + 1
+    nonce = Account.nonce(Chain.chain_state().accounts, sender) + 1
```
d-velev: just nonce instead of Account.nonce or is it done for readability?

rdimitrovbg: Yes, is it done for readability.

apps/aecore/lib/aecore/structures/account.ex
```diff
@@ -70,13 +81,42 @@ defmodule Aecore.Structures.Account do
...
   def transaction_out(account_state, amount, nonce) do
     account_state
     |> Map.put(:nonce, nonce)
     |> transaction_in(amount)
   end
 
+  @spec get_account_state(AccountStateTree.tree(), Wallet.pubkey()) :: Account.t()
+  def get_account_state(tree, key) do
+    case AccountStateTree.get(tree, key) do
+      :none ->
```
d-velev: Not important but just returning empty() is be faster

apps/aecore/lib/aecore/oracle/oracle.ex
```diff
@@ -218,18 +224,17 @@ defmodule Aecore.Oracle.Oracle do
                                                                         acc ->
       if calculate_absolute_ttl(tx.ttl, height_included) <= block_height do
         acc
+        |> Chainstate.destruct()
         |> pop_in([:oracles, :registered_oracles, address])
```
d-velev: Acces.key(oracles) can be used instead of destructing the chainstate, this will remove the to_struct line as well

apps/aecore/lib/aecore/chain/chain_state_wrapper.ex
```diff
@@ -1,4 +1,4 @@
-defmodule Aecore.Chain.ChainState do
+defmodule Aecore.Chain.ChainStateWrapper do
```
thepiwo: why is a wrapper needed at all?

thepiwo:
@gspasov can you re-review?

DanielaIvanova:
@thepiwo @d-velev 
We have a following question:
When we are looking for an account_state in the account_state_tree for a given key and this key does not exist:
There are 2 options:
- empty() - This is the implementation at this moment. We are not making a record in the account_state_tree.
- we can make a blank record for this key in the account_state_tree.
Which option suit better in this case?

thepiwo:
@DanielaIvanova I guess the first option is the viable one, the other one would lead to different state if there was a try to get some key.

thepiwo:
@DanielaIvanova @rdimitrovbg is this ready to merge?

rdimitrovbg:
@thepiwo 
Yes, it is ready.

PR #319, opened 2018-03-29
[GH-#305] Fixed Transaction pattern match

### LINKED ISSUE #305: refactor preprocess_check to throw its errors

gspasov:
resolves #305 
Additionally I added two minor test fixes and fixed 2 credo warnings about the `try do` statement

### DIFF
--- a/apps/aecore/lib/aecore/chain/chain_state.ex
+++ b/apps/aecore/lib/aecore/chain/chain_state.ex
@@ -91,11 +91,11 @@ defmodule Aecore.Chain.ChainState do
 
   @spec validate_tx(SignedTx.t(), chainstate(), integer()) :: {boolean(), chainstate()}
   defp validate_tx(tx, chainstate, block_height) do
-    try do
-      {true, apply_transaction_on_state!(tx, chainstate, block_height)}
-    catch
-      {:error, _} -> {false, chainstate}
-    end
+    {true, apply_transaction_on_state!(tx, chainstate, block_height)}
+  catch
+    {:error, reason} ->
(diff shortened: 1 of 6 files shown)


gspasov:
It's just redundant to have it. If some of the code in the function throws an error the `catch` block below will handle it.

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -91,11 +91,11 @@ defmodule Aecore.Chain.ChainState do
 
   @spec validate_tx(SignedTx.t(), chainstate(), integer()) :: {boolean(), chainstate()}
   defp validate_tx(tx, chainstate, block_height) do
-    try do
```
d-velev: why is the try here removed?

gspasov: That's how the code [has always been](https://github.com/aeternity/elixir-node/blob/c7bee713505ed40e7b1dc2fb4c89fa359704a268/apps/aecore/lib/aecore/chain/block_validation.ex#L119)

d-velev: https://imgur.com/a/RilxQ this is what I see, a catch without a try

d-velev: nevermind, i see what it is now

gspasov: Yes, this is a [implicit](https://github.com/rrrene/credo/issues/224#issue-188811717) version of try/catch

PR #320, opened 2018-03-29
Gh 304

0legIv:
Added check for nonce to always be bigger than previous for all DataTx. Added new test for it

### DIFF
--- a/apps/aecore/lib/aecore/structures/data_tx.ex
+++ b/apps/aecore/lib/aecore/structures/data_tx.ex
@@ -99,8 +99,8 @@ defmodule Aecore.Structures.DataTx do
     end
   end
 
-  def is_nonce_valid?(accounts, tx) do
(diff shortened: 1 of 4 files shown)


apps/aecore/lib/aecore/structures/data_tx.ex
```diff
@@ -72,6 +73,11 @@ defmodule Aecore.Structures.DataTx do
     accounts_state = chainstate.accounts
     tx_type_state = Map.get(chainstate, tx.type, %{})
 
+    if !nonce_valid?(accounts_state, tx) do
+      Logger.error("Nonce is too small")
```
d-velev: Looks good, just one thing - if we log before throwing here I think we will log the same thing 2 times since the caught message is logged in a different place in this PR https://github.com/aeternity/elixir-node/pull/319/files#diff-d36d6614178f722c7eba656b1410801fR97

Issue #261, opened 2018-03-12
adjust timestamp validation


thepiwo:
adjust timestamp validation, for the block not to be more than a configurable value in the future (30min as default)

(keep the old timestamp validation code, as I assume this will get changed again in the future)

HristianHristov:
@thepiwo Wasn't this done in #323 ? Why it's still open?

thepiwo:
because the closes annotation has to be typed twice if it should close two issues, which has been done with comma.

PR #323, opened 2018-03-30
[GH-#322, GH-#261] Epoch gossip API, adjust timestamp validation

### LINKED ISSUE #322: Match epoch gossip API
https://aeternity.github.io/epoch-api-docs/#/gossip

d-velev:
closes #322, #261


thepiwo:
One important endpoint in epoch is `/ping` I think

d-velev:
yes, sync works

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -86,28 +86,69 @@ defmodule Aecore.Chain.Worker do
     GenServer.call(__MODULE__, :top_height)
   end
 
+  @spec get_header_by_base58_hash(String.t()) :: Header.t() | {:error, atom()}
+  def get_header_by_base58_hash(hash) do
+    try do
+      decoded_hash = Header.base58c_decode(hash)
+      get_header(decoded_hash)
+    rescue
```
gspasov: What are you rescuing here? I did't see a throw or raise from the above funcs

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -86,28 +86,69 @@ defmodule Aecore.Chain.Worker do
...
   end
 
   @spec get_block_by_base58_hash(String.t()) :: Block.t()
   def get_block_by_base58_hash(hash) do
-    decoded_hash = Header.base58c_decode(hash)
-    get_block(decoded_hash)
+    try do
+      decoded_hash = Header.base58c_decode(hash)
+      get_block(decoded_hash)
+    rescue
```
gspasov: Again, what do you rescue from here?

apps/aehttpserver/lib/aehttpserver/web/controllers/new_tx_controller.ex
```diff
@@ -5,10 +5,14 @@ defmodule Aehttpserver.Web.NewTxController do
...
 
   def new_tx(conn, _params) do
-    conn.body_params
-    |> Serialization.tx(:deserialize)
-    |> Pool.add_transaction()
+    deserialized_tx = Serialization.tx(conn.body_params, :deserialize)
 
-    json(conn, %{:status => :new_tx_added})
+    case Pool.add_transaction(deserialized_tx) do
+      :error ->
```
gspasov: Shouldn't we have the successful case first here?

d-velev: Does it matter?

apps/aecore/lib/aecore/chain/worker.ex
d-velev: the base58c functions don't have different return values, I need to check if it was decoded correctly, if it fails then it returns invalid_hash

gspasov: I only see this crashing if the hash is <= 4bytes so the pattern match in the func will fail, but this could be prevented if we have a guard on the function with the correct size for the hash, `when byte_size(hash) == 32`

apps/aehttpserver/lib/aehttpserver/web/controllers/new_tx_controller.ex
gspasov: That's just the flow of reading the code, usually we have the success case first unless we have to pattern match the error before that.

thepiwo:
@d-velev will you adjust, requested changes?

d-velev:
I will

d-velev:
I'll leave it the way it is, the rescue fragment will do the job, epoch does it in a similar way and even if the hash is too short or something like that you will still get a "Header not found", also the decode and encode take input of different lengths so a check like that won't do.

thepiwo:
@gspasov can you re-review?

thepiwo:
@d-velev do you still plan on adjusting this? https://github.com/aeternity/elixir-node/pull/323#pullrequestreview-109725072

d-velev:
I will adjust it in a moment

PR #324, opened 2018-03-30
[GH-#258] Implementig new Target calculation

### LINKED ISSUE #258: implement difficulty target calculation
Documentation from https://github.com/aeternity/epoch/blob/08b58bb4af575fcf91822b431b1c858b32a27bea/apps/aecore/src/aec_pow.erl

## Target threshold and difficulty

The mining rate is controlled by setting a target threshold. The PoW nonce is accepted if a hash value (the hash of the header for SHA-256, the hash of the solution graph for Cuckoo Cycles, converted to an integer) is below this target threshold.

A lower target represents a harder task (requires the hash to start with a number of zeros).

The target threshold relates to another value: the difficulty. This is proportional to the hardness of the PoW task:

Difficulty = <Target of difficulty 1> / Target, a floating point value.

Bitcoin uses 0x00000000FFFF0000000000000000000000000000000000000000000000000000 as Difficulty 1 target (0x1d00ffff in scientific notation, see below). For Cuckoo Cycle we need a lighter filtering of solutions than for SHA-256 as the basic algorithm is much slower than a simple hash generation, so we use the largest possible value: 0xFFFF000000000000000000000000000000000000000000000000000000000000 (0x2100ffff in scientific notation) as difficulty 1.

We store the current target threshold in the block header in scientific notation. Difficulty is used to select the winning fork of new blocks: the difficulty of a chain of blocks is the sum of the difficulty of each block.

Integers represented in scientific notation:
2^24 * <base-2 exponent + 3> + the first 3 most significant bytes (i.e.,
the significand, see https://en.wikipedia.org/wiki/Significand).
The + 3 corresponds to the length of the significand (i.e., the int value is 0.<significand> * 8^<exponent>). 
https://en.bitcoin.it/wiki/Difficulty#How_is_difficulty_stored_in_blocks.3F

HristianHristov:
Resolves #258


HristianHristov:
@thepiwo  Yes, we have the test for difficulty and test for scientific and integer transformation.

apps/aecore/lib/aecore/pow/hashcash.ex
```diff
@@ -1,9 +1,11 @@
 defmodule Aecore.Pow.Hashcash do
+  alias Aeutil.Scientific
+  use Bitwise
+
   @moduledoc """
```
gspasov: @moduledoc should be on the second row

apps/aeutil/lib/scientific.ex
```diff
@@ -0,0 +1,67 @@
...
+        false
+    end
+  end
+
+  @spec break_scientific(integer()) :: tuple()
+  def break_scientific(scientific) do
+    significand_mask = bsl(1, 24) - 1
+
+    exp =
+      bxor(scientific, significand_mask)
```
gspasov: Add `scentific` var to the pipe

apps/aecore/lib/aecore/pow/hashcash.ex
```diff
@@ -18,10 +20,43 @@ defmodule Aecore.Pow.Hashcash do
...
   def verify(block_header_hash, difficulty) do
-    block_header_hash
-    |> Bits.extract()
-    |> Enum.take_while(fn bit -> bit == 0 end)
-    |> Enum.count() >= difficulty
+    {exp, significand} = Scientific.break_scientific(difficulty)
+    length = byte_size(block_header_hash)
+    zeros = 8 * max(0, length - exp)
+
+    case exp do
```
d-velev: This case looks weird, a cond is better

HristianHristov: I think it's better this way, there is no use switching to cond.

d-velev: This is literally a case transformed to be a cond, there is no matching on exp and you are just checking a bool expression, it's a trivial change so whatever

PR #326, opened 2018-03-30
[GH-#259] Limited maximum txs in a block

### LINKED ISSUE #259: adjust max tx per block
Instead of limiting block size by bytes of included transactions we will limit the number of transactions. Make configurable max amount of transactions per block.

(keep the old limitation by byte size code, as I assume this will get changed again in the future)

HristianHristov:
Resolves #259

### DIFF
--- a/apps/aecore/config/dev.exs
+++ b/apps/aecore/config/dev.exs
@@ -112,6 +112,7 @@ bytes_per_token =
 config :aecore, :tx_data,
   lock_time_coinbase: 10,
   miner_fee_bytes_per_token: bytes_per_token,
-  pool_fee_bytes_per_token: 100
+  pool_fee_bytes_per_token: 100,
+  max_txs_per_block: 100
 
 config :aecore, :block, max_block_size_bytes: 500_000
--- a/apps/aecore/lib/aecore/miner/worker.ex
+++ b/apps/aecore/lib/aecore/miner/worker.ex
@@ -231,7 +231,7 @@ defmodule Aecore.Miner.Worker do
 
       difficulty = Difficulty.calculate_next_target(blocks_for_difficulty_calculation)
 
-      txs_list = Map.values(Pool.get_pool())
+      txs_list = get_pool_values()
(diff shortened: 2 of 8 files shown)


apps/aecore/config/dev.exs
```diff
@@ -112,6 +112,7 @@ bytes_per_token =
 config :aecore, :tx_data,
   lock_time_coinbase: 10,
```
thepiwo: can you remove `lock_time_coinbase` from the configs?

HristianHristov:
@thepiwo  Ok, will be done

thepiwo:
use the default merge commit message in future, so github knows whats going on

apps/aecore/lib/aecore/miner/worker.ex
```diff
@@ -306,87 +283,24 @@ defmodule Aecore.Miner.Worker do
 
   ## Internal
 
+  defp get_pool_values() do
+    pool_values = Map.values(Pool.get_pool())
+    max_txs_for_block = Application.get_env(:aecore, :tx_data)[:max_txs_per_block] - 1
+
+    if Enum.empty?(pool_values) or length(pool_values) < max_txs_for_block do
```
d-velev: I think the Enum.empty check isn't needed since length will just return 0 if it's empty. Also, the check should be length(pool_values) <= max_txs_for_block since right now it's checking if the length is < 99 meaning that there will be 99 transactions at most including the coinbase.

apps/aecore/lib/aecore/miner/worker.ex
```diff
@@ -306,87 +283,24 @@ defmodule Aecore.Miner.Worker do
...
+      pool_values
+    else
+      Enum.slice(pool_values, max_txs_for_block)
```
d-velev: Does the slice work here? I checked the docs for slice, it either takes starting index and amount to slice or a range from which to slice (something like 0..10) so passing 2 parameters and the 2nd being just an integer shouldn't work

HristianHristov: You are right, it should be 0..max_txs_for_block

PR #327, opened 2018-03-30
[GH-#303] refactor fee deduction update to be in data tx

### LINKED ISSUE #303: refactor fee deduction update to be in data tx

0legIv:
Resolves #303


apps/aecore/test/aecore_tx_test.exs
```diff
@@ -137,4 +137,34 @@ defmodule AecoreTxTest do
...
+    fee = 40
+    # Balance of acc1 is more than amount and fee, send tokens to acc2
+    payload = %{receiver: acc2, amount: amount}
+    tx_data = DataTx.init(SpendTx, payload, acc1, fee, 1)
+    priv_key = Wallet.get_private_key("m/1")
+    {:ok, signed_tx} = SignedTx.sign_tx(tx_data, priv_key)
+    :ok = Pool.add_transaction(signed_tx)
+    :ok = Miner.mine_sync_block_to_chain()
+    # the balance of acc1 and acc2 is not changed because amount + fee > balance of acc1
+    assert Chain.chain_state().accounts[acc2] == nil
```
gspasov: Maybe assert again that acc1 has balance == 80, so that we are sure the transaction was not successful and the balance of acc1 didn't change

d-velev:
Since the oracles PR was merged, this has to be changed for the oracle transactions as well

apps/aecore/lib/aecore/structures/spend_tx.ex
```diff
@@ -92,7 +93,7 @@ defmodule Aecore.Structures.SpendTx do
 
     new_sender_account_state =
       sender_account_state
-      |> deduct_fee(fee)
+      |> DataTx.deduct_fee(fee)
```
cytadela8: I think this function is very similar to Account.transaction_in and out. Maybe it would be better to place it inside Aecore.Structures.Account? It operates on Account objects.

apps/aecore/lib/aecore/structures/transaction.ex
```diff
@@ -77,6 +77,4 @@ defmodule Aecore.Structures.Transaction do
               block_height :: non_neg_integer(),
               tx_type_state :: map()
             ) :: :ok | {:error, reason}
-
-  @callback deduct_fee(ChainState.account(), fee :: non_neg_integer()) :: ChainState.account()
```
cytadela8: If we leave deduct_fee in SpendTx then we should readd this callback.

cytadela8:
@thepiwo This is ready for merge, but it only adds one test, so it doesn't look like it "Resolves #303". Please decide what to do.

thepiwo:
I think its okay for now, do you want to propose a follow-up task @cytadela8 ?

PR #330, opened 2018-04-02
[GH-#282] Adjusted transaction json serialization

### LINKED ISSUE #282: adjust transaction json serialization
depended on #280 and #281 

### signed tx
 - [x] "tx": tx_data
 - [x] "block_height": height,
 - [x] "block_hash": block_hash,
 - [x] "hash" tx_hash,
 - [x] "signatures": [signature] 

### spend tx data
 - [x] "sender": sender_pubkey,
 - [x] "recipient": recipient_pubkey,
 - [x] "amount": amount,
 - [x] "fee: fee,
 - [x] "nonce": nonce,
 - [x] "vsn": version

Artur64:
Resolves #282 and fixes a bug in add_proof_to_txs function in Pool module


Artur64:
@thepiwo

thepiwo:
I am about to check

apps/aecore/lib/aecore/structures/account.ex
```diff
@@ -78,7 +78,11 @@ defmodule Aecore.Structures.Account do
   end
 
   def base58c_encode(bin) do
-    Bits.encode58c("ak", bin)
+    if bin == nil do
+      nil
```
thepiwo: why do we need this? seems like a case for possible errors

apps/aeutil/lib/serialization.ex
```diff
@@ -247,4 +248,32 @@ defmodule Aeutil.Serialization do
   end
 
   def deserialize_value(value, _), do: value
+
+  def serialize_txs_info_to_json(txs_info) when is_list(txs_info) do
+    serialize_txs_info_to_json(txs_info, [])
+  end
+
+  def serialize_txs_info_to_json([h | t], acc) do
```
thepiwo: add `@spec`

apps/aeutil/lib/serialization.ex
```diff
@@ -247,4 +248,32 @@ defmodule Aeutil.Serialization do
   end
 
   def deserialize_value(value, _), do: value
+
+  def serialize_txs_info_to_json(txs_info) when is_list(txs_info) do
```
thepiwo: add `@spec`

apps/aeutil/lib/serialization.ex
```diff
@@ -247,4 +248,32 @@ defmodule Aeutil.Serialization do
...
+      tx: %{
+        sender: Account.base58c_encode(h.sender),
+        recipient: Account.base58c_encode(h.payload.receiver),
+        amount: h.payload.amount,
+        fee: h.fee,
+        nonce: h.nonce,
+        vsn: h.payload.version
+      },
+      block_height: h.block_height,
+      block_hash: Header.base58c_encode(h.block_hash),
```
thepiwo: is `h.block_hash` existent? can't know the type, as there is no spec

apps/aeutil/lib/serialization.ex
```diff
@@ -247,4 +248,32 @@ defmodule Aeutil.Serialization do
...
+      hash: SignedTx.base58c_encode_root(h.txs_hash),
+      signatures: [base64_binary(h.signature, :serialize)]
+    }
+
+    acc = acc ++ [json_response_struct]
+    serialize_txs_info_to_json(t, acc)
+  end
+
+  def serialize_txs_info_to_json([], acc) do
```
thepiwo: add `@spec`

apps/aecore/lib/aecore/structures/account.ex
Artur64: Yes, because sometimes when we serialize a coin-based transaction - the senders field will be nil as well as signature

thepiwo: okay, didn't think of that

Artur64:
Found a bug in a case when user will want to get proofs included to his transactions, will try to fix it, i suppose we should wait with this pull request a little longer as i want to try to fix the problem with merkle proofs.

apps/aeutil/lib/serialization.ex
Artur64: Its a recursive function, so the spec of first function applies to the rest as well.

Artur64:
@thepiwo  I think i'm ready with bug, would like to get a review over this PR

PR #331, opened 2018-04-03
[GH-#260] Added minimum fee

### LINKED ISSUE #260: adjust minimum fee
adjust minimum fee, greater equals a configurable value

(keep the fee calculations code, as I assume this will get changed again in the future)

HristianHristov:
Resolves #260


apps/aecore/lib/aecore/structures/oracle_extend_tx.ex
```diff
@@ -92,7 +94,7 @@ defmodule Aecore.Structures.OracleExtendTx do
       !Map.has_key?(registered_oracles, sender) ->
         throw({:error, "Account isn't a registered operator"})
 
-      fee < calculate_minimum_fee(tx.ttl) ->
+      fee < calculate_minimum_fee(tx.ttl) && fee < @minimum_fee ->
```
cytadela8: There should be `||` here for correct behavior probably.

thepiwo: gues this wouldn't be compatible if our fee is lower than epochs

d-velev:
I see that the fee requirement is added to oracle transactions as well, is that the case in epoch? It would make their base oracle fee settings useless

HristianHristov:
@d-velev  So we should remove the minimum_fee from the oracles ?

d-velev:
@HristianHristov yes they have a fee requirement of their own

apps/aecore/lib/aecore/structures/spend_tx.ex
```diff
@@ -147,7 +147,8 @@ defmodule Aecore.Structures.SpendTx do
             Application.get_env(:aecore, :tx_data)[:miner_fee_bytes_per_token]
         end
 
-      tx.data.fee >= Float.floor(tx_size_bytes / bytes_per_token)
+      tx.data.fee >= Float.floor(tx_size_bytes / bytes_per_token) &&
```
thepiwo: this still uses byte size fee calculation

thepiwo: @HristianHristov what do you think about this?

HristianHristov: Yes It seems I forgot about this one, will be removed.

thepiwo:
@cytadela8 can you re-review?

PR #333, opened 2018-04-04
[GH-#325] check byte sizes for transactions

### LINKED ISSUE #325: check byte sizes for transactions
check byte size for all fixed length binary inputs into a transaction, for spend transactions this would be sender and receiver public key length

0legIv:
Resolves #325


apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -119,7 +119,7 @@ defmodule Aecore.Chain.BlockValidation do
   def validate_block_transactions(block) do
     block.txs
     |> Enum.map(fn tx ->
-      SignedTx.is_coinbase?(tx) || SignedTx.is_valid?(tx)
+      SignedTx.is_coinbase?(tx) || (SignedTx.is_valid?(tx) && SignedTx.valid_byte_size?(tx))
```
thepiwo: can you include `valid_byte_size` in `is_valid`

apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -32,6 +32,38 @@ defmodule Aecore.Structures.SignedTx do
     key == nil && signature == nil
   end
 
+  @spec valid_byte_size?(SignedTx.t()) :: boolean()
+  def valid_byte_size?(%SignedTx{data: %{payload: %{oracle_address: address}}}) do
```
thepiwo: I think there can be data_tx type matching and the specific validation happens inside the tx structure

apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -32,6 +32,38 @@ defmodule Aecore.Structures.SignedTx do
...
+      true
+    else
+      Logger.error("Wrong sender or receiver key size")
+      false
+    end
+  end
+
+  @spec valid_byte_size?(SignedTx.t()) :: boolean()
+  def valid_byte_size?(_tx) do
+    true
```
thepiwo: can the signature size be checked, is there an allowed maximum?

apps/aecore/lib/aecore/structures/data_tx.ex
```diff
@@ -44,10 +44,15 @@ defmodule Aecore.Structures.DataTx do
   use ExConstructor
 
   @spec init(tx_types(), payload(), binary(), integer(), integer()) :: DataTx.t()
-  def init(type, payload, sender, fee, nonce) do
+  def init(type, payload, sender, fee, nonce) when sender == nil or byte_size(sender) == 33 do
```
cytadela8: We shouldn't hardcode pubkey size like that. The constant should be defined somewhere.

thepiwo: https://en.wikipedia.org/wiki/Magic_number_(programming)

apps/aecore/lib/aecore/structures/data_tx.ex
```diff
@@ -44,10 +48,16 @@ defmodule Aecore.Structures.DataTx do
   use ExConstructor
 
   @spec init(tx_types(), payload(), binary(), integer(), integer()) :: DataTx.t()
-  def init(type, payload, sender, fee, nonce) do
+  def init(type, payload, sender, fee, nonce)
+      when sender == nil or byte_size(sender) == @pub_key_size do
```
thepiwo: you can use `Wallet.get_pub_key_size()`, just define constants with `@`

apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -32,9 +33,40 @@ defmodule Aecore.Structures.SignedTx do
     key == nil && signature == nil
   end
 
+  @spec valid_byte_size?(DataTx.t()) :: boolean()
+  def valid_byte_size?(%DataTx{payload: %{oracle_address: address}}) do
```
thepiwo: oracle specific check should be in oracle tx files

apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -32,9 +33,40 @@ defmodule Aecore.Structures.SignedTx do
...
+    if Wallet.key_size_valid?(address) do
+      true
+    else
+      Logger.error("Wrong oracle address size")
+      false
+    end
+  end
+
+  @spec valid_byte_size?(DataTx.t()) :: boolean()
+  def valid_byte_size?(%DataTx{
```
thepiwo: this check should be in data tx file

apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -32,9 +33,40 @@ defmodule Aecore.Structures.SignedTx do
...
+    if Wallet.key_size_valid?(sender) && Wallet.key_size_valid?(receiver) do
+      true
+    else
+      Logger.error("Wrong sender or receiver key size")
+      false
+    end
+  end
+
+  @spec valid_byte_size?(DataTx.t()) :: boolean()
+  def valid_byte_size?(_data) do
```
thepiwo: this case should be true, any other transactions should be rejected or fail

apps/aecore/lib/aecore/structures/signed_tx.ex
```diff
@@ -32,9 +33,40 @@ defmodule Aecore.Structures.SignedTx do
...
+    true
+  end
+
   @spec is_valid?(SignedTx.t()) :: boolean()
   def is_valid?(%SignedTx{data: data} = tx) do
-    if Signing.verify(Serialization.pack_binary(data), tx.signature, data.sender) do
+    if Signing.verify(Serialization.pack_binary(data), tx.signature, data.sender) &&
+         valid_byte_size?(data) do
```
thepiwo: `valid_byte_size?` can be called in each specific transactions `is_valid?`

apps/aecore/lib/aecore/structures/spend_tx.ex
```diff
@@ -7,12 +7,15 @@ defmodule Aecore.Structures.SpendTx do
...
   alias Aecore.Chain.ChainState
-  alias Aecore.Wallet
+  alias Aecore.Wallet.Worker, as: Wallet
   alias Aecore.Structures.Account
   alias Aecore.Txs.Pool.Worker, as: Pool
 
   require Logger
 
+  # Gets valid public key size - 33
+  @pub_key_size Wallet.get_pub_key_size()
```
thepiwo: same as in data tx

apps/aecore/lib/aecore/structures/data_tx.ex
0legIv: The problem is that I can't use remote functions in Guards, that's why I use constant @pub_key_size, and initialize it with the function Wallet.get_pub_key_size() in the beginning of module.

apps/aecore/lib/aecore/structures/signed_tx.ex
0legIv: Maybe in SpendTx? Because in this function I check the byte_size for SpendTx, and DataTx is something common for all txs types.

0legIv: Another way, I can put byte_size validations for all types of txs in DataTx module. Since it is common for all txs types, probably it would be better to handle all the checks there.

apps/aecore/lib/aecore/structures/data_tx.ex
thepiwo: do you see around using guards, otherwise let's keep it like this

apps/aecore/lib/aecore/structures/signed_tx.ex
thepiwo: each check should be in the module where the field is defined initially

thepiwo:
@0legIv can the signature length be checked, maybe with some max?

thepiwo:
@cytadela8 can you re-review and merge?

PR #334, opened 2018-04-05
[GH-#269] Replace sha256 with blake2 for Oracles

### LINKED ISSUE #269: replace sha256 with blake2 as default hash function
research where blake2 is used in the core, other than oracles https://www.pivotaltracker.com/epic/show/3860576, document findings here

do after #212 is merged

HristianHristov:
Resolves #269


HristianHristov:
Sorry for missunderstanding, but the issue name was for the oracles: https://github.com/aeternity/elixir-node/issues/269
Will be done soon.

thepiwo:
oh, then it was my misunderstanding, but as oracles are merged, we can now adjust it for everything

necessary.sh
```diff
@@ -0,0 +1,5 @@
+curl -fsSL -o libsodium-src.tar.gz "https://github.com/jedisct1/libsodium/releases/download/1.0.16/libsodium-1.0.16.tar.gz"
+mkdir libsodium-src
+tar -zxf libsodium-src.tar.gz -C libsodium-src --strip-components=1
+cd libsodium-src/
+sudo ./configure &&  sudo make -j$(nproc) &&  sudo make install && sudo ldconfig
```
cytadela8: Do we really need to download libsodium sources? Isn't there an ubuntu package with this lib?

cytadela8: Also: you shouldn't run `make -j$(nproc)` as root

cytadela8: If we really need to download and compile libsodium then maybe we can avoid installing it to system? You already use `LD_LIBRARY_PATH` to some path. This configured properly should be enough.

thepiwo: compilation is needed, as repos are outdated

thepiwo: but this file is not needed, a note in the readme is sufficient I think

thepiwo: I did this in the `Dockerfile`

HristianHristov: Ok I will make a note in the readme and remove this script

README.md
```diff
@@ -27,6 +27,9 @@ Make sure you have installed the following packages to make sure that the Wallet
 sudo apt-get install autoconf autogen
 sudo apt-get install libtool
 sudo apt-get install libgmp3-dev
+wget -O libsodium-src.tar.gz https://github.com/jedisct1/libsodium/releases/download/1.0.16/libsodium-1.0.16.tar.gz
+ mkdir libsodium-src && tar -zxf libsodium-src.tar.gz -C libsodium-src --strip-components=1
+  cd libsodium-src && ./configure --prefix=$HOME/.libsodium && make -j$(nproc) && make install && cd ..
```
thepiwo: installation must be sudo

apps/aeutil/lib/hash.ex
```diff
@@ -0,0 +1,22 @@
...
+    :sha3.hash(@hash_bytes_size * 8, data)
+  end
+
+  @spec hash_sha256(binary()) :: binary()
+  def hash_sha256(data) when is_binary(data) do
+    :crypto.hash(:sha256, data)
+  end
+
+  @spec hash(binary()) :: binary()
+  defdelegate hash(opts), to: __MODULE__, as: :hash_blake2b
```
d-velev: If this exists then Hash.hash_blake2b should be replaced with Hash.hash right?

d-velev: this was commented on, didn't see

thepiwo:
@HristianHristov what is the state on this?

HristianHristov:
@thepiwo Currently I'm trying to fix the Cuckoo tests, but for some reason I can't use the generichash function, because it couldn't find the :enacl_nif.so

thepiwo:
@d-velev @cytadela8 can you re-review?

PR #337, opened 2018-04-05
[GH-#283] Adjusted transaction fields encoding

### LINKED ISSUE #283: adjust transaction fields encoding
dependend on #282 

- [x] hash as base58c "th"
- [x] signatures as list of base58c "sg"
- [x] sender as base58c "ak"
- [x] recipient as base58c "ak"

Artur64:
Resolves #283


apps/aeutil/lib/serialization.ex
```diff
@@ -266,6 +266,12 @@ defmodule Aeutil.Serialization do
   end
 
   defp serialize_txs_info_to_json([h | t], acc) do
+    tx_hash =
```
thepiwo: either rename this, as it is not the hash or include hashing with a pipe

cytadela8: Also you should call some `hash_tx` function here so hashing is defined in one place. I think we have `SignedTx.hash_tx`

thepiwo:
@cytadela8 can your re-review and merge?

cytadela8:
@Artur64 please resolve conflicts

PR #343, opened 2018-04-10
[GH-#329] Removed try/catch statements and refactored error handiling

### LINKED ISSUE #329: Remove try/catch blocks
In Elixir `try/catch` blocks are **not** recommended. And are used only when they cannot be avoided in some other way. Read [here](https://elixir-lang.org/getting-started/try-catch-and-rescue.html#throws) and [here](https://stackoverflow.com/questions/40280887/elixir-try-catch-vs-try-rescue#comment67823125_40280887)

gspasov:
resolves #329


gspasov:
Yes, all `throw statements` are refactored to return `:error` and all the `Logger messages` now have the `module name` in front of the error so it's easier to see where it came from. An `inspect(data)` could be included in the error handling as well, yes.

apps/aecore/lib/aecore/miner/worker.ex
```diff
@@ -262,7 +260,7 @@ defmodule Aecore.Miner.Worker do
       create_block(top_block, chain_state, difficulty, valid_txs, timestamp)
     catch
```
d-velev: did you miss this try?

apps/aecore/lib/aecore/chain/chain_state.ex
```diff
@@ -23,34 +24,48 @@ defmodule Aecore.Chain.ChainState do
...
-        else
-          throw({:error, "Invalid transaction"})
-        end
+    {:ok, Map.put(chainstate, :accounts, new_accounts_state)}
+  end
 
-      true ->
-        throw({:error, "Invalid transaction"})
+  def apply_transaction_on_state(%{data: %{sender: sender}} = tx, chainstate, block_height)
+      when not is_nil(sender) do
```
rdimitrovbg: Would not it be better if the guard was `when is_binary(sender)` ?
Because `{}` is not nil to and in the same way the third match of function is unreachable.

apps/aecore/lib/aecore/structures/data_tx.ex
```diff
@@ -72,25 +71,52 @@ defmodule Aecore.Structures.DataTx do
...
+    case tx.payload
+         |> tx.type.init()
+         |> tx.type.process_chainstate!(
+           tx.sender,
+           tx.fee,
+           tx.nonce,
+           accounts_state,
+           tx_type_state
+         ) do
+      {:error, reason} ->
```
rdimitrovbg: Do not you think the more valuable information should be kept first place instead of `{: error, reason}` ?

gspasov: Yes, but in this case I can only pattern match the :error, since the value i need is `{new_accounts_state, new_tx_type_state}`

thepiwo:
@meivantodorov @rdimitrovbg @cytadela8 can you review?

apps/aecore/lib/aecore/chain/chainstate.ex
```diff
@@ -32,43 +32,60 @@ defmodule Aecore.Chain.Chainstate do
...
+  def calculate_and_validate_chain_state(txs, chainstate, block_height) do
     txs
     |> Enum.reduce(chainstate, fn tx, chainstate ->
-      apply_transaction_on_state!(tx, chainstate, block_height)
+      case apply_transaction_on_state(tx, chainstate, block_height) do
+        {:ok, updated_chainstate} ->
+          updated_chainstate
+
+        {:error, _reason} ->
+          chainstate
```
cytadela8: (important) If any transaction fails this whole function has to fail.

cytadela8: (Failing transactions shouldn't be skipped, it should cause block validation failure)

apps/aecore/lib/aecore/chain/chainstate.ex
```diff
@@ -77,27 +94,23 @@ defmodule Aecore.Chain.Chainstate do
...
-    {true, apply_transaction_on_state!(tx, chainstate, block_height)}
-  catch
-    {:error, _} -> {false, chainstate}
+  @doc """
+  Goes through all the transactions and only picks the valid ones
+  """
+  @spec get_valid_txs(list(), Chainstate.t(), non_neg_integer()) :: list()
+  def get_valid_txs(txs_list, chainstate, block_height) do
+    txs_list
+    |> List.foldl([], fn tx, valid_txs_list ->
```
cytadela8: (important) If an transaction is valid next transaction should be executed on updated chainstate.

Explenation:
Transactions can have dependencies between each other. For example let's say there was created a two transactions using the same funds. They both are valid independently, but together they aren't. This function is supposed to return first one, because we use this function to filter transactions so we can create a valid block.

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -448,19 +446,19 @@ defmodule Aecore.Chain.Worker do
 
     top_chain_state =
       case Persistence.get_all_accounts_chain_states() do
-        chain_states when chain_states == %{} -> state.blocks_data_map[top_hash].chain_state
+        %{} -> state.blocks_data_map[top_hash].chain_state
```
cytadela8: AFAIK this doesn't work as intended. `%{}` matches all maps so the first case is always used.

cytadela8: (The difference is between how matching and `==` works in Elixir/Erlang. `==` is stricter)

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -83,41 +83,37 @@ defmodule Aecore.Peers.Sync do
 
   @spec add_peer_blocks_to_sync_state(String.t(), binary()) :: :ok
   def add_peer_blocks_to_sync_state(peer_uri, from_block_hash) do
-    if Chain.has_block?(from_block_hash) do
-      remove_running_task(peer_uri)
+    with false <- Chain.has_block?(from_block_hash),
```
cytadela8: The if was cleaner I think, but this can stay as is.

apps/aecore/lib/aecore/tx/data_tx.ex
```diff
@@ -49,66 +49,109 @@ defmodule Aecore.Tx.DataTx do
...
+          type: type,
+          payload: payload,
+          sender: sender,
+          fee: fee,
+          nonce: nonce
+        } = tx,
+        %{accounts: accounts} = chainstate,
+        block_height
+      ) do
+    sender_account_state = Account.get_account_state(accounts, sender)
```
cytadela8: I would suggest invoking `validate_nonce` and `validate_sender` inside this function instead of in `Aecore.Chain.Chainstate`

apps/aecore/test/aecore_chain_state_test.exs
```diff
@@ -76,7 +76,10 @@ defmodule AecoreChainstateTest do
   def apply_txs_on_state!(txs, chainstate, block_height) do
     txs
     |> Enum.reduce(chainstate, fn tx, chainstate ->
-      Chainstate.apply_transaction_on_state!(tx, chainstate, block_height)
+      case Chainstate.apply_transaction_on_state(tx, chainstate, block_height) do
+        {:ok, new_state} -> new_state
+        {:error, _reason} -> chainstate
```
cytadela8: We should probably fail the test or report some error in `{:error, _}` case

apps/aeutil/lib/parser.ex
```diff
@@ -5,9 +5,9 @@ defmodule Aeutil.Parser do
 
   def to_atom!(key) when is_atom(key), do: key
   def to_atom!(key) when is_binary(key), do: String.to_atom(key)
-  def to_atom!(_key), do: throw("Key is neither atom nor string")
+  def to_atom!(key), do: key
```
cytadela8: This is a behavior change and I'm not sure how this will effect our code. If we want this behaviour change you shold remove the `!` in function name.

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -15,59 +15,77 @@ defmodule Aecore.Chain.BlockValidation do
...
-    root_hash = Chainstate.calculate_root_hash(new_chain_state)
+        root_hash = Chainstate.calculate_root_hash(new_chain_state)
 
-    target =
-      Difficulty.calculate_next_difficulty(
-        new_block.header.time,
-        blocks_for_target_calculation
-      )
+        server = self()
+        work = fn -> Cuckoo.verify(new_block.header) end
```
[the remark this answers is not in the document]
gspasov: I've missed that

apps/aecore/lib/aecore/chain/worker.ex
gspasov: Yes, you are correct but see [here](https://stackoverflow.com/questions/27493029/elixir-error-erlang-error-guard-expr?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa) to see that the master code won't work as well apparently

apps/aecore/lib/aecore/tx/data_tx.ex
gspasov: They are in the ChainState module for readability reasons. That way the code has good flow of reading, and understandability. If they are inside the DataTx it's not visible that they are called.

apps/aeutil/lib/parser.ex
gspasov: This is used only in the `Serialization` module for converting string map keys to atom keys. I will remove the `!` though.

apps/aecore/lib/aecore/peers/sync.ex
gspasov: Well, having in mind that the `Sync` will be rewritten so this is not important anymore.

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -15,59 +15,77 @@ defmodule Aecore.Chain.BlockValidation do
...
-  @spec calculate_and_validate_block!(
+  @spec calculate_and_validate_block(
           Block.t(),
           Block.t(),
           Chainstate.account_chainstate(),
           list(Block.t())
-        ) :: {:error, term()} | :ok
-
-  def calculate_and_validate_block!(
+        ) :: {:ok, Chainstate.chainstate()} | {:error, String.t()}
```
rdimitrovbg: Here in the same way. If it is the Chainstate itself should be `Chainstate.t()`

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -195,50 +188,55 @@ defmodule Aecore.Chain.Worker do
   end
 
   @spec get_block_by_height(non_neg_integer(), binary() | nil) ::
```
rdimitrovbg: Here the specification is wrong.

apps/aecore/lib/aecore/chain/chainstate.ex
gspasov: I think there is a fundamental problem with the use of this function. First I don't see why this function would crash inside the [`miner`](https://github.com/aeternity/elixir-node/blob/03880f4a5ab5fa25ad0033809827618e4be1d500/apps/aecore/lib/aecore/miner/worker.ex#L311) because we pass already filtered transactions to it. So why would we need to check them again?

If the crashing is ment for the [`BlockValiedation`]() usage then maybe this should be a seperate func.

I'll give you a roadmap of what we are doing atm and you tell me what you think about it. I find it strange to say the least:
1st -> We call [`Miner.mine_sync_block_to_chain()`](https://github.com/aeternity/elixir-node/blob/03880f4a5ab5fa25ad0033809827618e4be1d500/apps/aecore/lib/aecore/miner/worker.ex#L77)
2nd -> We call [`Miner.candidate()`](https://github.com/aeternity/elixir-node/blob/03880f4a5ab5fa25ad0033809827618e4be1d500/apps/aecore/lib/aecore/miner/worker.ex#L225)
3rd -> We [`filter the invalid txs`](https://github.com/aeternity/elixir-node/blob/03880f4a5ab5fa25ad0033809827618e4be1d500/apps/aecore/lib/aecore/miner/worker.ex#L244)
4th -> We call [`Miner.create_block`](https://github.com/aeternity/elixir-node/blob/03880f4a5ab5fa25ad0033809827618e4be1d500/apps/aecore/lib/aecore/miner/worker.ex#L307) passing those valid txs which we validated by fee and added the coinbase tx as well
5th -> We call [`Chain.calculate_and_validate_chain_state()`](https://github.com/aeternity/elixir-node/blob/03880f4a5ab5fa25ad0033809827618e4be1d500/apps/aecore/lib/aecore/miner/worker.ex#L310) which I don't see a reason to fail having in mind that the txs are already filtered and shouldn't crash the function
6th -> We get the returned block from the `Miner.candidate()` func and we pass it to [`Chain.add_block`](https://github.com/aeternity/elixir-node/blob/03880f4a5ab5fa25ad0033809827618e4be1d500/apps/aecore/lib/aecore/miner/worker.ex#L81)
7th -> Inside the Chain.add_block() func we call [`BlockValidation.calculate_and_validate_block()`](https://github.com/aeternity/elixir-node/blob/09c7fe3331727e278030d923328345bb8ed03a06/apps/aecore/lib/aecore/chain/block_validation.ex#L26) which inside itself calls `Chain.calculate_and_validate_chain_state()` again, where we pass the same filtered transactions we got from `Miner.candidate()` func. And then again why do we need to call `Chain.calculate_and_validate_chain_state()` when we just did so inside the [`Miner.create_block()`](https://github.com/aeternity/elixir-node/blob/03880f4a5ab5fa25ad0033809827618e4be1d500/apps/aecore/lib/aecore/miner/worker.ex#L307) func

So I don't see any reason why `Chain.calculate_and_validate_chain_state()` would creash. If it does then the problem is in the filtering of the invalid txs, because when we filter the invalid txs we call the same function we call inside the `Chain.calculate_and_validate_chain_state()` which is `Chain.apply_transaction_on_state()`.

Correct me if I'm missing something here.

cytadela8: You are missing that the block may come from the network and Chain.add_block is supposed to validate the block before adding it. (The block isn't fully validated before reaching this function)

`Chain.calculate_and_validate_chain_state()` implements our full validation of transactions logic. Before if any transaction would be invalid for any reason this function would throw. Now I understand we want to use return values instead of throwing due to performance reasons. So we should return something like `:error` or `{:error, reason}` in case of failure of any transaction.

As for the miner: the miner validates new blocks twice or maybe three times. This could be optimized, but executing different logic for blocks from miner and those from network may create harder to debug problems (if miner creates invalid blocks). The overhead of validating blocks twice should be negligible in practice compared to cost of running miner. Even in our tests.

As I said briefly in the oryginal comment: block with any invalid transaction is invalid. So failure (of one of the checks) when applying one of the transcriptions to chainstate should cause whole block validation to fail.

apps/aecore/lib/aecore/chain/worker.ex
cytadela8: Interesting.

apps/aecore/lib/aecore/tx/data_tx.ex
cytadela8: If they are inside DataTx, DataTx is more self contained, ChainState "knows less about what DataTx does" and for me personally it has better flow of reading.

apps/aecore/lib/aecore/chain/chainstate.ex
gspasov: Well okay, then why don't we change the `Chain.calculate_and_validate_chain_state()` to not check the txs having in mind that we already have such a function. Duplicating code is not a good approach either. We could then chain the functions to go after one another in the `BlockValidation.validate_and_calculate_block()` and don't check twice the txs in the Miner. Another idea is to add an atom to the `BlockValidation` func to check weather we are comming from the miner of from the network and not call the same funcs again. But I think this should be a seperate issue having in mind that it will change some backbone code. It should be more visible than doing it in this branch.

cytadela8: What checking function do you have in mind? There are 3 functions executions for each transaction (and I don't think we should add more):
- `is_valid?` - checks transactions internals
- `preprocess_check` - checks if transaction can execute on top of given chainstate
- `process_chainstate` - executes transaction of top of given chainstate

Generally it's not possible to validate transactions without executing them due to dependencies.

It's not possible to validate a list of transactions without executing them due to potential dependencies. If you have a list of transactions [T1,T2, T3, T4, ...] and a chainstate C. To validate T3 you have to apply T1 and T2 to C and execute preporocess_check with resulting chainstate on T3. Due to that validating a list of transactions has a side effect of creating chainstate with all of the transactions executed.

As for the Miner again: I think even currently there exists a function in `Aecore.Chain` where you can supply already validated block to be included.

cytadela8: If we call `Chain.calculate_and_validate_chain_state()` twice in some flow then we can fix that. As for the return value: if we would remove returning errors from `Chain.calculate_and_validate_chain_state()`, then we would have to calculate transactions twice for validation of blocks from network.

PR #344, opened 2018-04-10
[GH-#273] changed miner to mean28s-generic

### LINKED ISSUE #273: adjust miner (prod) configuration
research and adjust miner configuration to epoch implementation (in prod configuration)

HristoHr:
Close #273


thepiwo:
maybe we can use the mean28 binary, we already have. @meivantodorov do you the difference to mean28**s-generic**?

meivantodorov:
I have not tested the mean miner, but I think that we can use our `mean` There is `generic` in the `epoch` project, because they are compiling one with avx2 instruction requirements. At some point I will check that.

thepiwo:
changes will be included in #347

Issue #253, opened 2018-03-12
implement patricia merkle trees


thepiwo:
docs (maybe addtitonal research is needed):
 - https://github.com/ethereum/wiki/wiki/Patricia-Tree
 - https://ethereum.stackexchange.com/questions/6415/eli5-how-does-a-merkle-patricia-trie-tree-work
 - https://blog.ethereum.org/2015/11/15/merkling-in-ethereum/

thepiwo:
@meivantodorov @HristianHristov this is stuck for 11 days now, can we PR and merge?

meivantodorov:
@thepiwo this is still work in progress, and it is not `stuck`. When it is ready, there will be PR.

thepiwo:
when do you think it will be ready, as #341 is depended on this and there have not been any commits in 11 days.

meivantodorov:
The work was focused on improving [this](https://github.com/exthereum/merkle_patricia_tree) library. I hope that after 2 days there will be PR.

PR #345, opened 2018-04-11
[GH-#253] Patricia merkle tree

meivantodorov:


mix.exs
```diff
@@ -53,7 +53,10 @@ defmodule EpochElixir.Mixfile do
       {:aewallet, github: "aeternity/elixir-wallet"},
       {:msgpax, "~> 2.1.1"},
       {:bip0173, "~> 0.1.2"},
-      {:erl_base58, "~> 0.0.1"}
+      {:erl_base58, "~> 0.0.1"},
+      {:merkle_patricia_tree,
+       git: "https://github.com/meivantodorov/merkle_patricia_tree.git",
+       branch: "meivantodorov/deleting-node"}
```
thepiwo: can you add this in master branch, then I will for it to aeternity

meivantodorov: It is still in progress, after it is finished You will fork it.

thepiwo: ok, ping me once its done

apps/aeutil/lib/patricia_merkle_tree.ex
```diff
@@ -0,0 +1,95 @@
...
+    end
+  end
+
+  @doc """
+  Retrieve value from trie and construct proof.
+  """
+  @spec lookup_with_proof(Trie.key(), Trie.t()) :: :none | {:ok, Trie.value(), Trie.t()}
+  def lookup_with_proof(key, trie) do
+    {value, proof} = Proof.construct_proof({trie, key, new(:proof)})
+    {:ok, value, proof}
```
cytadela8: It's `@spec`-ed this can return `:none`, but it will never return it. It should probably return `:none` for non existent key?

meivantodorov: Thanks, I will fix that.

apps/aeutil/lib/patricia_merkle_tree.ex
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+  @doc """
+  Retrieve value from trie and construct proof.
+  """
+  @spec lookup_with_proof(Trie.key(), Trie.t()) :: :none | {:ok, Trie.value(), Trie.t()}
+  def lookup_with_proof(key, trie) do
+    case Proof.construct_proof({trie, key, new(:proof)}) do
+      {nil, _proof} -> :none
+      {value, proof} -> {:ok, value, proof}
```
gspasov: You could name `value` as `val` and follow the previous `lookup()` function variable naming.

rdimitrovbg:
Would be really great if you place the tree itself as first argument in all of APIs' function. The advantages of this is using them in with a pipe ` |> ` operator. 
In an example if you need to update a multyple keys with new value

Example:
``` elixir
tree
|> Api.fun1(key1,val1)
|> Api.fun1(key2,val2)
|> Api.fun_n(k_n,v_n)
....
```

PR #348, opened 2018-04-11
[GH-#332] Fixed Credo issues and made Travis to check credo

### LINKED ISSUE #332: fix credo config and issues without ignoring rules

gspasov:
resolves #332


apps/aecore/lib/aecore/structures/chainstate.ex
```diff
@@ -82,13 +83,13 @@ defmodule Aecore.Structures.Chainstate do
...
         if valid_chainstate do
-          {valid_txs_list ++ [tx], updated_chainstate}
+          {[tx | valid_txs_list], updated_chainstate}
         else
           {valid_txs_list, chainstate_acc}
         end
       end)
 
-    valid_txs_list
+    Enum.reverse(valid_txs_list)
```
d-velev: Are you sure we need to keep the same order in this list as before? If not then just returning valid_txs_list is better

apps/aecore/lib/aecore/oracle/oracle.ex
```diff
@@ -231,20 +237,28 @@ defmodule Aecore.Oracle.Oracle do
...
-                                                                            response_height_included:
-                                                                              response_height_included
-                                                                          }},
-                                                                         acc ->
-      query_absolute_ttl = calculate_absolute_ttl(query.query_ttl, query_height_included)
+  def remove_expired_interaction_objects(
+        chain_state,
+        block_height
+      ) do
+    inter_objects = chain_state.oracles.interaction_objects
```
d-velev: rename it to interaction_objects

apps/aeutil/lib/scientific.ex
```diff
@@ -1,4 +1,9 @@
 defmodule Aeutil.Scientific do
+  @moduledoc """
+  Converts integer to scientific notation.
+  It's used for the storing the target in the block and recalculating the next targer
```
d-velev: should be "It's used for storing the target in the block and recalculating the next target"

apps/aecore/lib/aecore/structures/chainstate.ex
gspasov: Yes, it crashes if it's not in the correct order

PR #350, opened 2018-04-12
[GH-#274] naming system

### LINKED ISSUE #274: implement naming system according to spec
https://github.com/aeternity/protocol/blob/master/AENS.md

thepiwo:
~~fee adjustment is waiting for #331 to merge~~
~~binary length checks wait for #333 to merge~~

~~documentation will be added to Readme.md~~

resolves #274


thepiwo:
@gspasov @d-velev @0legIv can you review?

apps/aecore/lib/aecore/naming/name_util.ex
```diff
@@ -0,0 +1,92 @@
...
+    if allowed_registrar do
+      validate_name_length(name)
+    else
+      {:error, "name doesn't end with allowed registrar"}
+    end
+  end
+
+  @spec validate_name_length(String.t()) :: :ok | {:error, String.t()}
+  defp validate_name_length(name) do
+    case String.length(name) > 0 && String.length(name) < 253 do
```
gspasov: Shouldn't this max length name value be stored somewhere and not "hardcoded" here?

apps/aecore/lib/aecore/naming/name_util.ex
```diff
@@ -0,0 +1,92 @@
...
+  defp split_name(name), do: String.split(name, @split_name_symbol)
+
+  @spec validate_label_length([]) :: :ok
+  defp validate_label_length([]) do
+    :ok
+  end
+
+  @spec validate_label_length(list(String.t())) :: :ok | {:error, String.t()}
+  defp validate_label_length([label | remainder]) do
+    case String.length(label) > 0 && String.length(label) <= 63 do
```
gspasov: Again maybe have a variable for the max label length?

apps/aecore/lib/aecore/tx/pool/worker.ex
```diff
@@ -152,6 +157,21 @@ defmodule Aecore.Tx.Pool.Worker do
 
       %OracleExtendTx{} ->
         tx.data.fee >= OracleExtendTx.calculate_minimum_fee(tx.data.payload.ttl)
+
+      %NameClaimTx{} ->
+        NameClaimTx.is_minimum_fee_met?(tx)
```
gspasov: Can't we just make a default variable `_` here and catch all the Txs where we just call `is_minimum_fee_met?()`?

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -500,6 +505,27 @@ defmodule Aecore.Chain.Worker do
           OracleQueryTx ->
             [tx.data.sender, tx.data.payload.oracle_address]
 
+          OracleResponseTx ->
+            tx.data.sender
```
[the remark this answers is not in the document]
thepiwo: I prefer it being handled explicitly, so you don't forget about adjusting when adding a new transaction. @d-velev do you have an opinion about it?

apps/aecore/lib/aecore/account/account.ex
```diff
@@ -74,6 +83,151 @@ defmodule Aecore.Account.Account do
...
+          Wallet.pubkey(),
+          Wallet.privkey(),
+          String.t(),
+          binary(),
+          non_neg_integer(),
+          non_neg_integer()
+        ) :: {:ok, SignedTx.t()}
+  def pre_claim(sender, sender_priv_key, name, name_salt, fee, nonce) do
+    payload = %{commitment: Naming.create_commitment_hash(name, name_salt)}
+    spend_tx = DataTx.init(NamePreClaimTx, payload, sender, fee, nonce)
```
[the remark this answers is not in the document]
thepiwo: will look into it

apps/aecore/lib/aecore/naming/name_util.ex
```diff
@@ -0,0 +1,92 @@
...
+
+    case validate_normalized_name(normalized_name) do
+      :ok -> {:ok, normalized_name}
+      {:error, reason} -> {:error, reason}
+    end
+  end
+
+  @spec normalize_name(String.t()) :: String.t()
+  def normalize_name(name),
+    do:
```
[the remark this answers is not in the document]
thepiwo: Had it for debug reasons, will adjust.

apps/aecore/lib/aecore/naming/name_util.ex
```diff
@@ -0,0 +1,92 @@
...
+
+  @spec validate_normalized_name(String.t()) :: :ok | {:error, String.t()}
+  defp validate_normalized_name(name) do
+    allowed_registrar =
+      Enum.any?(@name_registrars, fn registrar -> String.ends_with?(name, registrar) end)
+
+    if allowed_registrar do
+      validate_name_length(name)
+    else
+      {:error, "name doesn't end with allowed registrar"}
```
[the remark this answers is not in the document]
thepiwo: will adjust, forgot about it

apps/aecore/lib/aecore/tx/pool/worker.ex
thepiwo: Prefer to have this explicit again, to not forget about it for new transactions. Whats your opinion about it?

apps/aecore/lib/aecore/naming/name_util.ex
```diff
@@ -0,0 +1,92 @@
...
+
+  @split_name_symbol "."
+
+  @name_registrars [@split_name_symbol <> "aet", @split_name_symbol <> "test"]
+
+  @spec normalized_namehash!(String.t()) :: binary()
+  def normalized_namehash!(name) do
+    case normalize_and_validate_name(name) do
+      {:ok, normalized_name} -> namehash(normalized_name)
+      {:error, error} -> throw(error)
```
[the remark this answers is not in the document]
thepiwo: This would lead to very deep nesting for correct error handling, as I tried implementing this. Maybe you can propose a solution for this, which is more elegant.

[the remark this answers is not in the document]
gspasov: Okay, here is my proposal. I see that you want to check the validity of the name before doing any of the NamingTxs, but sometimes you check it at creation time, and sometimes not. I propose to check it only in creation time with a `case` statement. To escape the repeated `case` statement you could make a generic naming_tx that will get the payload, check the naming and create a `data_tx` followed by a `spend_tx`. I know that you have 2 cases where in the payload the name has a different key. Can't you make it the same throughout the naming transaction payloads?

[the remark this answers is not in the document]
thepiwo: can you add a commit for this, otherwise I'd leave it like this and maybe adjust in the future?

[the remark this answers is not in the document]
gspasov: I'll add a commit with my suggestion.

apps/aecore/lib/aecore/naming/tx/name_claim_tx.ex
```diff
@@ -126,10 +126,10 @@ defmodule Aecore.Naming.Tx.NameClaimTx do
           tx_type_state
         ) :: :ok | {:error, String.t()}
   def preprocess_check(tx, sender, account_state, fee, _nonce, _block_height, naming_state) do
-    pre_claim_commitment = Naming.create_commitment_hash(tx.name, tx.name_salt)
+    {:ok, pre_claim_commitment} = Naming.create_commitment_hash(tx.name, tx.name_salt)
```
thepiwo: this will fail if `tx.name` is not valid, which is the first occurrence this will happen in a naming workflow, should be handled more nice (best in the `cond do`). This was the point were I stopped in my try to refactor this.

gspasov: Well, as far as I see it, if the `tx.name` is invalid, the `validate()` func will fail, so `preprocess_check()` won't even be called

apps/aecore/lib/aecore/oracle/tx/oracle_extend_tx.ex
```diff
@@ -11,6 +11,8 @@ defmodule Aecore.Oracle.Tx.OracleExtendTx do
   alias Aecore.Account.Account
   alias Aecore.Wallet.Worker, as: Wallet
 
+  require Logger
```
d-velev: Nothing is logged here, don't need this.

apps/aecore/lib/aecore/naming/tx/name_claim_tx.ex
```diff
@@ -0,0 +1,166 @@
...
+  end
+
+  @spec get_chain_state_name :: Naming.chain_state_name()
+  def get_chain_state_name, do: :naming
+
+  @doc """
+  Claims a name for one account after it was pre-claimed.
+  """
+  @spec process_chainstate(
+          SpendTx.t(),
```
d-velev: Should be NameClaimTx.t() right? I see that it's the same for the other naming transactions as well

apps/aecore/lib/aecore/naming/tx/name_claim_tx.ex
```diff
@@ -0,0 +1,166 @@
...
+
+    {updated_accounts_chainstate, updated_naming_chainstate}
+  end
+
+  @doc """
+  Checks whether all the data is valid according to the NameClaimTx requirements,
+  before the transaction is executed.
+  """
+  @spec preprocess_check(
+          SpendTx.t(),
```
d-velev: Same as above

apps/aecore/lib/aecore/naming/naming.ex
```diff
@@ -0,0 +1,161 @@
...
+  @spec init_empty :: state()
+  def init_empty, do: %{}
+
+  @spec create_commitment(
+          binary(),
+          Wallet.pubkey(),
+          non_neg_integer(),
+          non_neg_integer()
+        ) :: commitment()
+  def create_commitment(hash, owner, created, expires),
```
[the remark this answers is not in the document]
thepiwo: I think it is ok, as there are one expression

apps/aecore/lib/aecore/naming/tx/name_claim_tx.ex
thepiwo: didn't see this issue from

Issue #294, opened 2018-03-20
Implement State Channels


cytadela8:
parts:
- [x] multisignature transactions
- [x] channel transactions
- [x] channel state manager

PR #355, opened 2018-04-19
[GH-#228] CoinbaseTx, refactoring and multisig transactions

### LINKED ISSUE #228: extract coinbase transaction as own structure
this should include 3 fields, account (which is the miner and receiver of the amount), height of the block and amount

cytadela8:
- separate CoinbaseTx from DataTx
- DataTx now has `senders` list instead of one `sender` pubkey. Analogically SignedTx has `sigs` instead of `sig`.
- Moved parts of transactions serialization from Serialization module to structures modules. Serialization was adjusted so existing serialization isn't affected despite adding multisig txs.
- `process_chainstate!` and `preprocess_check` now has arity only 5. Instead of passing some elements of DataTx in random order we pass datatx.
- `AccountStateTree.get` returns empty account when none accounts is found. This simplifies handling in other places.
- new function `AccountStateTree.update` - applies given as argument function to given account. This simplifies a lot of transaction appllication on accounts.
- Account last_update update was moved to transactio_in!
- [NEW] Wallet.key_size_valid? was not boolean althoug name and spec suggested that.
- [NEW] mix format sometimes creates lengths longer then 100 chars although it targets 98 chars. Removed that check from credo.
- [NEW] Better Pool worker message when transaction doesn't validate
- fix various problems, refactored transactions handling code, added sanity checks

Notes:
- Serialization should probably be refactored more.
- There is no multisig tests, because currently it's impossible to create valid multisig transaction. The multisig code is included so other changes to transactions code won't cause (any more) conflicts with my Channels code.
- Tx sender key size test was removed, becouse that was moved to SignedTx.is_valid and there is no good way to check that as wrong sender size will also cause signature validation failure.

this closes #228


cytadela8:
I will also review my code tomorrow as changes associated with merging might have added some problems.

apps/aecore/lib/aecore/account/account.ex
```diff
@@ -70,68 +78,56 @@ defmodule Aecore.Account.Account do
...
+          ChainState.account()
+  def transaction_in!(account_state, block_height, amount) do
     new_balance = account_state.balance + amount
-    Map.put(account_state, :balance, new_balance)
-  end
-
-  @doc """
-  Deducts balance from a given address (public key)
-  """
-  @spec transaction_out(Account.t(), integer(), integer()) :: Account.t()
```
thepiwo: is this removed on purpose?

apps/aecore/lib/aecore/account/tx/spend_tx.ex
```diff
@@ -54,88 +57,87 @@ defmodule Aecore.Account.Tx.SpendTx do
...
+          SpendTx.t(),
+          DataTx.t()
+        ) :: {ChainState.accounts(), tx_type_state()}
+  def process_chainstate!(accounts, %{}, block_height, %SpendTx{} = tx, data_tx) do
+    sender = DataTx.sender(data_tx)
+
+    new_accounts =
+      accounts
+      |> AccountStateTree.update(sender, fn acc ->
+        Account.transaction_in!(acc, block_height, tx.amount * -1)
```
thepiwo: I see that this is why `transaction_out` is removed, but we should either keep `transaction_out` and have the `* -1` there, or change the name to something more generic like `apply_transaction`.

apps/aecore/lib/aecore/tx/data_tx.ex
```diff
@@ -38,80 +36,120 @@ defmodule Aecore.Tx.DataTx do
...
+  def nonce(%DataTx{nonce: nonce}) do
+    nonce
+  end
+
+  def payload(%DataTx{payload: payload}) do
+    payload
+  end
+
+  def sender(tx) do
+    List.last(senders(tx))
```
thepiwo: why should this be last, either rename or comment in code with explaination

apps/aecore/lib/aecore/tx/data_tx.ex
```diff
@@ -38,80 +36,120 @@ defmodule Aecore.Tx.DataTx do
...
-        Logger.error("Fee not enough")
+      !Enum.member?(valid_types(), type) ->
+        Logger.error("Invalid tx type=#{type}")
+        false
+
+      fee < 0 ->
+        Logger.error("Negative fee")
         false
 
-      !Wallet.key_size_valid?(sender) ->
```
thepiwo: the key-size should now be checked for both senders

thepiwo:
@cytadela8 I changed the name, as we don't PR #294 yet

apps/aecore/lib/aecore/account/account.ex
cytadela8: Yes, this is due to separating nonce update. I think this way is less confusing. We want to do only one nonce update per transaction always, but frequently we do many changes to account balance. (for example one fee deduction and one transfer to other account)

apps/aecore/lib/aecore/account/tx/spend_tx.ex
cytadela8: I will change the name.

apps/aecore/lib/aecore/tx/data_tx.ex
cytadela8: I will add more explanation and will consider changing the name from sender to "main_sender" or something like that.

apps/aecore/lib/aecore/tx/data_tx.ex
cytadela8: I check the sender size when I check the signature. I will consider changing it.

apps/aecore/lib/aecore/account/tx/spend_tx.ex
thepiwo: this solves my https://github.com/aeternity/elixir-node/pull/355#discussion_r182962819 as well

apps/aecore/lib/aecore/tx/data_tx.ex
thepiwo: is the last the main? Isn't every sender equal in multisig? one is maybe the initiator

apps/aecore/lib/aecore/tx/data_tx.ex
thepiwo: I think this would be the right place to have the check

cytadela8:
@thepiwo Great comments. As for the ordering of senders.

What I do currently: the last sender in the list is special and his nonce is updated. He also pays the fee in default fee implementation. The order of signatures has to match the order of senders. (SignedTx.sign_tx assures correct signature order) The order of senders can affect transaction code. For example channels have "initiator" and "responder". I serialize it to senders=[responder_pubkey, initiator_pubkey].

I considered adding some const function in every transaction that returns number of senders, but "who said" there have to fixed number of senders per transaction. This is a little bit less strict. I will think about this a little bit and maybe add validate_number_of_senders to every tx instead. What do you think?

As I understand epoch code: changes to our serialization will be enough to match their implementation fully. I removed separate nonces for every sender to match specs.

thepiwo:
Is their multisig implementation like this, seems weird to me, but ok, shouldn't this be in general, a "shared" account?

cytadela8:
I don't know how they implement multisig. Our API will be compatible. (Everything will be serialized to the same form and interpreted in the same way)

apps/aecore/lib/aecore/tx/data_tx.ex
cytadela8: Core team decided that there is one special sender whose nonce is updated. I decided to make the last one special. (Although I will probably change it to the first one for performance.)

cytadela8:
@thepiwo please, rereview

apps/aecore/lib/aecore/account/tx/coinbase_tx.ex
```diff
@@ -0,0 +1,140 @@
...
+
+      DataTx.fee(data_tx) != 0 ->
+        Logger.error("Fee has to be 0")
+        false
+
+      !Wallet.key_size_valid?(receiver) ->
+        Logger.error("Wrong receiver key size")
+        false
+
+      length(DataTx.senders(data_tx)) != 0 ->
```
thepiwo: Shouldn't this be `!empty?`

thepiwo: In other places you separate `senders = DataTx.senders(data_tx)`, better be consistent

cytadela8:
This is ready, but I can't pass `credo -a` and `mix format` at the same time.

`mix format` creates a 101 character line. If i split it `mix format` removes my endlines.

thepiwo:
@cytadela8 maybe work around using different variable/function names? Otherwise we would have to remove `mix format` from the CI

cytadela8:
I would rather consider removing line limit for credo, as that's handled by `mix format`

thepiwo:
Okay, do that

cytadela8:
I added the solution and one more comment info to the PR description

apps/aecore/lib/aecore/account/account.ex
```diff
@@ -70,68 +102,32 @@ defmodule Aecore.Account.Account do
...
 
   @doc """
-  Adds balance to a given address (public key)
+  Adds balance to a given Account state and updates last update block.
   """
-  @spec transaction_in(Account.t(), integer()) :: Account.t()
-  def transaction_in(account_state, amount) do
+  @spec apply_transfer!(ChainState.account(), non_neg_integer(), integer()) ::
+          ChainState.account()
+  def apply_transfer!(account_state, block_height, amount) do
```
gspasov: I don't like the throw here. And I don't even see a reason for it. It is used in `process_chainstate` to check if the sender has enough coins to pay for the fees and the amount itself, but we check this in `preprocess_check`. That's the reason of the `preprocess_check` anyways.

apps/aecore/lib/aecore/account/account.ex
```diff
@@ -70,68 +102,32 @@ defmodule Aecore.Account.Account do
...
-  Return the last_updated for a given key.
-  """
-  @spec last_updated(AccountStateTree.tree(), Wallet.pubkey()) :: non_neg_integer()
-  def last_updated(tree, key) do
-    get_account_state(tree, key).last_updated
-  end
+  @spec apply_nonce!(ChainState.account(), integer()) :: ChainState.account()
+  def apply_nonce!(%Account{nonce: current_nonce} = account_state, new_nonce) do
+    if current_nonce >= new_nonce do
+      throw({:error, "#{__MODULE__}: Invalid nonce"})
```
gspasov: I don't like this throw either. Why don't we just check this in `preprocess_check` or `validate` funcs? Even if there is some reason to check it again.. for whatever reason, I'm sure the `throw` could be resoved in another way.

apps/aecore/lib/aecore/account/tx/coinbase_tx.ex
```diff
@@ -0,0 +1,138 @@
...
+  @doc """
+  Checks transactions internal contents validity
+  """
+  @spec validate(CoinbaseTx.t(), DataTx.t()) :: :ok | {:error, String.t()}
+  def validate(%CoinbaseTx{amount: amount, receiver: receiver}, data_tx) do
+    senders = DataTx.senders(data_tx)
+
+    cond do
+      amount < 0 ->
+        {:error, "#{__MODULE__}: Value cannot be a negative number"}
```
gspasov: In the error message, maybe it should state `The amount` not `Value`.

apps/aecore/lib/aecore/chain/chainstate.ex
```diff
@@ -59,68 +57,18 @@ defmodule Aecore.Chain.Chainstate do
...
-
-    new_accounts_state =
-      AccountStateTree.put(chainstate.accounts, data.payload.receiver, updated_receiver_state)
-
-    {:ok, Map.put(chainstate, :accounts, new_accounts_state)}
-  end
+  def apply_transaction_on_state(chainstate, block_height, tx) do
+    case SignedTx.validate(tx) do
+      :ok ->
+        SignedTx.process_chainstate(chainstate, block_height, tx)
```
gspasov: Why do you think it's better to hide the called functions in some nested calls, instead of having them explicit here as before?

apps/aecore/lib/aecore/tx/signed_tx.ex
```diff
@@ -56,21 +75,42 @@ defmodule Aecore.Tx.SignedTx do
...
-  def sign_tx(%DataTx{} = tx, priv_key) when byte_size(priv_key) == 32 do
-    signature = Signing.sign(Serialization.pack_binary(tx), priv_key)
 
-    if byte_size(signature) <= get_sign_max_size() do
-      {:ok, %SignedTx{data: tx, signature: signature}}
+  @spec sign_tx(DataTx.t() | SignedTx.t(), binary(), binary()) ::
+          {:ok, SignedTx.t()} | {:error, binary()}
+  def sign_tx(%DataTx{} = tx, pub_key, priv_key) do
+    signatures =
+      for _ <- DataTx.senders(tx) do
```
gspasov: Why do you fill the signatures with `nil` values?

apps/aecore/lib/aecore/account/account.ex
cytadela8: This is a sanity check. Better to have an unhanded error in block processing then to accept an invalid block.

apps/aecore/lib/aecore/account/account.ex
cytadela8: This is also a sanity check only. It should never throw.

apps/aecore/lib/aecore/chain/chainstate.ex
cytadela8: Because more abstraction is better.

apps/aecore/lib/aecore/tx/signed_tx.ex
cytadela8: Because signatures can be added in any order and this makes ordering signatures easier.

apps/aecore/lib/aecore/tx/signed_tx.ex
```diff
@@ -126,6 +166,77 @@ defmodule Aecore.Tx.SignedTx do
...
+    end
+  end
+
+  @spec deserialize(map()) :: DataTx.t()
+  def deserialize(tx) do
+    signed_tx = Serialization.deserialize_value(tx)
+    data = DataTx.deserialize(signed_tx.data)
+
+    cond do
+      signed_tx.signature != nil ->
```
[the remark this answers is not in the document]
cytadela8: `signature` exists in serialized form of the tx. When deserializing we set signature**s** to `[signature]`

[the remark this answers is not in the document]
cytadela8: (See serialization code)

apps/aecore/lib/aecore/tx/signed_tx.ex
```diff
@@ -56,21 +75,42 @@ defmodule Aecore.Tx.SignedTx do
...
-      {:error, "Wrong signature size"}
+      {:error, "#{__MODULE__}: Not in senders"}
     end
   end
 
-  def sign_tx(%DataTx{} = _tx, priv_key) do
-    {:error, "#{__MODULE__}: Wrong key size: #{inspect(priv_key)}"}
-  end
-
   def sign_tx(tx, _priv_key) do
```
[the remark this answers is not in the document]
cytadela8: fixed

apps/aecore/lib/aecore/tx/signed_tx.ex
```diff
@@ -126,6 +166,77 @@ defmodule Aecore.Tx.SignedTx do
...
   def base58c_decode_signature(_) do
-    {:error, "Wrong data"}
+    {:error, "#{__MODULE__}: Wrong data"}
+  end
+
+  @spec serialize(SignedTx.t()) :: map()
+  def serialize(%SignedTx{} = tx) do
+    signatures_length = length(tx.signatures)
+
+    cond do
```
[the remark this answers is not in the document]
cytadela8: fixed

apps/aecore/lib/aecore/account/tx/coinbase_tx.ex
cytadela8: fixed

thepiwo:
@gspasov can you approve, so @cytadela8 can merge if he is done

PR #358, opened 2018-04-20
[GH-#356] Adeed target to difficulty calculation

### LINKED ISSUE #356: Target to Difficulty calculation
The Difficulty is calculated from the Target and is used in Sync.Difficulty is used to select the winning fork of new blocks: the difficulty of chain of blocks is the sum of the diffculty of each block. The Difficulty itself is: Difficulty = <Target of difficulty 1> / Target. Target of Difficulty 1 is equal to 0xffff000000000000000000000000000000000000000000000000000000000000 as integer.
Sources: https://github.com/aeternity/epoch/blob/master/apps/aecore/src/aec_sync.erl
               https://github.com/aeternity/epoch/blob/master/apps/aecore/src/aec_pow.erl

HristianHristov:
Resolves #356

### DIFF
--- a/apps/aeutil/lib/scientific.ex
+++ b/apps/aeutil/lib/scientific.ex
@@ -6,6 +6,8 @@ defmodule Aeutil.Scientific do
 
   use Bitwise
 
+  @highest_target_int 0xFFFF000000000000000000000000000000000000000000000000000000000000
+
   @spec scientific_to_integer(integer) :: integer()
   def scientific_to_integer(scientific) do
     {exp, significand} = break_scientific(scientific)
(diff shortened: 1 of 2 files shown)


apps/aeutil/lib/scientific.ex
```diff
@@ -6,6 +6,8 @@ defmodule Aeutil.Scientific do
 
   use Bitwise
 
+  @highest_target_int 0xFFFF000000000000000000000000000000000000000000000000000000000000
```
gspasov: Maybe get this value from a `get_highest_target_int()` func so we don't preload the value in memory?

thepiwo: Is this for performance optimization? The variable would be needed for every block validation.

HristianHristov:
@thepiwo This is just the maximum target represented in integer(not scientific). We wouldn't need it in the validation.

Issue #357, opened 2018-04-20
Adjust Oracle structures to match Erlang implementation


Artur64:
To continue the work on issue #335 , our Oracle structures must match the Erlang implementations.

Here are some examples of incompatibilities:
1) TTLs of registered Oracles chain-state -  Our structure - a map with ttl_type and ttl itself , their implementation - have only field expires: https://github.com/aeternity/epoch/blob/master/apps/aeoracle/src/aeo_oracles.erl#L48 ,
which, i suppose, is being calculated here: https://github.com/aeternity/epoch/blob/master/apps/aeoracle/src/aeo_oracles.erl#L83
2) Query/Response formats: Our structure - a map , their implementation - UTF8 encoded string. 
(more info at https://github.com/aeternity/epoch/blob/master/apps/aeoracle/src/aeo_oracles.erl#L39 )
3) Query/response fields in our oracles interaction objects state should have fully composed DataTx with all data inside, not just %Aecore.Oracle.Tx.OracleQueryTx{}/some string 

And so on...

Useful links:
1) Erlang Oracle implementations: https://github.com/aeternity/epoch/tree/master/apps/aeoracle
2) OracleTxs: https://github.com/aeternity/epoch/blob/master/apps/aeoracle/include/oracle_txs.hrl
3) Their serializations file which is a bit outdated but still have some types described: https://github.com/aeternity/protocol/blob/master/serializations.md#oracles
4) Oracle utils https://github.com/aeternity/epoch/blob/master/apps/aeoracle/src/aeo_utils.erl

Of course there are more of incompatibilities, so we need to inspect whole Erlang code, find and make them compatible.

thepiwo:
Did you notice other differences, if so could you specify them, because of "And so on..."

Artur64:
Query field in our oracles interaction objects state should have fully composed DataTx with all data inside, not just %Aecore.Oracle.Tx.OracleQueryTx{}

Artur64:
Same goes for response field

thepiwo:
field renaming and structure can maybe be solved in serialization (as with transaction and blocks)

thepiwo:
closed by https://github.com/aeternity/elixir-node/pull/367

PR #367, opened 2018-04-27
[GH-#357] Adjust Oracles structure to match Erlang implementation

d-velev:


Artur64:
Why query_format and response_format are still maps? In Epoch specs , both of these fields are binaries, "Utf8 encoded string". I cant encode maps with RLP,  but i can provide a workaround - convert these maps to jsons (Poison library return jsons as a string in Elixir) and encode them with RLP. The downside of this workaround might be in utf8 encoding, as i mentioned before they expect a "Utf8 encoded string", not a structure.

d-velev:
They haven't implemented query/response validation in epoch yet since they are waiting for the contract team to do the necessary preparations there (they are going to use types from the contract language) so there is no way to replicate their query/response handling

Artur64:
Ok, so, @thepiwo whats your opinion?

thepiwo:
If strings are "compatible" for now, lets use them, otherwise merge as is

cytadela8:
@thepiwo I merged, but for some reason I get " Invalid query or response format definition:" in Oracle registration now.

cytadela8:
Note: my transaction refactoring fixed an error that might have caused format to always be considered correct, although that doesn't justify how tests pass on master.

thepiwo:
@d-velev can you look into the issues, @cytadela8 mentioned?

thepiwo:
what is the state of this?

PR #372, opened 2018-05-09
[GH-#370] Print of a patricia merkle trie

### LINKED ISSUE #370: pretty print function for patricia merkle tree
@cytadela8 recommended to provide a way to print patricia merkle trees content nicely in the shell, for debugging and logging purposes

meivantodorov:
Providing API for printing a patricia merkle trie in the shell as struct or as key value pair

Closes #370 

Merge after #366


thepiwo:
how can I call this? `Aeutil.PatriciaMerkleTree.print_trie(Chain.top_block_chain_state.accounts)` doesn't work for me

meivantodorov:
> can you add PatriciaMerkleTree to the iex aliases?

Thanks, I will.

> how can I call this? Aeutil.PatriciaMerkleTree.print_trie(Chain.top_block_chain_state.accounts) doesn't work for me.

You need #366 to be merged first. Until then You can make simple trie and test it until then.Like:
```
 t = Aeutil.PatriciaMerkleTree.new(:account) |> Aeutil.PatriciaMerkleTree.enter("k1", "v1") |> Aeutil.PatriciaMerkleTree.enter("k2", "v2")
 Aeutil.PatriciaMerkleTree.print_trie t
```

thepiwo:
How will I do it for the accounts tree in chain?

thepiwo:
@meivantodorov How will I do it for the accounts tree in chain?

thepiwo:
@meivantodorov could you have another look at this?

cytadela8:
I did the supplied example and got following results:

```
iex(7)> PatriciaMerkleTree.print_trie t                                                                                                            
~~~Trie~~~
Node: ext (prefix: [3, 1, 3, 1, 3])
  Node: branch (value: "")
    [0] Node: <empty>
    [1] Node: leaf ([]="v1")
    [2] Node: leaf ([]="v2")
    [3] Node: <empty>
    [4] Node: <empty>
    [5] Node: <empty>
    [6] Node: <empty>
    [7] Node: <empty>
    [8] Node: <empty>
    [9] Node: <empty>
    [10] Node: <empty>
    [11] Node: <empty>
    [12] Node: <empty>
    [13] Node: <empty>
    [14] Node: <empty>
    [15] Node: <empty>
~~~/Trie/~~~
```

So this is a step in good direction, but it's still not what I intended when suggesting this feature to thepiwo.

The current problem is that the state off all accounts is stored in merkle tree and that's not easily readable. Sometimes it's useful to be able to print the states off all accounts. In future when we change to merkle trees for all parts of chainstate this problem will also apply to oracles, namings, channels, etc.

To be able to easily examine contents of a merkle tree I would suggest we should create a function that takes a merkle tree as stored in chainstate and prints all pairs: leaf path - value. Optionally we could also create a function that takes a merkle tree and returns all keys (leaf paths).

_note: of course currently only accounts are stored in a merkle tree, but in future other parts of chainstate will also be_

meivantodorov:
@cytadela8 When the all chainstates are changed from merkle trie to patricia merkle trie You can then print what You want. This is not merkle trie print functionality. There is an issue for getting all keys of a given patricia merkle trie.

thepiwo:
the task is supposed to "provide a way to print patricia merkle trees content nicely"

meivantodorov:
> the task is supposed to "provide a way to print patricia merkle trees content nicely"

"Nicely" keeping the structure of the trie, or printing the keys and the values, or different way of "nicely" ?

cytadela8:
I would consider nicely as printing keys and values pairs. That's what I meant when I proposed this task.

meivantodorov:
@cytadela8 
> I would consider nicely as printing keys and values pairs.

Thank You, I will do it so as it will be more helpful.

thepiwo:
Sorry for the misunderstanding, with nicely we meant easily human readable, so the content is understandable, not the structure

thepiwo:
@meivantodorov can you give an update on the status?

apps/aeutil/lib/patricia_merkle_tree.ex
```diff
@@ -86,12 +87,54 @@ defmodule Aeutil.PatriciaMerkleTree do
...
   def delete(trie, key), do: Trie.delete(trie, key)
+
+  @doc """
+  Providing pretty print of a given trie in the shell.
+  Depending on the atom it can print structure or key value pairs
+
+  ## Examples
+
+  If we want to print as pair
+      iex> Aeutil.PatriciaMerkleTree.new(:test_trie) |> Aeutil.PatriciaMerkleTree.enter("111", "val1") |> Aeutil.PatriciaMerkleTree.enter("112", "val2") |> Aeutil.PatriciaMerkleTree.print_trie(:as_pair)
```
thepiwo: I think one example to print one chain state tree would be sufficient

meivantodorov: > I think one example to print one chain state tree would be sufficient

After #366 is merged, I can provide such an example.

thepiwo:
will close this util unblocked and resolved

PR #399, opened 2018-05-22
[GH-#364] New peer connections using TCP & noise

### LINKED ISSUE #364: New peer connections using TCP & noise

d-velev:
**Changes in this PR:**

- Old peer and sync modules are removed, new peer modules are introduced. A peer connection is established through TCP via ranch (an erlang library that manages TCP connections in a pool, it handles connection accepting and starts a new peer process whenever a new peer connects).  

- As before, the Peers module is the module that keeps track of a node's peers and has the functions that are used to add a peer, list all peers etc.

- As far as supervision goes, the Supervisor module has 3 children - the Peers module, the PeerConnectionSupervisor which supervises the individual peer connection processes and a ranch connection listener process which is responsible for starting the PeerConnection processes whenever there's an incoming connection. 

- The PeerConnection module handles the connection between 2 peers, it contains functions for sending and handling incoming messages which may be blocks, transactions, headers etc. Peer messages are communicated via the core team's enoise library. The documentation for those can be found here - https://github.com/aeternity/protocol/blob/30652d59858afb0663dfea95f2023a3254da0a6f/sync/p2p_messages.md. 

closes #364


apps/aecore/lib/aecore/peers/peer_connection.ex
```diff
@@ -0,0 +1,528 @@
...
+  defp send_request_msg(msg, pid), do: GenServer.call(pid, {:send_request_msg, msg})
+
+  defp send_msg_no_response(msg, pid) when byte_size(msg) > @max_packet_size - 2 do
+    number_of_chunks = div(@fragment_size + byte_size(msg) - 1, @fragment_size)
+    send_chunks(pid, 1, number_of_chunks, msg)
+  end
+
+  defp send_msg_no_response(msg, pid), do: GenServer.call(pid, {:send_msg_no_response, msg})
+
+  defp send_chunks(pid, n, m, msg) when n == m do
```
gspasov: What does `n` and `m` stand for? Could they be more descriptive or add a comment?

d-velev: Peer messages are documented here https://github.com/aeternity/protocol/blob/30652d59858afb0663dfea95f2023a3254da0a6f/sync/p2p_messages.md - n is the current fragment index and m is the total fragments

d-velev:
@thepiwo About the multinode script, tried running it but I'm guessing the directories are messed up, it fails to start

thepiwo:
The docker-compose script does work on master, so it should work here as well:
```
docker-compose build
docker-compose up
```

I am not able to start two nodes with `PORT=4000` and `PORT=4001`.

docker-compose.yml
```diff
@@ -26,7 +26,7 @@ services:
         - elixir-node-4001-db:/elixir_node/apps/aecore/priv/rox_db_4001
       stdin_open: true
       tty: true
-      command: -e "pubkey = Client.get_info(\"elixir-node-4000:4000\") |> elem(1) |> Map.get(:public_key)" -e "decoded_pubkey = Account.base58c_decode(pubkey)" -e "peer = Map.put(%{}, :host, 'elixir-node-4000') |> Map.put(:port, 3015) |> Map.put(:pubkey, decoded_pubkey)" -e "Aecore.Peers.Worker.try_connect(peer)" -e "Aecore.Miner.Worker.resume()"
+      command: -e "pubkey = Aehttpclient.Client.get_info(\"elixir-node-4000:4000\") |> elem(1) |> Map.get(:public_key)" -e "decoded_pubkey = Aecore.Account.Account.base58c_decode(pubkey)" -e "peer = Map.put(%{}, :host, 'elixir-node-4000') |> Map.put(:port, 3015) |> Map.put(:pubkey, decoded_pubkey)" -e "Aecore.Peers.Worker.try_connect(peer)" -e "Aecore.Miner.Worker.resume()"
```
0legIv: Does it work? Because when I tested docker-compose, I found out that it doesn't memorize the variables if you pass them from one -e command to another, so you can't initialize for example pubkey variable, and after that to use its value in another -e command, tell me if I am wrong, because I am not sure about that.

d-velev: I couldn't run the docker since some dependency fails to compile, I haven't tested it so I need someone else to run it, I'll change it to one combined command and push, if you could test it it would be great.

dialyzer.ignore-warnings
```diff
@@ -311,4 +336,14 @@ apps/aecore/lib/aecore/tx/data_tx.ex:616: The call 'Elixir.Aecore.Tx.DataTx':ini
...
\ No newline at end of file
+<<<<<<< HEAD
```
thepiwo: this is not merged

README.md
```diff
@@ -126,14 +126,15 @@ The DataTx strucure hold:
...
 
 - To get all peers:
 
   `Peers.all_peers()`
 
-- To add peer:
+- Connect to a peer by specifying the address (host), port (SYNC_PORT) and peer pubkey (different from the keypair which is used for transaction signing):
 
-  `Peers.add_peer(uri)`
+  `Peers.try_connect(%{host: host, port: port, pubkey: pubkey})`
```
thepiwo: how do I find these from another peer? maybe we can introduce a `Peers.connection_info()`

d-velev: We can do that yes, not sure if it's really needed if peer discovery is done entirely automatically through the trusted nodes (right?)

d-velev:
Going to close this until RLP is done.

apps/aecore/lib/aecore/peers/peer_connection.ex
```diff
@@ -529,27 +530,148 @@ defmodule Aecore.Peers.PeerConnection do
...
+
+    list = [
+      @p2p_msg_version,
+      serialized_result,
+      :binary.encode_unsigned(type),
+      reason,
+      serialized_object
+    ]
+
+    ExRLP.encode(list)
```
gspasov: I think the variable list here and in all the other `rlp_encode()` functions is redundant. Just pass the list as the argument. In some of the functions it is already done in this way, why not do it for all of them?

apps/aecore/lib/aecore/peers/peer_connection.ex
```diff
@@ -529,27 +530,148 @@ defmodule Aecore.Peers.PeerConnection do
...
-        peers: peers,
-        port: port
-      }) do
+  # RLP for peer messages
+
+  defp rlp_encode(@p2p_response, %{result: result, type: type, object: object, reason: reason}) do
+    serialized_result = bool_bin(result)
+
+    reason =
+      case reason do
```
gspasov: If `reason` is a String variable, this could be interchanged with the following: 
```elixir
reason = reason || <<>>
```
This is going to work only if `reason`cannot be `false`.

apps/aecore/lib/aecore/account/tx/coinbase_tx.ex
```diff
@@ -0,0 +1,137 @@
+defmodule Aecore.Account.Tx.CoinbaseTx do
```
thepiwo: don't readd CoinbaseTx

apps/aecore/lib/aecore/keys/peer.ex
```diff
@@ -0,0 +1,117 @@
+defmodule Aecore.Keys.Peer do
```
thepiwo: can't we use a lot of functionality from the wallet here?

apps/aecore/lib/aecore/keys/peer.ex
```diff
@@ -0,0 +1,117 @@
...
+
+  def generate_keypair do
+    %{public: sign_pub, secret: sign_sec} = :enacl.sign_keypair()
+    pubkey = :enacl.crypto_sign_ed25519_public_to_curve25519(sign_pub)
+    privkey = :enacl.crypto_sign_ed25519_secret_to_curve25519(sign_sec)
+    pubkey <> privkey
+  end
+
+  def load_keypair do
+    case File.ls(keypair_dir()) do
```
thepiwo: the file handling code would make more sense in the Utils

apps/aecore/lib/aecore/peers/peer_connection.ex
thepiwo: can you name it more descriptive anyways

apps/aecore/lib/aecore/peers/peer_connection.ex
```diff
@@ -0,0 +1,775 @@
+defmodule Aecore.Peers.PeerConnection do
```
thepiwo: there are still hardcoded constants left, can you make them @ variables

thepiwo: this file is really long, is there a way to split that makes sense?

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -1,329 +1,161 @@
...
-      end
-
-    {:reply, is_synced, state}
+  def try_connect(peer_info) do
+    GenServer.cast(__MODULE__, {:try_connect, peer_info})
   end
 
-  def handle_call({:add_peer, uri}, _from, state) do
-    add_peer(uri, state)
+  def rlp_encode_peers(peers) do
```
thepiwo: the encode/decode functions would make more sense somewhere in a util

dialyzer.ignore-warnings
```diff
@@ -51,8 +51,10 @@ Function get_account_balance/1 has no local return
 Function get_account_txs/1 has no local return
 Function get/1 has no local return
 Function get/2 has no local return
+=======
```
thepiwo: there is some merging indicators left

apps/aecore/lib/aecore/keys/peer.ex
gspasov: Yes, we are going to, but I'll add it in the Sync PR #414

thepiwo: yes okay

PR #404, opened 2018-05-30
[GH-#396] Fix incorrect function specifications

### LINKED ISSUE #396: Fix incorrect function specifications

DanielaIvanova:
Resolves #396


apps/aecore/lib/aecore/account/tx/spend_tx.ex
```diff
@@ -115,10 +117,10 @@ defmodule Aecore.Account.Tx.SpendTx do
   before the transaction is executed.
   """
   @spec preprocess_check(
-          ChainState.account(),
+          Chainstate.accounts(),
           tx_type_state(),
           non_neg_integer(),
-          SpendTx.t(),
+          t(),
```
thepiwo: I think the full spec is nicer here

DanielaIvanova: For Elixir it is custom to write the types without the module name (If we are in the current module of course). @rdimitrovbg and I looked into some of the Elixir documentations and this is how they do it.

thepiwo: It is my personal opinion, If I just read some code I can understand way better what is meant, whats your opinion @d-velev @gspasov @cytadela8 ?

gspasov: If this is the way Elixir does it, it should be the way to go.

thepiwo: What is your personal opinion?

gspasov: To leave it like this.

thepiwo: @d-velev @meivantodorov @cytadela8 (@untra maybe?)

PR #409, opened 2018-06-01
[GH-#408] removed halt-exit-status for dialyzer in travis.yml

### LINKED ISSUE #408: Remove --halt-exit-status for dialyzer

0legIv:
resolves #408


d-velev:
We don't need the dialyzer.ignore-warnings file now as well right?

0legIv:
@d-velev I think we need it, because some of the errors are about unknown functions from another modules like this `:0: Unknown function 'Elixir.Aewallet.KeyPair':compress/1`. Dialyzer just can't see them, so they create some mess

thepiwo:
@0legIv we may need some of it, but just things that don't change the line numbers, so please clean or remove completely

thepiwo:

PR #410, opened 2018-06-04
[GH-#369] Represent naming state using PMT

### LINKED ISSUE #369: Represent naming state using patricia merkle tree

DanielaIvanova:
resolves #369


meivantodorov:
> I think it would be better to have the state transform for persistence, in the persistence module, than in the chain worker.

I think that persistence module should do exactly what the name says - `persist`.

thepiwo:
I assume the transformation is only done for persistence, so this has nothing to do with the chain either.

apps/aeutil/lib/serialization.ex
```diff
@@ -721,7 +721,7 @@ defmodule Aeutil.Serialization do
   def tag_to_type(24), do: OracleResponseTx
   def tag_to_type(25), do: OracleExtendTx
   def tag_to_type(30), do: Name
-  def tag_to_type(31), do: NameCommitmentTx
+  def tag_to_type(31), do: NameCommitment
```
thepiwo: is this changed on purpose?

DanielaIvanova: Yes, because was not working. Have to be like now

PR #411, opened 2018-06-05
[GH-#307] add multi node sync testcase

### LINKED ISSUE #307: add multi node sync testcase
fail test if nodes don't sync
maybe has to be done in python

0legIv:
resolves #307 
Modified docker-compose file, it runs 3 nodes. The first one adds spend tx and all of the oracle txs and mines them. The second one adds all of the naming txs and mines them.
Commands to run docker-compose:
`docker-compose build`
`docker-compose up`


thepiwo:
will close this, until it:
- actually fails tests if something is not working
- is somewhat extensible
- checks if stuff is synced across nodes

PR #414, opened 2018-06-07
[GH-#359] Building a new Sync implementation using the Noise protocol

### LINKED ISSUE #359: Rewrite the Sync module using the Noise protocol / framework
For starters we could use this erlang implementation of [Noice](https://github.com/aeternity/enoise)

More on how the Sync is implemented in Epoch:
- [Sync protocol](https://github.com/aeternity/protocol/blob/sync/SYNC.md)
- [Block sync](https://github.com/aeternity/epoch/wiki/Block-Synch)

If more info about Noice is needed use the following links:
- [Noice protocol](https://noiseprotocol.org/noise.html)

If interested one can read about the [Jobs](https://github.com/uwiger/jobs) library integrated as process scheduler.

gspasov:
resolves #359 

merge after PR #399


thepiwo:
is this independent from #399 ?

apps/aecore/lib/aecore/account/tx/coinbase_tx.ex
```diff
@@ -0,0 +1,137 @@
+defmodule Aecore.Account.Tx.CoinbaseTx do
```
thepiwo: coinbase should not be readded

apps/aecore/lib/aecore/keys/peer.ex
```diff
@@ -0,0 +1,108 @@
...
+          false ->
+            {:error, "#{__MODULE__} Derivation of Peer keypair failed"}
+        end
+
+      err ->
+        err
+    end
+  end
+
+  def generate_keypair do
```
thepiwo: this should be done, by resolving https://github.com/aeternity/elixir-wallet/issues/102

gspasov:
No, it includes PR #399 for testing purpose. This PR should be merged after #399 is ready.

apps/aecore/lib/aecore/keys/peer.ex
gspasov: Well it could be left like this until I resolve the issue in the elixir-wallet repo

thepiwo: okay

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -1,329 +1,164 @@
...
-
-  def init(initial_peers) do
-    {:ok, initial_peers}
+  def have_peer?(peer_pubkey) do
+    GenServer.call(__MODULE__, {:have_peer?, peer_pubkey})
   end
 
-  def handle_call(:is_chain_synced, _from, %{peers: peers} = state) do
-    local_latest_block_height = Chain.top_height()
+  def broadcast_block(%Block{} = block) do
```
cytadela8: Isn't this unused now, because we use Events?

gspasov:
The `hash_pool` has the following structure: 
```elixir
map :: %{peer: peer_id} | %{block: block}
[{{block_height, block_hash}, map}, ...]
```
It is used to store all the possible blocks that each peer has to offer, where it stores in the beginning only the height, hash, and peer_id and after we decide we need the block at that height with that hash we fetch the block and the peer_id map is replaced with the block map.
(Of course the peer_id is there to state from which peer we are getting the block)
 
```elixir
enqueue(:forward, %{status: :created, block: info}))
```

This code is handled in the `process_job` function, where we pattern match 
```elixir
{:forward, %{block: block}, peer_id} ->
```

The idea of the `status` was so that we know weather this block is fetched or forwarded, but I think we tried to make the fetch go through the `jobs` but we couldn't manage to do it. It's something to be deleted.

mix.exs
```diff
@@ -6,7 +6,7 @@ defmodule EpochElixir.Mixfile do
       app: :elixir_node,
       apps_path: "apps",
       version: "0.1.0",
-      elixir: "~> 1.6.4",
+      elixir: "~> 1.6",
```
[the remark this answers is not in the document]
gspasov: From auto-merging I suppose, will fix it.

mix.lock
```diff
@@ -1,9 +1,9 @@
 %{
-  "aewallet": {:git, "https://github.com/aeternity/elixir-wallet.git", "3f2f9df45f76e69ff20cdf2172af5ef596f794ed", []},
+  "aewallet": {:git, "https://github.com/aeternity/elixir-wallet.git", "3f2f9df45f76e69ff20cdf2172af5ef596f794ed", [ref: "3f2f9df"]},
   "base58check": {:git, "https://github.com/quanterall/base58check.git", "e012da96735852902df672d1543298cdf8f9664c", []},
   "bip0173": {:hex, :bip0173, "0.1.2", "b771e275692ed3353ecfecb38330693cbba90d2ac4a2e8fd1908a86ec7e6318f", [:mix], [], "hexpm"},
   "bunt": {:hex, :bunt, "0.2.0", "951c6e801e8b1d2cbe58ebbd3e616a869061ddadcc4863d0a2182541acae9a38", [:mix], [], "hexpm"},
-  "certifi": {:hex, :certifi, "2.3.1", "d0f424232390bf47d82da8478022301c561cf6445b5b5fb6a84d49a9e76d2639", [:rebar3], [{:parse_trans, "3.2.0", [hex: :parse_trans, repo: "hexpm", optional: false]}], "hexpm"},
+  "certifi": {:hex, :certifi, "2.0.0", "a0c0e475107135f76b8c1d5bc7efb33cd3815cb3cf3dea7aefdd174dabead064", [:rebar3], [], "hexpm"},
```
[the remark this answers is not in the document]
gspasov: Auto-merging again..

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -0,0 +1,607 @@
...
+      _other ->
+        Logger.debug(fn -> "Unknown job" end)
+    end
+
+    process_jobs(t)
+  end
+
+  @spec enqueue(atom(), map()) :: list()
+  defp enqueue(opts, msg) do
+    for pid <- Peers.all_pids() do
```
d-velev: Enum.each is better in this case

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -0,0 +1,607 @@
...
+      Jobs.enqueue(:sync_jobs, {opts, msg, pid})
+    end
+  end
+
+  # Send our block to the Remote Peer
+  defp do_forward_block(block, peer_id) do
```
d-velev: Don't think this and the function below are needed

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -0,0 +1,607 @@
...
+  defp do_fetch_block_ext(hash, peer_id) do
+    case PeerConnection.get_block(hash, peer_id) do
+      {:ok, block} ->
+        case BlockValidation.block_header_hash(block.header) === hash do
+          true ->
+            Logger.debug(fn ->
+              "#{__MODULE__}: Block #{inspect(block)} fetched from #{inspect(peer_id)}"
+            end)
+
+            {:ok, true, block}
```
d-velev: what does the true do?

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -0,0 +1,607 @@
...
+        err
+    end
+  end
+
+  # Try to fetch the pool of transactions
+  # from the Remote Peer we are connected to
+  defp do_fetch_mempool(peer_id) do
+    {:ok, pool} = PeerConnection.get_mempool(peer_id)
+    Logger.debug(fn -> "#{__MODULE__}: Mempool received from #{inspect(peer_id)}" end)
+    Enum.each(pool, fn {_hash, tx} -> Pool.add_transaction(tx) end)
```
d-velev: Heads up: get_mempool now returns a list of all transactions, not a map (should be changed to fn tx -> ...)

apps/aecore/lib/aecore/peers/worker.ex
```diff
@@ -155,6 +168,14 @@ defmodule Aecore.Peers.Worker do
     end
   end
 
+  def peer_id(peer_id) when is_binary(peer_id) do
```
d-velev: what is this used for?

apps/aecore/lib/aecore/peers/worker/supervisor.ex
```diff
@@ -1,15 +1,17 @@
 defmodule Aecore.Peers.Worker.Supervisor do
   @moduledoc """
   Supervises the Peers, PeerConnectionSupervisor and ranch acceptor
```
d-velev: "Supervises the Peers, PeerConnectionSupervisor, Sync, Jobs and ranch acceptor processes with a one_for_all strategy"

apps/aecore/lib/aecore/peers/worker.ex
gspasov: Removed

apps/aecore/lib/aecore/peers/sync.ex
gspasov: Look like it does nothing.

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -1,364 +1,607 @@
...
+  alias Aeutil.Scientific
+  alias Aecore.Tx.Pool.Worker, as: Pool
+  alias Aecore.Peers.Jobs
+  alias Aecore.Peers.Worker, as: Peers
+  alias Aecore.Peers.Events
 
   require Logger
 
-  @peers_target_count Application.get_env(:aecore, :peers)[:peers_target_count]
+  @typedoc "Structure of a peer to sync with"
```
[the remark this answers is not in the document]
gspasov: I'll leave it as is, because if I move it to the Peers module it will clash with the other Peer structure, for connecting the Peers. Maybe it could be changed in the future as the Sync will be improved.

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -1,364 +1,607 @@
...
   end
 
-  @spec remove_running_task(String.t()) :: :ok
-  def remove_running_task(peer_uri) do
-    GenServer.call(__MODULE__, {:remove_running_task, peer_uri})
+  @doc """
+  Fetches the remote Pool of transactions
+  """
+  @spec fetch_mempool(pid()) :: :ok | {:error, String.t()}
+  def fetch_mempool(peer_id) do
```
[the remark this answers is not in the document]
gspasov: This is so we could use the `:jobs` library.

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -1,364 +1,607 @@
...
-            Logger.error(fn -> message end)
-        end
-      end)
+  def handle_call({:start_sync, peer_id, remote_hash}, _from, state) do
+    :ok = Jobs.enqueue(:sync_jobs, {:start_sync, peer_id, remote_hash})
+    {:reply, :ok, state}
+  end
 
-      earliest_block = Enum.at(blocks, Enum.count(blocks) - 1)
+  def handle_call({:fetch_mempool, peer_id}, _from, state) do
```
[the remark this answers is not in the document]
gspasov: Don't remember, I'll see if it could be a `cast`

gspasov:
Encountered problem where the root_hash that is calculated from taking a block from another node, does not match. Will close the PR until this issue is resolved.

gspasov:
The fetching of the transactions will stay a `handle_call` until a process is made that checks if there is a job to be dequeued. This will be added in a separate issue.

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -141,7 +141,7 @@ defmodule Aecore.Chain.Worker do
     end
   end
 
-  @spec get_header(binary()) :: Block.t() | {:error, reason()}
+  @spec get_header(binary()) :: Header.t() | {:error, reason()}
```
thepiwo: can you rename this to get_header_by_hash maybe, to describe what it actually does?

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -506,7 +506,8 @@ defmodule Aecore.Chain.Worker do
 
     if Enum.empty?(blocks_map) do
       [block_hash] = Map.keys(state.blocks_data_map)
-      Persistence.add_block_by_hash(block_hash, state.blocks_data_map[block_hash])
+      %{block: block} = state.blocks_data_map[block_hash]
+      Persistence.add_block_by_hash(block_hash, block)
```
thepiwo: why does get... lead to add... ?

thepiwo: this does not return header right now?

gspasov: 1st Q: I'm not quite sure why, it was like this in master already;
2nd Q: The problem was that we added in the persistence the whole `%{block: block}` map instead of just adding the block itself;

thepiwo: 1. okay lets keep it for now
2. but you can return `.header` right?

gspasov: Yes, it's inside the `block` map data

thepiwo: So if it supposed to return the header, do so

thepiwo:
getting some error on connecting a 3rd peer 
```
iex(9)> [error] Process #PID<0.17892.0> raised an exception
** (UndefinedFunctionError) function Aecore.Chain.Worker.get_header/1 is undefined or private. Did you mean one of:

      * get_header_by_base58_hash/1
      * get_header_by_hash/1
      * get_header_by_height/1
      * get_headers_forward/2

    (aecore) Aecore.Chain.Worker.get_header(<<148, 117, 29, 46, 227, 36, 250, 75, 143, 60, 246, 13, 42, 9, 217, 211, 81, 204, 23, 241, 58, 90, 162, 216, 227, 154, 146, 144, 175, 0, 235, 8>>)
    (aecore) lib/aecore/peers/peer_connection.ex:460: Aecore.Peers.PeerConnection.handle_get_header_by_hash/2
```

thepiwo:
I fixed the issue. Nakamoto consensus still seems to be broken. An existing chain wouldn't get reorganized if there is a longer one available in the network. I proposed #451 as follow up task to fix it.

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -0,0 +1,606 @@
...
+      :tx_received -> enqueue(:forward, %{status: :received, tx: info})
+    end
+
+    Jobs.dequeue(:sync_jobs)
+    {:noreply, state}
+  end
+
+  def handle_info({:DOWN, _ref, :process, pid, reason}, %{sync_pool: sync_pool} = state) do
+    Logger.info("Worker stopped with reason: #{inspect(reason)}")
+    {:noreply, %{state | sync_pool: Enum.filter(sync_pool, fn peer -> peer.peer != pid end)}}
```
d-velev: It is said that peer is String.t() (in the type definition), what exactly is it (I don't understand peer.peer != pid)? List.delete should be better here besides that. Is the process that is being monitored a sync job?

gspasov: Good catch, the peer is a pid() actually. And List.delete is a good option too.

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -0,0 +1,606 @@
...
+    else
+      :ok
+    end
+
+    {:reply, new?, %{state | sync_pool: new_pool}}
+  end
+
+  def handle_call({:sync_in_progress, peer_pid}, _from, %{sync_pool: pool} = state) do
+    result =
+      case Enum.find(pool, false, fn peer -> Map.get(peer, :pid) == peer_pid end) do
```
cytadela8: Is there some important reason for `pool` to be a list instead of map (pid -> peer)?

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -0,0 +1,606 @@
...
+  def handle_call({:fetch_next, peer_pid, height_in, hash_in, result}, _from, state) do
+    hash_pool =
+      case result do
+        {:ok, block} ->
+          block_height = block.header.height
+          block_hash = BlockValidation.block_header_hash(block.header)
+
+          # If the hash of this block does not fit wanted hash, it is discarded
+          # (In case we ask for block with hash X and we get a block with hash Y)
+          List.keyreplace(
```
cytadela8: Also: Is there some reason for state.hash_pool to be a list instead of map ({block_height, block_hash} -> block | peer)?

apps/aecore/lib/aecore/peers/sync.ex
```diff
@@ -0,0 +1,606 @@
...
+  defp agree_on_height(_peer_pid, _r_header, _r_height, l_height, agreed_hash)
+       when l_height == 0 do
+    {0, agreed_hash}
+  end
+
+  defp agree_on_height(peer_pid, r_header, r_height, l_height, agreed_hash)
+       when r_height == l_height do
+    r_hash = BlockValidation.block_header_hash(r_header)
+
+    case Persistence.get_block_by_hash(r_hash) do
```
cytadela8: You should probably query `Aecore.Chain` for blocks. (It implements a cache)

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -408,17 +425,27 @@ defmodule Aecore.Chain.Worker do
...
           :top_height => new_block.header.height
         },
-        :block_info => %{new_block_hash => %{refs: new_refs}}
+        :block_info => %{new_block_hash => %{refs: new_refs}},
+        :total_diff => %{:total_difficulty => new_total_diff}
       })
 
       ## We send the block to others only if it extends the longest chain
-      Peers.broadcast_block(new_block)
+      if Enum.empty?(Peers.all_pids()) do
```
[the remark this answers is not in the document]
gspasov: I don't think so. Why do we need to add jobs to the queue if we are sure that they are not going to be successful? This way we are filtering unnecessary enqueues of jobs.

apps/aecore/lib/aecore/peers/peer_connection.ex
```diff
@@ -592,7 +593,7 @@ defmodule Aecore.Peers.PeerConnection do
       port,
       share,
       genesis_hash,
-      difficulty,
+      :erlang.term_to_binary(difficulty),
```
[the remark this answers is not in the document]
gspasov: RLP doesn't work with floating point, so we need to format the difficulty in some way for the RLP to work. We are not encoding the difficulty here, the RLP does the encoding, here we are just changing the format so RLP could do it's job. `(encode the data)`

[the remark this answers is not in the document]
thepiwo: we and epoch shouldn't use `term_to_binary` anywhere, how does epoch solve this?

apps/aecore/lib/aecore/peers/sync.ex
gspasov: Maybe not, it was made this way, because that's how `epoch` did it. But I don't find a reason why to change it for the sake of it right now. I feel that if we decide to change it, to do so in an upcoming issue.

apps/aecore/lib/aecore/peers/sync.ex
gspasov: Same as my previous comment. `epoch` did it like so, and I don't see a reason why to change it. If, for some reason, we decide to do so, will be issues a separate issue for this change.

thepiwo:
could you add module and data to all logging statements?

apps/aecore/lib/aecore/peers/peer_connection.ex
```diff
@@ -593,7 +593,7 @@ defmodule Aecore.Peers.PeerConnection do
       port,
       share,
       genesis_hash,
-      :erlang.term_to_binary(difficulty),
+      :erlang.float_to_binary(difficulty),
```
thepiwo: does epoch use `binary_to_float`?

gspasov: Yes, see [here](https://github.com/aeternity/epoch/blob/b724d07ebf5162586012ea54ed8e00eed9e74a88/apps/aecore/src/aec_peer_messages.erl#L174).

thepiwo: thank you!

apps/aecore/lib/aecore/peers/sync.ex
thepiwo: The goal is to copy specification, not translate erlang code to elixir. The question from @cytadela8 seems appropriate.

apps/aecore/lib/aecore/peers/sync.ex
thepiwo: The goal is to copy specification, not translate erlang code to elixir. The question from @cytadela8 seems appropriate.

Issue #413, opened 2018-06-07
Adjust cuckoo


meivantodorov:
Match epoch's cuckoo pow in order to:
- verify epoch's solutions; 
- generate correct solution using serialized header so an epoch node to be able to verify it;

thepiwo:
closed with #415

PR #415, opened 2018-06-07
[GH-#413] Cuckoo adjustments

meivantodorov:
#### PR description
- Successfully verified  [epoch](https://github.com/aeternity/epoch) solution from valid block in their testnet;
- They now can verify our solutions;
- Packed nonce and hash instead of passing nonce to the miner;
- Fixed improper representation of the solution;
- Fixed pow cuckoo eunit test;
- Serializing/Deserializing pow improvements;


apps/aeutil/lib/serialization.ex
```diff
@@ -554,7 +554,13 @@ defmodule Aeutil.Serialization do
     header_txs_hash_size = Application.get_env(:aecore, :bytes_size)[:txs_hash]
     header_root_hash_size = Application.get_env(:aecore, :bytes_size)[:root_hash]
     pow_evidence_size = Application.get_env(:aecore, :bytes_size)[:pow_total_size]
-    pow_to_binary = pow_to_binary(header.pow_evidence)
+
+    pow_to_binary =
+      if header.pow_evidence != :no_value do
+        pow_to_binary(Enum.reverse(header.pow_evidence))
```
Artur64: I think we should get rid of this Enum.reverse and just use Enum.sort over accumulator in pow_to_binary function.

apps/aeutil/lib/serialization.ex
```diff
@@ -606,7 +612,7 @@ defmodule Aeutil.Serialization do
     %Header{
       height: height,
       nonce: nonce,
-      pow_evidence: pow_evidence,
+      pow_evidence: Enum.reverse(pow_evidence),
```
Artur64: Same here

Artur64:
If you want , i can commit my changes.

PR #418, opened 2018-06-11
[GH-#416] Added ttl fields to all transactions, RLP adjustments

### LINKED ISSUE #416: Add TTL to all transactions
Add TTL fields to all tx's and make ttl's optional.(0 value of TTL means infinity).

Adjust RLP encodings.

Artur64:
Resolves #416 .


Artur64:
Nvm, i will add this check.

PR #420, opened 2018-06-13
[GH-#370] Pretty print of given PMT

### LINKED ISSUE #370: pretty print function for patricia merkle tree

meivantodorov:
### PR description
Providing pretty print in the shell of given PMT. The trie can be printed as struct or key-value pair.

closes #370


meivantodorov:
> how can I e.g. print the latest accounts state? Chain.get_chain_state_by_height(1).accounts |> PatriciaMerkleTree.print_trie(:as_pair) doesn't really show me any useful data.

@thepiwo you have to do it useful, it only prints only what is in the trie.

thepiwo:
I assume the account balances are in this tree and expect them to be printed

meivantodorov:
And they are, just they are rlp serialized.

thepiwo:
I understand, didn't think of it. For this task to have a useful result we would need them deserialized, otherwise it seems pretty useless

PR #426, opened 2018-06-19
[GH-#349] Represent oracles state as PMT

### LINKED ISSUE #349: represent oracles state as patricia merkle tree

DanielaIvanova:
resolves #349


apps/aecore/lib/aecore/chain/chainstate.ex
```diff
@@ -67,8 +62,9 @@ defmodule Aecore.Chain.Chainstate do
       %Chainstate{} = new_chainstate ->
         {:ok,
          new_chainstate
-         |> Oracle.remove_expired_oracles(block_height)
-         |> Oracle.remove_expired_interaction_objects(block_height)}
+         |> Oracle.remove_expired_oracles(block_height)}
+
+      #  |> Oracle.remove_expired_interaction_objects(block_height)}
```
thepiwo: is this not needed any more? if so, remove

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -654,25 +623,31 @@ defmodule Aecore.Chain.Worker do
...
+      {key = :naming, value}, acc_state ->
         Map.put(acc_state, key, value.root_hash)
 
-      # TODO
-      # This workaround was made until the Oracles were converted to PatriciaMerkleTree #GH-349
+      {key = :oracles, value}, acc_state ->
+        Map.put(acc_state, key, %{otree: value.otree.root_hash, ctree: value.ctree.root_hash})
+
       {key, value}, acc_state ->
         Map.put(acc_state, key, value)
```
thepiwo: this case is now an error, right, we should handle this

apps/aecore/lib/aecore/oracle/oracle.ex
```diff
@@ -41,12 +43,30 @@ defmodule Aecore.Oracle.Oracle do
         }
 
   @type t :: %{
-          registered_oracles: registered_oracles(),
-          interaction_objects: interaction_objects()
+          otree: registered_oracles(),
```
thepiwo: is this naming a must? otherwise choose more descriptive names

apps/aecore/lib/aecore/oracle/oracle.ex
```diff
@@ -41,12 +43,30 @@ defmodule Aecore.Oracle.Oracle do
...
+          ctree: interaction_objects()
         }
 
   @type ttl :: %{ttl: non_neg_integer(), type: :relative | :absolute}
 
+  ### getters ---------------------------------------------------------
```
thepiwo: we don't need these

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
```diff
@@ -0,0 +1,195 @@
+defmodule Aecore.Oracle.OracleStateTree do
```
thepiwo: this is split into sections using comments, consider moving the sections in seperate files

apps/aecore/lib/aecore/oracle/tx/oracle_extend_tx.ex
```diff
@@ -29,6 +30,19 @@ defmodule Aecore.Oracle.Tx.OracleExtendTx do
   @spec get_chain_state_name() :: :oracles
   def get_chain_state_name, do: :oracles
 
+  ### getter -----------------------------------------------------------
```
thepiwo: these are not needed

apps/aecore/lib/aecore/tx/transaction.ex
```diff
@@ -44,7 +43,7 @@ defmodule Aecore.Tx.Transaction do
   # Callbacks
 
   @doc "The name for state chain entry to be passed for processing"
-  @callback get_chain_state_name() :: Chainstate.chain_state_types() | :none
+  @callback get_chain_state_name() :: Chainstate.chain_state_types() | :none | nil
```
thepiwo: why can this be nil?

DanielaIvanova:
There is a problem with Patricia Merkle Tree library. This PR is blocked until @meivantodorov resolves the problem.

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
DanielaIvanova: I cleaned up the module and there is no need to split in different files

DanielaIvanova:
@thepiwo can you review

thepiwo:
will do

.iex.exs
```diff
@@ -26,3 +26,6 @@ alias Aeutil.Parser
 alias Aeutil.Scientific
 alias Aeutil.Serialization
 alias Aeutil.PatriciaMerkleTree
+
+alias MerklePatriciaTree.Trie.Inspector
```
thepiwo: why do you use this, instead of the PatriciaMerkleTree.pretty_print? I would prefer only aliasing our own components

apps/aecore/lib/aecore/chain/chainstate.ex
```diff
@@ -67,8 +62,7 @@ defmodule Aecore.Chain.Chainstate do
       %Chainstate{} = new_chainstate ->
         {:ok,
          new_chainstate
-         |> Oracle.remove_expired_oracles(block_height)
-         |> Oracle.remove_expired_interaction_objects(block_height)}
+         |> Oracle.remove_expired_oracles(block_height)}
```
thepiwo: if it is  only one command, the pipe is not needed.

are you sure `Oracle.remove_expired_interaction_objects` is not needed any more (maybe check with @d-velev)

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -680,25 +649,31 @@ defmodule Aecore.Chain.Worker do
       {key = :accounts, root_hash}, acc_state ->
         Map.put(acc_state, key, PatriciaMerkleTree.new(key, root_hash))
 
-      # TODO
-      # This workaround was made until the Oracles were converted to PatriciaMerkleTree #GH-349
+      {_key = :oracles, %{otree: o_root_hash, ctree: c_root_hash}}, acc_state ->
+        otree = %{:otree => PatriciaMerkleTree.new(:oracles, o_root_hash)}
+        ctree = %{:ctree => PatriciaMerkleTree.new(:oracles_cache, c_root_hash)}
```
thepiwo: can we rename `otree`, `o_root_hash` and `ctree`, `c_root_hash` to something nicer?

apps/aecore/lib/aecore/oracle/oracle.ex
```diff
@@ -294,21 +237,40 @@ defmodule Aecore.Oracle.Oracle do
...
   end
 
+  @spec get_registered_oracles :: map()
+  def get_registered_oracles do
+    otree = Chain.chain_state().oracles.otree
+    keys = PatriciaMerkleTree.all_keys(otree)
+
+    registered_oracles_key =
+      Enum.reduce(keys, [], fn x, acc ->
+        if byte_size(x) == 33 do
```
thepiwo: please don't use magic numbers, move it to a named constant https://en.wikipedia.org/wiki/Magic_number_(programming)

apps/aecore/lib/aecore/oracle/oracle.ex
```diff
@@ -294,21 +237,40 @@ defmodule Aecore.Oracle.Oracle do
     end
   end
 
+  @spec get_registered_oracles :: map()
+  def get_registered_oracles do
+    otree = Chain.chain_state().oracles.otree
+    keys = PatriciaMerkleTree.all_keys(otree)
+
+    registered_oracles_key =
+      Enum.reduce(keys, [], fn x, acc ->
```
thepiwo: give x a semantic name

apps/aecore/lib/aecore/oracle/oracle.ex
```diff
@@ -294,21 +237,40 @@ defmodule Aecore.Oracle.Oracle do
...
+        if byte_size(x) == 33 do
+          [x | acc]
+        else
+          acc
+        end
+      end)
+
+    Enum.reduce(registered_oracles_key, %{}, fn x, acc ->
```
thepiwo: give x a semantic name

apps/aecore/lib/aecore/oracle/oracle.ex
```diff
@@ -384,22 +347,28 @@ defmodule Aecore.Oracle.Oracle do
...
       ) do
     has_response =
       case Serialization.transform_item(has_response, :int) do
         1 -> true
         0 -> false
       end
 
+    new_response =
+      case response do
+        "undefined" -> String.to_atom(response)
```
thepiwo: why can this be "undefined" string?

apps/aecore/lib/aecore/oracle/tx/oracle_query_tx.ex
```diff
@@ -124,27 +124,21 @@ defmodule Aecore.Oracle.Tx.OracleQueryTx do
...
-        response: :undefined,
-        expires: Oracle.calculate_absolute_ttl(tx.query_ttl, block_height),
-        response_ttl: tx.response_ttl.ttl,
-        fee: tx.query_fee
-      })
-
-    updated_oracle_state = %{
-      oracle_state
-      | interaction_objects: updated_interaction_objects
+    io = %{
```
thepiwo: give `io` a more semantic name

apps/aecore/lib/aecore/tx/data_tx.ex
```diff
@@ -560,35 +564,35 @@ defmodule Aecore.Tx.DataTx do
...
          query_ttl_value,
-         encoded_response_ttl_type,
+         response_ttl_type,
          response_ttl_value,
          fee,
          ttl
        ]) do
-    query_ttl_type =
-      encoded_query_ttl_type
+    q_ttl_type =
```
thepiwo: the previous name was nicer

apps/aecore/lib/aecore/tx/data_tx.ex
```diff
@@ -560,35 +564,35 @@ defmodule Aecore.Tx.DataTx do
...
+      query_ttl_type
       |> Serialization.transform_item(:int)
       |> Serialization.decode_ttl_type()
 
-    response_ttl_type =
-      encoded_response_ttl_type
+    r_ttl_type =
```
thepiwo: the previous name was nicer

apps/aecore/lib/aecore/tx/data_tx.ex
```diff
@@ -560,35 +564,35 @@ defmodule Aecore.Tx.DataTx do
...
+      response_ttl_type
       |> Serialization.transform_item(:int)
       |> Serialization.decode_ttl_type()
 
-    query_data = decode_format(encoded_query_data)
+    q_data = decode_format(query_data)
```
thepiwo: the previous name was nicer

apps/aecore/lib/aecore/tx/data_tx.ex
```diff
@@ -605,27 +609,27 @@ defmodule Aecore.Tx.DataTx do
...
          query_fee,
-         encoded_ttl_type,
+         ttl_type,
          ttl_value,
          fee,
          ttl
        ]) do
-    ttl_type =
-      encoded_ttl_type
+    ttl_t =
```
thepiwo: the previous name was nicer

apps/aecore/lib/aecore/tx/data_tx.ex
```diff
@@ -605,27 +609,27 @@ defmodule Aecore.Tx.DataTx do
...
+      ttl_type
       |> Serialization.transform_item(:int)
       |> Serialization.decode_ttl_type()
 
-    query_format = decode_format(encoded_query_format)
+    q_format = decode_format(query_format)
```
thepiwo: the previous name was nicer

apps/aecore/lib/aecore/tx/data_tx.ex
```diff
@@ -605,27 +609,27 @@ defmodule Aecore.Tx.DataTx do
...
 
-    response_format = decode_format(encoded_response_format)
+    r_format = decode_format(response_format)
```
thepiwo: the previous name was nicer

apps/aehttpclient/test/aehttpclient_test.exs
```diff
@@ -16,7 +16,8 @@ defmodule AehttpclientTest do
     base58_encoded_top_block_hash = Header.base58c_encode(Chain.top_block_hash())
     Pool.get_and_empty_pool()
     add_txs_to_pool()
-    assert {:ok, _} = Client.get_info("localhost:4000")
+
+    # assert {:ok, _} = Client.get_info("localhost:4000") -> Commented until the problem with phoenix dependencies is resolved
```
thepiwo: will you open a issue to resolve this?

.iex.exs
DanielaIvanova: @meivantodorov 
Can you check this

apps/aecore/lib/aecore/chain/chainstate.ex
DanielaIvanova: Function Oracle.remove_expired_oracles is removing expired oracles and interaction objects

apps/aecore/lib/aecore/chain/worker.ex
DanielaIvanova: What about `oracle_tree: oracle_root_hash` and `cache_tree: cache_root_hash` ?

thepiwo: perfect, maybe even `oracle_cache_...`

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -680,25 +649,40 @@ defmodule Aecore.Chain.Worker do
...
 
-      # TODO
-      # This workaround was made until the Oracles were converted to PatriciaMerkleTree #GH-349
+      {key = :oracles, value}, acc_state ->
+        Map.put(acc_state, key, %{
+          oracle_tree: value.oracle_tree.root_hash,
+          oracle_cache_tree: value.oracle_cache_tree.root_hash
+        })
+
       {key, value}, acc_state ->
```
thepiwo: @DanielaIvanova this case is not needed anymore, as every chainstate is now handled explicitly, right?

DanielaIvanova: I will remove it

apps/aecore/lib/aecore/oracle/oracle.ex
DanielaIvanova: Because we can't encode atoms with RLP

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
```diff
@@ -0,0 +1,185 @@
...
+
+  defp enter(tree, key, value) do
+    PatriciaMerkleTree.enter(tree, key, value)
+  end
+
+  defp delete(tree, key) do
+    PatriciaMerkleTree.delete(tree, key)
+  end
+
+  defp lookup?(tree, key, where) do
```
d-velev: lookup. where and which_tree sound weird as names

rdimitrovbg: @d-velev 
Yes, I'm agree about the incorrectness of the function name ```lookup```, it should be ```exists```. After all it checks whether the obj under the key exists. We should correct this.

Since the existence of more than one Trie, the function is intended to work more generic way. To give the ability to ease the distinction from ```WHERE``` you want to make the selection and the function's name ```which_tree``` speaks for itself.

thepiwo:
@DanielaIvanova can you describe the issues you told us in the standup here again?

DanielaIvanova:
@thepiwo 
Yesterday we discussed with @rdimitrovbg  about the implementation how oracles and queries are deleted from oracle tree. We have to implement a functionality that  refunds money to query sender. If some oracle is expired and the query doesn't have response from the oracle we have to refund the fee to the query

thepiwo:
@DanielaIvanova @rdimitrovbg is the problem with refunding now resolved?

DanielaIvanova:
@thepiwo 
Yes, we think that everything is working properly. We can test more detailed(with several nodes) when the Sync is merged into master

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
```diff
@@ -0,0 +1,212 @@
...
+  def get_query(tree, key) do
+    get(tree.oracle_tree, key)
+  end
+
+  @spec exists_query?(oracles_state(), binary()) :: boolean()
+  def exists_query?(tree, key) do
+    exists?(tree, key, :oracle_query)
+  end
+
+  defp initialize_deletion({oracles_state, _accounts_state} = trees, expires) do
```
thepiwo: This could get a better name, why do we need a separate function from prune anyways?

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
```diff
@@ -0,0 +1,212 @@
...
+    oracles_state.oracle_cache_tree
+    |> PatriciaMerkleTree.all_keys()
+    |> Enum.reduce(trees, fn cache_key_encoded, new_trees_state ->
+      cache_key_encoded
+      |> cache_key_decode()
+      |> filter_expired(expires, cache_key_encoded, new_trees_state)
+    end)
+  end
+
+  defp filter_expired({expires, data}, expires, cache_key_encoded, trees) do
```
thepiwo: What's the difference to prune?

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
```diff
@@ -0,0 +1,212 @@
...
+        :oracle_cache_tree,
+        delete(updated_oracles_state.oracle_cache_tree, cache_key_encoded)
+      ),
+      updated_accounts_state
+    }
+  end
+
+  defp filter_expired(_, _, _, trees), do: trees
+
+  defp delete_expired({:oracle, oracle_id}, {oracles_state, accounts_state}) do
```
thepiwo: What's the difference to prune?

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
thepiwo: Could we move Oracle logic out of this file and have it only handle the actual state tree changes

thepiwo: In naming we have the `apply_block_height_on_state!` to apply logic not connected to one specific transaction in the general Naming module

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
rdimitrovbg: @thepiwo 
Prune is API function and is implicitly - called function.
Initialize_deletion prepares the keys from the given tree and calls function filter_expired. In filter_expired function we filter and delete expired oracles and queries therefore we have 2 separate logics. That's why our code is segmented.

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
rdimitrovbg: @thepiwo 
Same as https://github.com/aeternity/elixir-node/pull/426#discussion_r198509268

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
rdimitrovbg: @thepiwo 
Same as https://github.com/aeternity/elixir-node/pull/426#discussion_r198509268

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
DanielaIvanova: Could you be more specific about the logic which should be moved out and where?

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
thepiwo: okay, but as commented as well it would be good to split this logic from actual tree handling

DanielaIvanova:
@thepiwo, @d-velev , @meivantodorov 
Could you review, please

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
```diff
@@ -4,7 +4,7 @@ defmodule Aecore.Oracle.OracleStateTree do
   """
   alias Aeutil.PatriciaMerkleTree
   alias Aeutil.Serialization
-  alias Aecore.Oracle.Tx.OracleQueryTx
+  alias Aecore.Oracle.Tx.OracleQueryTx, as: Query
```
thepiwo: make the alias as `OracleQueryTx`, to be more clear what it references

apps/aecore/lib/aecore/oracle/tx/oracle_query_tx.ex
```diff
@@ -187,6 +187,16 @@ defmodule Aecore.Oracle.Tx.OracleQueryTx do
     end
   end
 
+  def refund_sender(query, accounts_state) do
```
thepiwo: this is not logic of the query tx itself, but more of the oracle system overall

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
```diff
@@ -0,0 +1,193 @@
...
+    %{
+      oracle_tree: PatriciaMerkleTree.new(:oracles),
+      oracle_cache_tree: PatriciaMerkleTree.new(:oracles_cache)
+    }
+  end
+
+  @spec prune(Chainstate.t(), non_neg_integer()) :: Chainstate.t()
+  def prune(chainstate, block_height) do
+    {new_oracles_state, new_accounts_state} =
+      initialize_deletion({chainstate.oracles, chainstate.accounts}, block_height - 1)
```
thepiwo: this seems like oracle logic, not tree logic as well

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
```diff
@@ -0,0 +1,193 @@
...
+    get(tree.oracle_tree, key)
+  end
+
+  @spec exists_query?(oracles_state(), binary()) :: boolean()
+  def exists_query?(tree, key) do
+    exists?(tree, key, :oracle_query)
+  end
+
+  defp initialize_deletion({oracles_state, _accounts_state} = trees, expires) do
+    oracles_state.oracle_cache_tree
```
thepiwo: this seems like oracle logic, not tree logic as well

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
```diff
@@ -0,0 +1,193 @@
...
+        :oracle_cache_tree,
+        delete(updated_oracles_state.oracle_cache_tree, cache_key_encoded)
+      ),
+      updated_accounts_state
+    }
+  end
+
+  defp filter_expired(_, _, _, trees), do: trees
+
+  defp delete_expired({:oracle, oracle_id}, {oracles_state, accounts_state}) do
```
thepiwo: this deletes by id if I am not mistaking, so `delete_expired` doesn't represent the function well

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
```diff
@@ -0,0 +1,193 @@
...
+    {
+      Map.put(oracles_state, :oracle_tree, delete(oracles_state.oracle_tree, oracle_id)),
+      accounts_state
+    }
+  end
+
+  defp delete_expired({:query, oracle_id, query_id}, {oracles_state, accounts_state}) do
+    tree_id = oracle_id <> query_id
+    query = get_query(oracles_state, tree_id)
+    new_accounts_state = Query.refund_sender(query, accounts_state)
```
thepiwo: as `refund_sender` is called here this seems like oracle logic, not tree logic as well

thepiwo:
would it be possible to use the same separation as in naming, with the `apply_block_height_on_state!` interface?

DanielaIvanova:
We discussed with @meivantodorov  about all these functions. We understand that this is oracle logic but it is also oracle state logic. We think that all these functions have to be in `OracleStateTree` module because `Oracle` should not handle the work with trees.

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
DanielaIvanova: Refer to: https://github.com/aeternity/elixir-node/pull/426#issuecomment-401778877

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
DanielaIvanova: Refer to: https://github.com/aeternity/elixir-node/pull/426#issuecomment-401778877

thepiwo:
Yes, you are right, not the whole function makes sense in the Oracle module. But the logic parts of it should be extracted to be in there.

If you wish we can keep it like this and I will open a PR myself to adjust it. Logic and Datasources should not be mixed, no matter if it is the tree in memory store or a database. So we can abstract each without affecting the other.

rdimitrovbg:
@thepiwo 

Well, then, at this stage, can we believe that the PR is ready to be merger with upstream master?

I like the idea for opening a new issue about abstracting logic and datasources.

thepiwo:
Okay, great, let's wait for #414 to be merged and schedule this afterwards

thepiwo:
@DanielaIvanova @rdimitrovbg will you need more testing with this, now that sync is merged, or is this finished and we create follow-up tasks, in case there are problems?

PR #427, opened 2018-06-20
[GH-#417] Multi node test framework

### LINKED ISSUE #417: Multi node test framework
Framework to simplify multi nodes testing. It must have such functionalities: 

- [ ] Add new node
- [ ] Add all types of txs to every node
- [ ] Run sync test

0legIv:
resolves #417 
Multi node test framework has such functionalities:

1. The state that keeps the important information about each node.
`state = %{`
            `process_port: process_port,  #  Process port to send commands to the node`
            `path: tmp_path,`
            `port: port,  # Server port`
            `sync_port: sync_port,  #  Sync port`
            `top_block: nil,  #  Top block structure (updated every time the function get_node_top_block is called)`
            `peers: %{},  #  Map with information about peerrs (updated every time the function get_all_peers is called)`
           ` registered_oracles: %{},   #  Map with information about registered oracles (updated every time the function registered_oracles(node_name) is called)`
            `oracle_interaction_objects: %{}`
`}`

2. Adding any number of nodes: they are copied to the temporary folder and with the port that is different from the port of the calling node.
**Function:** `Aeutil.MultiNodeTestFramework.new_node(node_name, port, sync_port)`

3. Send commands to the nodes using Port module: the command that is sent will be executed in a specified node iex shell with the respond. For now the respond is shown using IO.inspect. **But this will be changed, I just want some suggestions how to make it better.**

**Function:** `Aeutil.MultiNodeTestFramework.send_command(node_name, "Aecore.Chain.Worker.top_block"`)

4. There are already written all types of txs with the parameters taken from the tests and all frequently used functions like mine_sync_block_to_chain. So calling these functions you just need to specify the node name. 
**Examples:** 
`Aeutil.MultiNodeTestFramework.spend_tx(node_name)`
`Aeutil.MultiNodeTestFramework.mine_sync_block(node_name)`

5. The nodes can be synced and after that compared by their top_block. The top_block is written to the state every time the function get_node_top_block or compare_nodes is called.
**Example:**
`Aeutil.MultiNodeTestFramework.sync_two_nodes "node1", "node2"`
`Aeutil.MultiNodeTestFramework.compare_nodes "node1", "node2"`

6. The functionality to delete node releasing it's port and deleting the temporary folder.
**Function:** `Aeutil.MultiNodeTestFramework.delete_node(node_name)`


thepiwo:
@0legIv will you merge master and resolve ci issues?

apps/aecore/lib/aecore/multi_node_test_framework/worker.ex
```diff
@@ -0,0 +1,619 @@
...
+      "oracle = Chain.oracle_interaction_objects() |> Map.keys() |> Enum.at(0)"
+    )
+
+    send_command(node_name, "Oracle.respond(oracle, %{\"currency\" => \"BGN\"}, 5)")
+  end
+
+  # spend_tx
+
+  @spec spend_tx(String.t()) :: :ok | :unknown_node
+  def spend_tx(node_name1) do
```
gspasov: You don't need to add `1` to the name.

apps/aecore/lib/aecore/multi_node_test_framework/worker.ex
```diff
@@ -0,0 +1,619 @@
...
+
+      # deleting the file result.json
+      File.rm(path)
+      {:reply, :ok, new_state}
+    else
+      {:error, reason} -> {:reply, reason, state}
+    end
+  end
+
+  def handle_call({:update_registered_oracles_state, node_name}, _, state) do
```
gspasov: Because this handle_call is quite similar to `:update_oracle_interaction_objects_state`, what you could do is have a common function with the `with statement` and to put the different part in an anonymous function and pass it to the common function and execute it inside.

You could make this change in couple of places, I don't think I need to point you all of them.

Example:
```elixir

def handle_call() do
  function = fn() -> 
        Enum.reduce(oracles_decode32, %{}, fn {k, val}, _ ->
          atom_keys_map = for {nested_k, v} <- val, into: %{}, do: {String.to_atom(nested_k), v}
          new_map = put_in(oracles_decode32[k], atom_keys_map)
          new_map = put_in(new_map[k].oracle_address, Base.decode32!(new_map[k].oracle_address))
          put_in(new_map[k].sender_address, Base.decode32!(new_map[k].sender_address))
        end)  
   end
   common_func(function)
end

def common_func(function) do
    with {:ok, data} <- File.read(path),
         {:ok, decoded_data} <- Poison.decode(data) do
      oracles_decode32 = for {k, v} <- decoded_data, into: %{}, do: {Base.decode32!(k), v}
       function.(oracle_decode32)

         # deleting the file result.json
       File.rm(path)
       {:reply, :ok, new_state}
     else
       {:error, reason} -> {:reply, reason, state}
     end
   end
end  
```
This is a rough example, but I think you got my point.

apps/aecore/lib/aecore/multi_node_test_framework/worker.ex
```diff
@@ -0,0 +1,611 @@
...
+    get_node_top_block(node_name1)
+    get_node_top_block(node_name2)
+    GenServer.call(__MODULE__, {:compare_nodes_by_top_block, node_name1, node_name2})
+  end
+
+  @doc """
+    Updates registered oracles state for 2 nodes and compares them.
+    If registered oracles are equal - they are synced
+  """
+  def compare_nodes_by_registered_oracles(node_name1, node_name2) do
```
cytadela8: I think we should probably do a compare for other parts of chainstate. Also maybe one function to compare whole chainstate would be enough. Also I'm not sure if this function is needed at all. I think if top block (or even top block hash) matches then it's guaranteed for chainstates to be equal as chainstate hash is part of block (of course this assumes #453 is fixed).

apps/aecore/lib/aecore/multi_node_test_framework/worker.ex
```diff
@@ -0,0 +1,611 @@
...
+
+  # oracles
+  @spec oracle_interaction_objects(String.t()) :: :ok | :unknown_node
+  def oracle_interaction_objects(node_name) do
+    send_command(node_name, "int_object = Chain.oracle_interaction_objects()")
+
+    # converting the keys which are binary to string
+    send_command(
+      node_name,
+      "oracle_int_obj_for_encoding = for {k,v} <- int_object, into: %{}, do: {Base.encode32(k), v}"
```
cytadela8: Wouldn't it be actually easier to use `Serialization.rlp_encode(oracle, :oracle_query)` for serialization? (This will also be more resistant to changes in nodes) Or is there some reason that we can't serialize using existing serialization?

cytadela8: Or maybe we should use `:erlang.term_to_binary` as this will guarantee everything is always encoded and decoded properly (and we don't need any compatibility in tests with non elixir nodes)?

apps/aecore/lib/aecore/multi_node_test_framework/worker.ex
```diff
@@ -0,0 +1,611 @@
...
+
+    # encoding the map to the json
+    send_command(node_name, "{:ok, json} = Poison.encode(oracles_encoded)")
+
+    # writing it to the file
+    send_command(node_name, "path = System.cwd() <> \"/result.json\"")
+    send_command(node_name, "File.write(path, json)")
+
+    # updating the state of the oracle interaction objects after the one second
+    :timer.sleep(1000)
```
cytadela8: Sleep is something I really don't like in integration tests as: it makes them slow, they can still fail (and this may be hard to reproduce) because of too short sleep. 

But we should probably leave it as is, because sleep is very hard to get rid of in integration tests.

apps/aecore/lib/aecore/multi_node_test_framework/worker.ex
```diff
@@ -0,0 +1,611 @@
...
+
+  def handle_call({:new_node, node_name, port, sync_port}, _, state) do
+    cond do
+      Map.has_key?(state, node_name) ->
+        {:reply, :already_exists, state}
+
+      busy_port?(port) || busy_port?(sync_port) ->
+        {:reply, :busy_port, state}
+
+      true ->
```
cytadela8: This is good enough, but I think it's probably possible to use some Elixir features to just start multiple nodes with different environment variables (without copying all source code and compilation data).

apps/aecore/lib/aecore/multi_node_test_framework/worker.ex
0legIv: I use sleep here and in other places because the logic of receiving the result from each node is encoding the result and writing it to the json file and after that reading from it, and it needs some time to do this. So if there is a way to change this logic we can remove these timers.

meivantodorov:
I think that You can avoid sleep if You handle properly the responses in the handle_info. From what I see `Port` requires a pid where to send back the response. In fact You are using already handle_info to log errors.

I also see that `Port` provides monitoring, so You can use that as well if something goes wrong and the os terminate the process. This is what I'd try to do.

0legIv:
@thepiwo Yes, I have some errors in the tests now, I'm solving them

thepiwo:
I'll close this until the issues are resolved, reopen when ready for review

Issue #385, opened 2018-05-15
EVM integrated tests


nedelcho-delchev-tues:
- [x] implement json to AEVM internal values from EVM tests
 - [x] implement connection between EVM tests and AEVM
 - [ ] fix AEVM so that most of the EVM test pass

nedelcho-delchev-tues:
From 665 tests, 42 failures and most of them are from 3 missing instructions which we are implementing.
I'd say by the end of next week.

thepiwo:
@nedelcho-delchev-tues I suppose end of next week?

Issue #277, opened 2018-03-14
Solidity bytecode to opcodes


nedelcho-delchev-tues:
Convert a solidity smart contract bytecode to EVM opcodes

PR #433, opened 2018-06-24
[GH-#277] Initial AEVM implementation with EVM functionalities

### LINKED ISSUE #275: AEVM implementation

nedelcho-delchev-tues:
This is an initial implementation of the AEVM (Aeternity Virtual Machine) that contains all of the functionalities of the EVM (Ethereum Virtual Vachine). 

To run the VM you can use the following command: 
```
Aevm.loop(State.init_vm( %{ :code => State.bytecode_to_bin("0x60013b"),:address => 0,:caller => 0,:data => <<0::256, 42::256>>,:gas => 100_000,:gasPrice => 1,:origin => 0, :value => 0},%{:currentCoinbase => 0, :currentDifficulty => 0, :currentGasLimit => 10000, :currentNumber => 0, :currentTimestamp => 0},%{}, 0, %{:execute_calls => true}))
```

Where :code is the bytecode in hex that is going to be execute instruction by instruction

You can find each instruction(OP code) either from the file `op_codes.ex`,  [solidity opcodes section](http://solidity.readthedocs.io/en/v0.4.24/assembly.html) or from the tests.

Currently 629 tests pass from the official [EVM tests](http://ethereum-tests.readthedocs.io/en/latest/test_types/vm_tests.html)

closes #275, #277, #287 , #385


nedelcho-delchev-tues:
Performance test are disabled right now because each one takes ~10 min to execute if you want to run them
1. uncomment them from test/aevm_tests.exs
2. run `mix test test/aevm_test.exs --trace`

apps/aevm/lib/aevm.ex
```diff
@@ -0,0 +1,1430 @@
+defmodule Aevm do
+  use Bitwise
+
+  require OpCodes
+  require GasCodes
+  require OpCodesUtil
+  require AevmConst
+
+  def loop(state) do
```
thepiwo: add spec to a public function

apps/aevm/lib/aevm.ex
```diff
@@ -0,0 +1,1430 @@
...
+  defp exec(OpCodes._JUMP(), state) do
+    {position, state} = pop(state)
+    jumpdests = State.jumpdests(state)
+
+    if Enum.member?(jumpdests, position) do
+      jumpdest_cost = Gas.op_gas_cost(OpCodes._JUMPDEST())
+      state1 = Gas.update_gas(jumpdest_cost, state)
+      State.set_cp(position, state1)
+    else
+      throw({:error, "invalid_jump_dest, #{position}", state})
```
thepiwo: add the module identifier, for all occurrences in all files

apps/aevm/lib/aevm.ex
```diff
@@ -0,0 +1,1430 @@
...
+
+  defp call(state, op_code) do
+    if State.calldepth(state) < 1024 do
+      call1(state, op_code)
+    else
+      {0, state}
+    end
+  end
+
+  defp call1(state, op_code) do
```
thepiwo: can you give it a semantic correct name

apps/aevm/lib/aevm.ex
```diff
@@ -0,0 +1,1430 @@
...
+    {gas, state1} = pop(state)
```
thepiwo: find better names or representations for `stateN`

apps/aevm/lib/aevm.ex
```diff
@@ -0,0 +1,1430 @@
...
+      else
+        for _ <- 1..extend_to, into: <<>>, do: <<sign_bit::1>>
+      end
+
+    <<val::integer-unsigned-256>> = <<pad::bits, sign_bit::1, trunc_val::bits>>
+    val
+  end
+
+  defp call(state, op_code) do
+    if State.calldepth(state) < 1024 do
```
thepiwo: use named constant instead of magic numbers

apps/aevm/lib/aevm.ex
thepiwo: for all occurrences in all files (incl `wordN`)

apps/aevm/lib/aevm.ex
```diff
@@ -0,0 +1,1430 @@
+defmodule Aevm do
```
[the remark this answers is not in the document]
thepiwo: can you add one line names/descriptions to each function/patternmatch

apps/aevm/lib/gas.ex
```diff
@@ -0,0 +1,148 @@
...
+
+      words2 ->
+        first = round(GasCodes._GMEMORY() * words1 + Float.floor(words1 * words1 / 512))
+        second = round(GasCodes._GMEMORY() * words2 + Float.floor(words2 * words2 / 512))
+        first - second
+    end
+  end
+
+  def dynamic_gas_cost("CALL", state) do
+    # TODO: account creation?
```
thepiwo: todo?

apps/aevm/lib/aevm_const.ex
```diff
@@ -0,0 +1,7 @@
+defmodule AevmConst do
+  require Bitwise
+
+  defmacro mask256 do quote do: Bitwise.bsl(1, 256) - 1 end
```
thepiwo: is 256 word size? so name it (for all magic number occurrences in the PR)

d-velev:
Any reason for defining constants as macros? Maybe using module attributes will improve performance

cheezus1:
Module attributes cannot be accessed outside of the module and cannot be utilised to be used in pattern matching outside of the module, while macros can. As far as perfomance is concerned, I couldn't find any resource stating that macros can cause performance loss.

apps/aevm/lib/aevm_util.ex
```diff
@@ -0,0 +1,374 @@
...
+
+  @spec smod(integer(), integer()) :: non_neg_integer()
+  def smod(value1, value2) do
+    <<svalue1::integer-signed-256>> = <<value1::integer-unsigned-256>>
+    <<svalue2::integer-signed-256>> = <<value2::integer-unsigned-256>>
+    result = rem(rem(svalue1, svalue2 + svalue2), svalue2)
+    result &&& AevmConst.mask256()
+  end
+
+  def pow(op1, op2) when is_integer(op1) and is_integer(op2) and op2 >= 0, do: pow(1, op1, op2)
```
gspasov: You could make a specific Module for such long guards and combine them into one. Check Defining custom guard expressions [here](https://hexdocs.pm/elixir/master/guards.html).

apps/aevm/lib/aevm_util.ex
```diff
@@ -0,0 +1,374 @@
...
+          value
+      end
+
+    state1 = State.set_pc(old_pc + bytes, state)
+
+    {value, state1}
+  end
+
+  def create_account(_value, _area, state) do
+    # TODO
```
gspasov: Is this a forgotten `TODO`?

apps/aevm/lib/gas.ex
```diff
@@ -0,0 +1,156 @@
...
+        else
+          gas_one_64_substracted
+        end
+      else
+        gas
+      end
+  end
+
+  def dynamic_gas_cost("DELEGATECALL", state) do
+    # TODO
```
gspasov: Forgotten `TODO`?

apps/aevm/lib/state.ex
```diff
@@ -0,0 +1,285 @@
...
+    current_number = current_number(state)
+
+    cond do
+      nth_block >= current_number ->
+        0
+
+      _a == 256 ->
+        0
+
+      # h == 0 -> 0
```
gspasov: Should this be removed ?

apps/aevm/lib/aevm_util.ex
nedelcho-delchev-tues: No, it will be implemented in an iteration to come

apps/aevm/lib/gas.ex
nedelcho-delchev-tues: No, it will be implemented in an iteration to come

apps/aevm/lib/state.ex
nedelcho-delchev-tues: This is an actual case but will be implemented in some other iteration

thepiwo:
whats the progress here?

apps/aevm/lib/aevm.ex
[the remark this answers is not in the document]
cytadela8: I don't know it this isn't maybe considered resolved, but I wouldn't split. This file contains only one function so you don't have to be able to easily read whole file, because you know exactly whats in it without reading it. You can also easily find every function. It's also probably more efficient to do this this way.

apps/aevm/lib/aevm_util.ex
cytadela8: I agree a `non_neg_integer` guard would be nice.

apps/aevm/lib/gas.ex
```diff
@@ -0,0 +1,157 @@
...
+
+  @spec update_gas(integer(), map()) :: map() | {:error, String.t(), map()}
+  def update_gas(gas_cost, state) do
+    curr_gas = State.gas(state)
+
+    if curr_gas >= gas_cost do
+      gas_after = curr_gas - gas_cost
+      State.set_gas(gas_after, state)
+    else
+      throw({:error, "out_of_gas", state})
```
cytadela8: I think we generally prefer to not to use `throw` in our code, but I understand that it's very difficult to avoid it.

apps/aevm/lib/state.ex
```diff
@@ -0,0 +1,289 @@
...
+  defp export_env(state) do
+    %{
+      :currentCoinbase => State.current_coinbase(state),
+      :currentDifficulty => State.current_difficulty(state),
+      :currentGasLimit => State.current_gas_limit(state),
+      :currentNumber => State.current_number(state),
+      :currentTimestamp => State.current_timestamp(state)
+    }
+  end
+  end
```
cytadela8: Last `end` shouldn't be indented.

apps/aevm/lib/gas.ex
nedelcho-delchev-tues: Yeah it is kinda difficult, we use the thrown state/message in the EVM tests because some of them fail on purpose and we have to catch the error as to not be considered an actual test fail

cheezus1:
@cytadela8 @thepiwo 
Can someone of you suggest which parts of the code we should add docs to?

thepiwo:
can you add a how to use description to the readme file?

cytadela8:
I think that most function should have at least a short comment about it's function and purpose. Only if some function is very obvious or very is some special reason the comment can be omitted. Generally if something is not obvious it should be commented. When in doubt I just write a short comment about a function purpose and what it does. I made a short summery about your comments in different files. In some files you included nice comments. In others there is only a file description comment. If you don't have time to comment everything, try to comment at least the main functions in different files or maybe the less obvious ones. 

Summery of comments in files:
- `apps/aevm/lib/aevm.ex` - contains comments
- `apps/aevm/lib/aevm_util.ex` - DOESN'T contain comments. (I found only one sentence at beginning of file)
- `apps/aevm/lib/gas.ex` - DOESN'T contain comments
- `apps/aevm/lib/gas_codes.ex` - contains comments
- `apps/aevm/lib/memory.ex` - DOESN'T contain comments
- `apps/aevm/lib/op_codes.ex` - contains some comments
- `apps/aevm/lib/stack.ex` - contains some comments
- `apps/aevm/lib/state.ex` - DOESN'T contain comments

PR #448, opened 2018-06-27
[GH-#421] Persisting chainstates by given block hash

### LINKED ISSUE #421: Adjust function Chain.get_chain_state_by_height
Function ```Chain.get_chain_state_by_height```  is not working properly and should be adjusted.
How to reproduce this problem: 
1. Start the node and mine some blocks
2. Call ```Chain.get_chain_state_by_height(2)``` At this moment everything will work as intended
3.  Exit the node
4. Start the node again and call ```Chain.get_chain_state_by_height(2)``` again. You will get the following error: ```{:error, "Elixir.Aecore.Chain.Worker: Chainstate was delated"}```

meivantodorov:
In order to get chainstates by given block height we have to persist not only the latest PMT root hash but all previous as well. 

Closes #421


thepiwo:
What is this feature needed for, I the future we would have to remove it for performance reasons anyways. State in Bitcoin is about 3GB in Ethereum over 50GB. We don't want to store this amount of data for every block.

meivantodorov:
> What is this feature needed for, I the future we would have to remove it for performance reasons anyways. State in Bitcoin is about 3GB in Ethereum over 50GB. We don't want to store this amount of data for every block.

There is no other way to make the `get_chainstate_by_height` to work. If we do not need to get previous chainstates, we can just remove it.

thepiwo:
I did ask what we need it for.

meivantodorov:
> I did ask what we need it for.

@thepiwo there is bug in the `get_chainstate_by_height`. This issue is about that. The solution that I provided resolves the bug. If You think that we do not need it we can close this issue and make new one to remove `get_chainstate_by_height` from the chain worker.

thepiwo:
Is this function used for something?
We could find a way to recompute older chainstates for reorganization purposes that is more efficient.

meivantodorov:
> Is this function used for something?
> We could find a way to recompute older chainstates for reorganization purposes that is more efficient.

@cytadela8 wrote initially this function, maybe he can give more info. I discovered that it is not working and @DanielaIvanova confirmed that. I'm okey if we do not need it at all.

thepiwo:
okay, lets keep this fix for now and eventually revise later

cytadela8:
Generally I think we should persist last "n" chainstates for some small n. Like about 10-50. That's approximately what Ethereum clients are doing currently AFAIK. Such chainstates are useful in case there is a fork. So we don't have to regenerate current chainstate form genesis. I would merge this for now and add some pruning maybe later to address thepiwo valid concern.

As for my `get_chainstate_by_height` function. It's not used by any code, but it's useful in debugging when one wants to explore old chainstates. (I would say it's part of our node interface.)

apps/aecore/lib/aecore/persistence/worker.ex
```diff
@@ -105,10 +105,10 @@ defmodule Aecore.Persistence.Worker do
     GenServer.call(__MODULE__, {:add_account_chain_state, {account, data}})
   end
 
-  @spec get_all_chainstates() ::
+  @spec get_all_chainstates(binary()) ::
           {:ok, chain_state :: map()} | :not_found | {:error, reason :: term()}
-  def get_all_chainstates do
-    GenServer.call(__MODULE__, :get_all_chainstates)
+  def get_all_chainstates(block_hash) do
```
cytadela8: Maybe rename this to `get_chainstate` while we are at it.

PR #475, opened 2018-07-11
[GH-#470] Check added for expired Oracles

### LINKED ISSUE #470: Adjust oracle extend preprocess check for expired Oracle
Ensure no fee is taken during Oracle Extend Tx in case of expired Oracle

rdimitrovbg:
Resolves #470

### DIFF
--- a/apps/aecore/test/aecore_oracle_test.exs
+++ b/apps/aecore/test/aecore_oracle_test.exs
@@ -21,7 +21,7 @@ defmodule AecoreOracleTest do
     end)
   end
 
-  @tag timeout: 120_000
+  @tag timeout: 160_000
   @tag :oracle_test
   test "register and query an oracle, check response, check if invalid transactions are filtered out" do
     Pool.get_and_empty_pool()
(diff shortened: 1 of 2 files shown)


DanielaIvanova:
ok, we will

d-velev:
Did you come across some case where the oracle expires but it doesn't get removed from state to add this check?

apps/aecore/test/aecore_oracle_test.exs
```diff
@@ -115,6 +115,36 @@ defmodule AecoreOracleTest do
...
+    Miner.mine_sync_block_to_chain()
+    Miner.mine_sync_block_to_chain()
+    oracle_tree = Chain.chain_state().oracles.oracle_tree
+    oracle_key = oracle_tree |> PatriciaMerkleTree.all_keys() |> List.first()
+
+    oracle = OracleStateTree.get_oracle(Chain.chain_state().oracles, oracle_key)
+    expires_oracle = 12
+    assert oracle.expires == expires_oracle
+
+    Miner.mine_sync_block_to_chain()
```
thepiwo: would it be possible to use smaller numbers, so we don't have to do so much `Miner.mine_sync_block_to_chain()`

DanielaIvanova:
@d-velev 
We added a check in oracle extended tx because before that we could make extended tx for oracle which is expired(and actually is not being extended) but was taking the fee from the node. With this check we are avoiding this problem and the test shows that the expired oracle not being extended.

d-velev:
If transactions get applied to state first and then expired objects are pruned, it should be possible to extend the oracle at the block of expiration

DanielaIvanova:
If the oracle expires at block `height_6`,  if I got it right,  I will be able  to extend the oracle at the same `block_height` (6th). Is it correct?
Because at the current state of code  you are not able to extend the oracle at the "expiry" block height .

d-velev:
Yes, that's what's supposed to happen. Why isn't it possible? Transactions get applied first, oracles are pruned second.

DanielaIvanova:
I will fix it

DanielaIvanova:
We are closing this issue because of misunderstanding of extended oracle transactions logic. The problem was fixed in #474.

PR #478, opened 2018-07-16
[GH-#474] Improved oracle cache tree pruning functionality

### LINKED ISSUE #474: Improve oracle cache tree pruning functionality

DanielaIvanova:
resolves #474


apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
```diff
@@ -189,12 +191,49 @@ defmodule Aecore.Oracle.OracleStateTree do
...
+
+    if oracle.expires > exp do
+      delete(oracles_state.oracle_cache_tree, expired_cache_key)
+    else
+      oracles_state.oracle_cache_tree
+    end
+  end
+
+  defp remove_expired_cache_key({exp, {:query, oracle_id, id}}, expired_cache_key, oracles_state) do
+    query_id = oracle_id <> id
```
gspasov: This is the only row added to the previous function and the other code is copied here. Maybe the repeated code could be extracted to another function, but this should not be leaved as is.

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
```diff
@@ -77,23 +77,22 @@ defmodule Aecore.Oracle.OracleStateTree do
   end
 
   defp filter_expired({expires, data}, expires, cache_key_encoded, trees) do
```
d-velev: seems like "expires" is unused here?

rdimitrovbg: @d-velev 
We use it at level function pattern matching

PR #479, opened 2018-07-17
[GH-#461] Fixed problem with getting account balance 

### LINKED ISSUE #461: account balance we never matching :none
Related to #460 while we are getting account balance we never match [:none](https://github.com/aeternity/elixir-node/blob/master/apps/aehttpserver/lib/aehttpserver/web/controllers/balance_controller.ex#L13) case

DanielaIvanova:
resolves #461


cytadela8:
Looks good. From what I understand: We remove a case that was unreachable? (because tree returns empty account if account doesn't exist)

PR #484, opened 2018-07-18
[GH-#453] Root chainstate hash

### LINKED ISSUE #453: Root chainstate hash
Root hash is defined currently as follows:
```
def calculate_root_hash(chainstate) do
  AccountStateTree.root_hash(chainstate.accounts)
end
```

Root hash should be calculated from all chainstates. Probably we should check how Epoch does that and do the same for compatibility.

rdimitrovbg:
resolves #453

### DIFF
--- a/apps/aecore/lib/aecore/chain/chainstate.ex
+++ b/apps/aecore/lib/aecore/chain/chainstate.ex
@@ -20,9 +20,12 @@ defmodule Aecore.Chain.Chainstate do
   @protocol_version_field_size 64
   @protocol_version 17
 
-  @doc """ This is the canonical root hash of an empty Patricia merkle tree """
-  @canonical_root_hash <<69, 176, 207, 194, 32, 206, 236, 91, 124, 28, 98, 196, 212, 25, 61, 56, 228,
-                    235, 164, 142, 136, 21, 114, 156, 231, 95, 156, 10, 176, 228, 193, 192>>
+  @doc """
+  This is the canonical root hash of an empty Patricia merkle tree
+  """
+  @canonical_root_hash <<69, 176, 207, 194, 32, 206, 236, 91, 124, 28, 98, 196, 212, 25, 61, 56,
+                         228, 235, 164, 142, 136, 21, 114, 156, 231, 95, 156, 10, 176, 228, 193,
+                         192>>
 
   @state_hash_bytes 32
 
--- a/apps/aecore/lib/aecore/chain/chainstate.ex
+++ b/apps/aecore/lib/aecore/chain/chainstate.ex
@@ -20,9 +20,9 @@ defmodule Aecore.Chain.Chainstate do
   @protocol_version_field_size 64
   @protocol_version 17
 
-  @canonical_root_hash <<69, 176, 207, 194, 32, 206, 236, 91, 124, 28, 98, 196, 212, 25, 61, 56,
-                         228, 235, 164, 142, 136, 21, 114, 156, 231, 95, 156, 10, 176, 228, 193,
-                         192>>
+  @doc """ This is the canonical root hash of an empty Patricia merkle tree """
+  @canonical_root_hash <<69, 176, 207, 194, 32, 206, 236, 91, 124, 28, 98, 196, 212, 25, 61, 56, 228,
+                    235, 164, 142, 136, 21, 114, 156, 231, 95, 156, 10, 176, 228, 193, 192>>
 
   @state_hash_bytes 32
 
--- a/apps/aecore/lib/aecore/chain/chainstate.ex
+++ b/apps/aecore/lib/aecore/chain/chainstate.ex
@@ -20,6 +20,10 @@ defmodule Aecore.Chain.Chainstate do
   @protocol_version_field_size 64
   @protocol_version 17
 
+  @canonical_root_hash <<69, 176, 207, 194, 32, 206, 236, 91, 124, 28, 98, 196, 212, 25, 61, 56,
+                         228, 235, 164, 142, 136, 21, 114, 156, 231, 95, 156, 10, 176, 228, 193,
+                         192>>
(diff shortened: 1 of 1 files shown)


apps/aecore/lib/aecore/chain/chainstate.ex
```diff
@@ -104,7 +110,24 @@ defmodule Aecore.Chain.Chainstate do
...
+    end)
+    |> Hash.hash_blake2b()
+  end
+
+  defp pad_empty(root_hash_binary)
+       when is_binary(root_hash_binary) and byte_size(root_hash_binary) === @state_hash_bytes do
+    root_hash_binary
+  end
+
+  defp pad_empty(_) do
```
cytadela8: When do we return zeros? (What's the case?)
I think we should adjust the checks so we crash when the provided hash has different length, not return zeros.

rdimitrovbg: You may be right @cytadela8 . I share a similar view with you. But that's the way they do it in the `epoch` and we just follow them.
https://github.com/aeternity/epoch/blob/master/apps/aecore/src/aec_trees.erl#L327-L328

rdimitrovbg: @cytadela8 
What I found is the following
There is a difference in the implementation of patricia merkle trie (PMT) library for erlang and for elixir

`epoch`
When creating a PMT, the library returns an `empty binary  = <<>>` for root_hash.
Then after, outside the PTM library, a check is made to determine whether the root_hash is empty or not.
And if it is `empty = <<>>`, the `{error, empty}` tuple is returned, otherwise `{ok, Hash}`. The both cases are then handled differently.

`elixir`
However, what is pledged as an elixir implementation is slightly different.
It is different from the moment the root_hash is generated.
There is a function that is called immediately after the creation of the PMT. The function is part of the library.
And again, a check is made to determine whether the root_hash is empty or not.
And if it is empty = <<>>, a second internal functionality is invoked which digest the empty binary and as a result returns `canonical binary`.

What we could do is to use this `canonical hash` as a `{error, empty}` to fit the scenario.

thepiwo: > What we could do is to use this canonical hash as a {error, empty} to fit the scenario.
I think you should do so

rdimitrovbg: @thepiwo 
Done

apps/aecore/lib/aecore/chain/chainstate.ex
```diff
@@ -20,6 +20,10 @@ defmodule Aecore.Chain.Chainstate do
   @protocol_version_field_size 64
   @protocol_version 17
 
+  @canonical_root_hash <<69, 176, 207, 194, 32, 206, 236, 91, 124, 28, 98, 196, 212, 25, 61, 56,
```
thepiwo: could you explain this again?

rdimitrovbg: @thepiwo 
The way I understand it is the following.
When we create a PMT and no records are made, its root_hash is a empty binary. << >>
For some reason, perhaps related to the policy of development I do not know, in the elixir implementation of PMT, they do this check on root_hash and if it is empty, they have this additional logic on it and return PMT with this canonical_root_hash (ONLY if the RH is <<>>)
else the check is false and the origin PMT is returned.

thepiwo: could you add a comment linking to the definition of this canonical root hash?

rdimitrovbg: @thepiwo 
This is the [function](https://github.com/aeternity/elixir-merkle-patricia-tree/blob/master/lib/trie.ex#L163-L174) which is called right after creating the PMT and deals with an empty root_hash at line 170

rdimitrovbg: <b>Correction</b> on https://github.com/aeternity/elixir-node/pull/484#discussion_r204439591
Not only if the RH is <<>>, <b> but also when the size of the resulting root_hash is greater than `max_rlp_len()` </b>

@thepiwo 
Which constant?

thepiwo: `<<69, 176, 207, 194, 32, 206, 236, 91, 124, 28, 98, 196, 212, 25, 61, 56, 228, 235, 164, 142, 136, 21, 114, 156, 231, 95, 156, 10, 176, 228, 193, 192>>`

rdimitrovbg: @thepiwo 
This is the canonical root hash it is the same for all PMT if they are empty.

thepiwo: can you add a comment in the file to explain that this is the root hash of an empty PMT?

thepiwo: and lets wait for @meivantodorov opinion

apps/aecore/lib/aecore/chain/chainstate.ex
```diff
@@ -104,7 +117,24 @@ defmodule Aecore.Chain.Chainstate do
...
-    AccountStateTree.root_hash(chainstate.accounts)
+    [
+      AccountStateTree.root_hash(chainstate.accounts),
+      NamingStateTree.root_hash(chainstate.naming),
+      OracleStateTree.root_hash(chainstate.oracles)
+    ]
+    |> Enum.reduce(<<@protocol_version::size(@protocol_version_field_size)>>, fn root_hash, acc ->
+      acc <> pad_empty(root_hash)
+    end)
+    |> Hash.hash_blake2b()
```
d-velev: Should be Hash.hash, other than that looks good

Issue #502, opened 2018-07-29
Proper channel state off chain signing form


cytadela8:
Current channel implementation uses removed function (`Serialization.pack_binary`) for packing ChannelStateOffChain for signing. ChannelStateOffChain should be serialized as in epoch for signing.

cytadela8:
After examining epoch implementation: they are doing something very complicated - the process creating the hash for signing includes making Merkle trees and getting their hash. I will do something simpler for now.

PR #503, opened 2018-07-29
[GH-502] Changed ChannelStateOffChain signing_form to RLP

cytadela8:
As in title. The new singing_form isn't as in epoch, but the epoch signing_form is very complicated. It looks like they are creating a merkle tree and adding this tree to their "trees" structure and then finally signing the hash of this structure.


cytadela8:
Ready for review.

Issue #465, opened 2018-07-08
No function clause matching while getting more than 100 blocks


meivantodorov:
Looks like [here](https://github.com/aeternity/elixir-node/blob/master/apps/aecore/lib/aecore/persistence/worker.ex#L289) we are receiving `%{block: %Block{}}` when we expect `%Block{}`

cytadela8:
What's going on with this?

gspasov:
Resolved with PR #506

PR #506, opened 2018-07-30
[GH-#466] Fixed bug where we can't get blocks out of memory

### LINKED ISSUE #466: Fail to get block by height while syncing
Raised error is `key :block not found`. [Link](https://github.com/aeternity/elixir-node/blob/master/apps/aecore/lib/aecore/chain/worker.ex#L157) to the case.

How to reproduce it:

First node A mines few blocks then node B is synced with node A. Then node B is restarted and is synced with node A (still mining) again.

gspasov:
resolves #466 and #465

### DIFF
--- a/apps/aecore/lib/aecore/chain/worker.ex
+++ b/apps/aecore/lib/aecore/chain/worker.ex
@@ -159,7 +159,7 @@ defmodule Aecore.Chain.Worker do
 
   @spec get_header_by_height(non_neg_integer()) :: Header.t() | {:error, reason()}
   def get_header_by_height(height) do
-    case get_block_info_by_height(height, nil) do
+    case get_block_info_by_height(height, nil, :block) do
       {:error, :chain_too_short} -> {:error, :chain_too_short}
       info -> {:ok, info.block.header}
     end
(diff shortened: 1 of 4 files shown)


apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -411,7 +397,7 @@ defmodule Aecore.Chain.Worker do
         ## Transfrom from chain state
         :chain_state => %{
           new_block_hash =>
-            transfrom_chainstate(:from_chainstate, Map.from_struct(new_chain_state))
+            transfrom_chainstate(:from_chainstate, {:ok, Map.from_struct(new_chain_state)})
```
rdimitrovbg: Would you give the correct function name?

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -443,7 +429,7 @@ defmodule Aecore.Chain.Worker do
       Persistence.batch_write(%{
         :chain_state => %{
           new_block_hash =>
-            transfrom_chainstate(:from_chainstate, Map.from_struct(new_chain_state))
+            transfrom_chainstate(:from_chainstate, {:ok, Map.from_struct(new_chain_state)})
```
rdimitrovbg: Would you give the correct function name?

PR #515, opened 2018-08-03
[GH-#509] Integers are now encoded as binaries before RLP encoding

### LINKED ISSUE #509: Adjust RLP encoding and decoding
ExRLP library producing wrong results.

Artur64:
resolves #509 .


apps/aeutil/lib/serialization.ex
```diff
@@ -548,6 +548,10 @@ defmodule Aeutil.Serialization do
     DataTx.rlp_decode(payload, datatx)
   end
 
+  def transform_item(item) when is_integer(item) do
```
thepiwo: `encode_unsigned` describes this better

DanielaIvanova:
After RLP encoding the accounts and oracles have same hashes in epoch and elixir nodes.

PR #522, opened 2018-08-08
[GH-521] More Travis chaching

### LINKED ISSUE #521: Travis speedup
Caching can be utilised for faster Travis builds.

cytadela8:
This PR adds few folders to Travis cache and decreases build time from about 8min30 to about 5min30.
Following was added to cache:
- libsodium build, but not configure
- cuckoo
- aevm tests files

Cache size increased from 323.83MB to 499.62MB

To speed up tests further we should create some mock mining for tests. (Running cuckoo is the time consuming part of tests) Aside from that installing rust takes 25sec, libsodium/configure takes 15sec and initializing build with downloading cache takes about 1min.

This closes #521

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,3 +28,6 @@ cache:
     - $HOME/.cargo
     - deps
     - _build
+    - aevm_external
+    - apps/aecore/src/cuckoo/c_src
+    - apps/aecore/priv/cuckoo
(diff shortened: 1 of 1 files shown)


cytadela8:
Build time decrease is hard to estimate, because cuckoo makes test time a little bit random. (and I believe Travis behaves also a little bit random) The time reduce should be at leasts 2min. (Note first build on new config is always longer, but subsequent are faster)

PR #523, opened 2018-08-09
[GH-#467] Refactored Key management module

### LINKED ISSUE #467: adjust keys api to create multiple separate curve25519 pairs
elixir-wallet can be removed as dependency in this case

gspasov:
resolves #467 

Changes:
- Signing keys use the ed25519 curve as in epoch;
- Now pubkey is 32 bytes as it should be;
- Removed GenServer from the keys management;
- Now Sign keys and Peer keys are being managed by the only Keys module;
- We can create new sign key pair by adding something to the path:
`Keys.keypair(:sign)` -> Created the default miner keypair for signing
`Keys.keypair(:sign, "1")` -> Creates different keypair (and this keypair is saved in different files)


apps/aecore/lib/aecore/keys/keys.ex
```diff
@@ -0,0 +1,198 @@
...
+  @doc """
+  Returns a signed version of the given binary
+  """
+  @spec sign(binary()) :: binary()
+  def sign(message) when is_binary(message) do
+    {_, privkey} = keypair(:sign)
+    sign(message, privkey)
+  end
+
+  @spec sign(binary(), sign_priv_key()) :: binary()
```
thepiwo: type-aliases for message and signature would be nice

apps/aecore/lib/aecore/keys/keys.ex
```diff
@@ -0,0 +1,207 @@
...
+      do: true
+
+  def key_size_valid?(pubkey)
+      when byte_size(pubkey) == @pub_size,
+      do: true
+
+  def key_size_valid?(_), do: false
+
+  @spec keypair(keypair_type(), String.t()) :: sign_keypair() | peer_keypair()
+  def keypair(type, suffix \\ "") do
```
d-velev: Can you split this into Keys.sign_pubkey, Keys.peer_pubkey and so on for testing purposes

apps/aecore/lib/aecore/keys/keys.ex
```diff
@@ -0,0 +1,207 @@
...
+
+  @spec sign(message(), sign_priv_key()) :: signature()
+  def sign(message, privkey) when is_binary(message) and is_binary(privkey) do
+    :enacl.sign_detached(message, privkey)
+  end
+
+  @doc """
+  Checks if the message is signed with the given public key
+  """
+  @spec verify(message(), signature()) :: true | false
```
d-velev: true | false can be replaced with boolean()

apps/aecore/lib/aecore/keys/keys.ex
```diff
@@ -0,0 +1,207 @@
...
+  @peer_key_decode "pp$"
+
+  @filename_sign_pub "sign_key.pub"
+  @filename_sign_priv "sign_key"
+
+  @filename_peer_pub "peer_key.pub"
+  @filename_peer_priv "peer_key"
+
+  @doc """
+  Returns a signed version of the given binary
```
d-velev: Returns a signature of the given binary

apps/aecore/lib/aecore/keys/keys.ex
gspasov: This was made intentionally, to keep the code as small and simple as possible. For testing just add an atom `:sign | :peer` to get the specific keypair

d-velev: It's better to have Keys.sign_pubkey instead of typing out Keys.keypair(:sign) |> elem(0) every time, people might not know at what position the key they want is

gspasov: Then I'll add to the `doc` of the function. I don't feel like changing a function for the sake of writing test easier.

Issue #491, opened 2018-07-20
Channel snapshot


cytadela8:
- [ ] Add channel snapshot tx
- [ ] Add snapshoting to Channel.Worker
- [ ] tests

cytadela8:
This will be done by Grzegorz Uriasz.

grzegorz225:
Resolved with PR #525

PR #525, opened 2018-08-11
[GH-491] Channel snapshot

grzegorz225:
Implemented ChannelSnapshotSoloTx transaction, added snapshots to Aecore.Channel.Worker, added tests.


apps/aecore/lib/aecore/channel/worker.ex
```diff
@@ -67,6 +67,11 @@ defmodule Aecore.Channel.Worker do
     closed(tx)
   end
 
+  def new_tx_mined(%SignedTx{data: %DataTx{type: ChannelSnapshotSoloTx}} = _tx) do
+    # TODO
```
cytadela8: I think we shouldn't do anything in this case (replace the comment with `SnapshotTx requires no action`)

cytadela8:
_Note: After doing this task we realised that SnapshotTx is in docs, but not in 0.16.0 of epoch. Fortunately this change has soft fork behaviour I won't prevent as from syncing with epoch. We can blacklist this transaction later for full compliance. _

thepiwo:
@cytadela8 can you explain the blacklist? so you suggest to merge, but "disable" the code? We can also not merge and add a label to the PR, or make a Channels branch were we merge to. What do you prefer @cytadela8 @grzegorz225 @d-velev

apps/aecore/lib/aecore/channel/tx/channel_snapshot_solo.ex
```diff
@@ -0,0 +1,205 @@
...
+  @spec validate(ChannelSnapshotSoloTx.t(), DataTx.t()) :: :ok | {:error, String.t()}
+  def validate(%ChannelSnapshotSoloTx{state: %ChannelStateOffChain{sequence: sequence}}, data_tx) do
+    senders = DataTx.senders(data_tx)
+
+    cond do
+      length(senders) != 1 ->
+        {:error, "#{__MODULE__}: Invalid senders size"}
+
+      sequence == 0 ->
+        {:error, "#{__MODULE__}: Can't slash with zero state"}
```
d-velev: This message doesn't make sense right?

apps/aecore/lib/aecore/channel/tx/channel_snapshot_solo.ex
```diff
@@ -0,0 +1,205 @@
...
+  def deduct_fee(accounts, block_height, _tx, data_tx, fee) do
+    DataTx.standard_deduct_fee(accounts, block_height, data_tx, fee)
+  end
+
+  @spec is_minimum_fee_met?(SignedTx.t()) :: boolean()
+  def is_minimum_fee_met?(tx) do
+    tx.data.fee >= Application.get_env(:aecore, :tx_data)[:minimum_fee]
+  end
+
+  def encode_to_list(%ChannelSnapshotSoloTx{} = tx, %DataTx{} = datatx) do
```
d-velev: It's datatx here and data_tx above, please make it consistent, also some structure fields are matched and some are accessed with ".",  could you change that to be consistent as well?

apps/aecore/lib/aecore/channel/worker.ex
```diff
@@ -511,6 +516,25 @@ defmodule Aecore.Channel.Worker do
...
+    with {:ok, tx} <-
+      ChannelStatePeer.snapshot(peer_state, fee, nonce, priv_key),
+         :ok <- Pool.add_transaction(tx) do
+        {:reply, :ok, state}
+      else
+      {:error, _} = err ->
+        {:reply, err, state}
+
+      :error ->
+        {:error, "#{__MODULE__}: Pool error"}
```
d-velev: why not reply here as well? also now that I see that Pool.add_transaction replies with error and no message which requires you to do 2 different checks just for errors, could you make it reply with {:error, message}

apps/aecore/lib/aecore/channel/channel_state_on_chain.ex
```diff
@@ -188,6 +188,49 @@ defmodule Aecore.Channel.ChannelStateOnChain do
...
+    }
+  end
+
+  @doc """
+  Validates Snapshot states.
+  """
+  @spec validate_snapshot(ChannelStateOnChain.t(), ChannelStateOffChain.t()) ::
+          :ok | {:error, binary()}
+  def validate_snapshot(
+        %ChannelStateOnChain{} = _channel,
```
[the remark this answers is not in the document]
d-velev: I see that this is done in other places as well, if it's done for readability then I don't think it's needed. Some functions are also missing specs

cytadela8:
@thepiwo Making channels branch or not merging is also an option. As for my blacklist idea. This code only adds a new transaction:
- Our node will be able to sync with epoch regardless off this PR (epoch doesn't use this transaction)
- If this is merged and nothing else is done our code will create a fork that will be considered invalid by epoch if someone creates a snapshotTx. (epoch 0.16.0 doesn't understand snapshotx) We can solve that by (for example) removing snapshotTx from whitelist in DataTx.valid_types. (We can put valid_types in config or have some other override so new tests still pass)

cytadela8:
@grzegorz225 Add example usage to `Readme.md`

grzegorz225:
Ok, I've updated the README and blacklisted this transaction so it can be merged into master without causing forks.

thepiwo:
@d-velev can you re-review?

PR #526, opened 2018-08-13
[GH-#507] Contracts and calls state trees 

### LINKED ISSUE #507: Contracts and calls state trees
We need a data structure to insert and retrieve contracts and calls from. 
This issues is going to contain contracts and calls because they are linked to the trees with the same name.

https://github.com/aeternity/protocol/blob/master/contracts/contract_state_tree.md

nedelcho-delchev-tues:
Implemented initial version of calls and contracts and their state trees 

resolves #507


apps/aecore/lib/aecore/contract/contract_state_tree.ex
```diff
@@ -0,0 +1,118 @@
...
+  @type hash :: binary()
+  @type contracts_state() :: Trie.t()
+
+  @spec init_empty() :: contracts_state()
+  def init_empty do
+    PatriciaMerkleTree.new(:contracts)
+  end
+
+  @spec insert_contract(contracts_state(), map()) :: contracts_state()
+  def insert_contract(contract_tree, contract) do
```
gspasov: There are a few places like this where you could pattern match a value out of the `%Contract` struct. In this case the `contract.id`

apps/aecore/lib/aecore/contract/contract_state_tree.ex
```diff
@@ -0,0 +1,118 @@
...
+  end
+
+  @spec get_contract(contracts_state(), binary()) :: map()
+  def get_contract(contract_tree, key) do
+    case PatriciaMerkleTree.lookup(contract_tree, key) do
+      {:ok, serialized} ->
+        {:ok, deserialized} = Serialization.rlp_decode_anything(serialized)
+
+        identified_id = Identifier.create_identity(key, :contract)
+        # identified_owner = Identifier.create_identity(deserialized.owner, :account)
```
gspasov: Is this supposed to be commented out?

apps/aeutil/lib/type_to_tag.ex
```diff
@@ -20,10 +20,10 @@ defmodule Aeutil.TypeToTag do
   def tag_to_type(34), do: {:ok, Aecore.Naming.Tx.NameUpdateTx}
   def tag_to_type(35), do: {:ok, Aecore.Naming.Tx.NameRevokeTx}
   def tag_to_type(36), do: {:ok, Aecore.Naming.Tx.NameTransferTx}
-  # Contract - 40
-  # Contract call - 41
-  # Contract create transaction - 42
-  # Contract call transaction - 43
+  def tag_to_type(40), do: {:ok, Aecore.Contract.Contract}
+  def tag_to_type(41), do: {:ok, Aecore.Contract.Call}
+  # def tag_to_type(42), do: {:ok, Aecore.Contract.ContractCreateTx}
```
thepiwo: are they supposed to be commented out?

apps/aeutil/lib/type_to_tag.ex
```diff
@@ -57,10 +57,10 @@ defmodule Aeutil.TypeToTag do
   def type_to_tag(Aecore.Naming.Tx.NameUpdateTx), do: {:ok, 34}
   def type_to_tag(Aecore.Naming.Tx.NameRevokeTx), do: {:ok, 35}
   def type_to_tag(Aecore.Naming.Tx.NameTransferTx), do: {:ok, 36}
-  # Contract - 40
-  # Contract call - 41
-  # Contract create transaction - 42
-  # Contract call transaction - 43
+  def type_to_tag(Aecore.Contract.Contract), do: {:ok, 40}
+  def type_to_tag(Aecore.Contract.Call), do: {:ok, 41}
+  # def type_to_tag(Aecore.Contract.ContractCreateTx), do: {:ok, 42}
```
thepiwo: are they supposed to be commented out?

apps/aecore/lib/aecore/contract/call.ex
```diff
@@ -0,0 +1,119 @@
...
+  """
+  alias Aecore.Chain.Identifier
+  alias Aecore.Contract.Call
+  alias Aeutil.Parser
+  alias Aeutil.Hash
+
+  @version 1
+
+  @type t :: %Call{
+          caller_address: Wallet.pubkey(),
```
thepiwo: This is not Wallet any more, also the alias is missing here

apps/aecore/lib/aecore/contract/contract.ex
```diff
@@ -0,0 +1,134 @@
+defmodule Aecore.Contract.Contract do
+  @moduledoc """
+  Aecore contact module implementation.
+  """
+  alias Aecore.Contract.Contract
+  alias Aecore.Chain.Identifier
+  alias Aecore.Keys.Wallet
```
thepiwo: incorrect alias to wallet

thepiwo:
@Artur64 @DanielaIvanova @gspasov will you review?

apps/aecore/lib/aecore/contract/call.ex
```diff
@@ -0,0 +1,127 @@
+defmodule Aecore.Contract.Call do
+  @moduledoc """
+  Aecore call module implementation.
+  """
+  alias Aecore.Chain.Identifier
+  alias Aecore.Contract.Call
+  alias Aeutil.Parser
+  alias Aeutil.Hash
+
```
cytadela8: You should add `use Aecore.Util.Serializable` here probably.

This makes sure you follow behaviour and adds `rlp_encode`, `rlp_decode` functions.

apps/aecore/lib/aecore/contract/call.ex
```diff
@@ -0,0 +1,127 @@
...
+        encoded_caller_address,
+        caller_nonce,
+        height,
+        encoded_contract_address,
+        gas_price,
+        gas_used,
+        return_value,
+        return_type
+      ]) do
+    {:ok, decoded_caller_address} = Identifier.decode_from_binary(encoded_caller_address)
```
cytadela8: Please handle errors correctly. I would use a `with` statement.

apps/aecore/lib/aecore/contract/call.ex
```diff
@@ -0,0 +1,127 @@
...
+    {:ok,
+     %Call{
+       caller_address: decoded_caller_address,
+       caller_nonce: :binary.decode_unsigned(caller_nonce),
+       height: :binary.decode_unsigned(height),
+       contract_address: decoded_contract_address,
+       gas_price: :binary.decode_unsigned(gas_price),
+       gas_used: :binary.decode_unsigned(gas_used),
+       return_value: return_value,
+       return_type: String.to_atom(return_type)
```
cytadela8: I'm not sure what the impact would be, but maybe we should check if `return_type` is one of the 3 valid.

apps/aecore/lib/aecore/contract/contract.ex
```diff
@@ -0,0 +1,136 @@
...
+  def decode_from_list(@version, [
+        owner,
+        vm_version,
+        code,
+        log,
+        active,
+        referers,
+        deposit
+      ]) do
+    {:ok, decoded_owner_address} = Identifier.decode_from_binary(owner)
```
cytadela8: Handle errors also here.

apps/aecore/lib/aecore/contract/contract.ex
```diff
@@ -0,0 +1,136 @@
+defmodule Aecore.Contract.Contract do
+  @moduledoc """
+  Aecore contact module implementation.
+  """
+  alias Aecore.Contract.Contract
+  alias Aecore.Chain.Identifier
+  alias Aeutil.Hash
+
```
cytadela8: `use Aecore.Util.Serializable` also here

apps/aecore/lib/aecore/contract/contract_state_tree.ex
```diff
@@ -0,0 +1,113 @@
...
+  def get_contract(contract_tree, key) do
+    case PatriciaMerkleTree.lookup(contract_tree, key) do
+      {:ok, serialized} ->
+        {:ok, deserialized} = Serialization.rlp_decode_anything(serialized)
+
+        identified_id = Identifier.create_identity(key, :contract)
+
+        identified_referers =
+          Enum.map(deserialized.referers, fn referer ->
+            {:ok, identified_referer} = Identifier.create_identity(referer, :contract)
```
cytadela8: This seams weird. Don't we store encoded Identifier inside contract encoding and decode it? I think after decoding we already have an Identifier structure in decoded contract.

cheezus1: Yes, my mistake. I tested and fixed it just now.

DanielaIvanova:
Maybe will be great if you add some documentation in `README.md`

cheezus1:
@cytadela8 can you review again?
@DanielaIvanova We will add documentation in the readme once we finish everything related with the creating and calling of contracts.

apps/aecore/lib/aecore/contract/call.ex
```diff
@@ -0,0 +1,140 @@
...
+        return_value,
+        return_type
+      ]) do
+
+    parsed_return_type =
+      case return_type do
+        return_type when return_type in ["ok", "error", "revert"] ->
+            String.to_atom(return_type)
+        _ ->
+          {:error, "#{__MODULE__}: decode_from_list: Invalid return_type: #{inspect(return_type)}"}
```
cytadela8: This doesn't return error. It just sets `parsed_return_type` to `{:error, ...}`.

apps/aecore/lib/aecore/contract/contract.ex
```diff
@@ -0,0 +1,142 @@
...
+      ]) do
+    decoded_active =
+      case :binary.decode_unsigned(active) do
+        0 -> false
+        1 -> true
+      end
+
+    decoded_referers =
+      Enum.map(referers, fn referer ->
+        {:ok, decoded_referer} = Identifier.decode_from_binary(referer)
```
cytadela8: This doesn't handle errors.

apps/aecore/lib/aecore/contract/contract.ex
```diff
@@ -0,0 +1,142 @@
...
+        code,
+        log,
+        active,
+        referers,
+        deposit
+      ]) do
+    decoded_active =
+      case :binary.decode_unsigned(active) do
+        0 -> false
+        1 -> true
```
cytadela8: You should handle situation where `active` represents other integer then 0 or 1.

apps/aecore/lib/aecore/contract/call.ex
cytadela8: (You can for example put the below with statement with `{:ok, %Call{...}}` inside this `case ... do`, or add a check to `with` and set `parsed_return_type` to `{:ok, ...}` on success)

apps/aecore/lib/aecore/contract/contract.ex
cytadela8: You can (for example) first decode and set to `{:error, ...}` on error and then use `Enum.reduce_while` to check if there was any error.

cytadela8:
@Artur64 @d-velev @gspasov Could you rereview?

apps/aecore/lib/aecore/contract/contract.ex
```diff
@@ -0,0 +1,167 @@
...
+        active,
+        referers,
+        deposit
+      ]) do
+    decoded_active = decode_active(active)
+
+    decoded_referers =
+      Enum.reduce_while(referers, [], fn referer, acc ->
+        with {:ok, decoded_referer} <- Identifier.decode_from_binary(referer) do
+          [decoded_referer | acc]
```
gspasov: This possibly will crash. 
`Enum.reduce_while` requires tuple `{:cont, var}` in order to continue. And tuple `{:halt, var}` to exit the reduce (which you already have).
Additionally, the way this is written, in the first time the check fails, you will exit the reduce with the `{error, reason}` value for `decoded_referers`. I don't know if this is the required end result.

The `with` statement here could be changed for a simple `case`, because you do only one check, therefore no need for it in this example.

apps/aecore/lib/aecore/contract/contract.ex
```diff
@@ -0,0 +1,167 @@
...
+        else
+          _ ->
+            {:halt,
+             {:error,
+              "#{__MODULE__}: decode_from_list: Invalid contract referer: #{inspect(referer)}"}}
+        end
+      end)
+
+    case decoded_active do
```
gspasov: This here doesn't look really good. In this case you could use the `with` statement very effectively, but in order to do so the func `decode_active` should return: `{:ok, true}` and `{:ok, false}` so that you catch it in the with like so: `{:ok, _} <- decode_active`.

The decoded_referers you could catch like: `[_] <- decoded_referers`, because you will either end up with list of values or with a simple error variable (This depends on how you do the `reduce_while` func).

Doing this you will save much lines and make it more readable/understandable

apps/aecore/lib/aecore/contract/contract_state_tree.ex
```diff
@@ -0,0 +1,107 @@
...
+
+  defp get_store(store_id, tree) do
+    keys = PatriciaMerkleTree.all_keys(tree)
+    store_id_bit_size = (@contract_key_size + 1) * 8
+
+    Enum.reduce(keys, %{}, fn key, store_acc ->
+      if byte_size(key) > @contract_key_size do
+        <<tree_store_id::size(store_id_bit_size), s_key::binary>> = key
+
+        if store_id == <<tree_store_id::size(store_id_bit_size)>> do
```
gspasov: The double if here could be escaped with a `with` statement as well, because in both `else` cases you return the same value

thepiwo:
whats the status here?

PR #550, opened 2018-08-20
[GH-417] Multi Node Test Framework

0legIv:
Implementation of Multi node test framework, has such functionalities:

1. Run unlimited number of nodes parallely, controlling their state from the main node.
2. Test all the types of transactions(except for the channels) in each node. 
3. Send tokens between nodes using miner accounts.
4. Sync the nodes and compare them by top block hash or top block(sync test)
5. See the oracle interaction objects state for each node.


apps/aetestframework/.formatter.exs
```diff
@@ -0,0 +1,4 @@
+# Used by "mix format"
```
thepiwo: i assume this file is not needed and these rules should be included in the parent formatter config

apps/aetestframework/.gitignore
```diff
@@ -0,0 +1,24 @@
+# The directory Mix will write compiled artifacts to.
```
thepiwo: this file is not needed

apps/aetestframework/config/config.exs
```diff
@@ -0,0 +1,30 @@
+# This file is responsible for configuring your application
```
thepiwo: this file is not needed

apps/aetestframework/mix.exs
```diff
@@ -0,0 +1,33 @@
...
+  def application do
+    [
+      extra_applications: [:logger]
+    ]
+  end
+
+  # Run "mix help deps" to learn about dependencies.
+  defp deps do
+    [
+      # {:dep_from_hexpm, "~> 0.3.0"},
```
thepiwo: this can be empty

apps/aetestframework/mix.exs
```diff
@@ -0,0 +1,33 @@
+defmodule Aetestframework.MixProject do
+  use Mix.Project
+
+  def project do
+    [
```
thepiwo: are all of these needed? and equal to the other mixfiles?

mix.lock
```diff
@@ -45,6 +45,7 @@
   "sext": {:git, "https://github.com/uwiger/sext.git", "615eebcf975ec4b4561c6f2b2bc433dabdb2be0f", [tag: "1.4.1"]},
   "sha3": {:git, "https://github.com/szktty/erlang-sha3.git", "dbdfd12ace2fc7e530e697fef81e1e72e90740f9", [ref: "dbdfd12"]},
   "ssl_verify_fun": {:hex, :ssl_verify_fun, "1.1.1", "28a4d65b7f59893bc2c7de786dec1e1555bd742d336043fe644ae956c3497fbe", [:make, :rebar], [], "hexpm"},
+  "temp": {:hex, :temp, "0.4.5", "c62ee92c778ddebb2738212fecea372f1a9beba10bacbd7e54e7efe29357a82e", [:mix], [], "hexpm"},
```
thepiwo: why is this added in the lockfile, but not mixfile?

0legIv: Unused library, will remove it

apps/aetestframework/mix.exs
0legIv: Yes, because it's a separate app and I can't remove the mix file as it's a part of the umbrella project structure

thepiwo: the other apps specify way less things here

thepiwo: e.g. https://github.com/aeternity/elixir-node/blob/master/apps/aeutil/mix.exs

apps/aetestframework/lib/worker.ex
```diff
@@ -0,0 +1,607 @@
...
+  end
+
+  @doc """
+    Checking if the port is busy
+  """
+  def busy_port?(port) do
+    :os.cmd('lsof -i -P -n | grep -w #{port}') != []
+  end
+
+  defp update_data(state, result, node, :peers) do
```
cytadela8: I think you could use 3 functions instead of having type of update as last parameter to this function.

apps/aetestframework/test/multi_node_sync_test.exs
```diff
@@ -0,0 +1,147 @@
+defmodule MultiNodeSyncTest do
+  use ExUnit.Case
+
+  alias Aetestframework.Worker, as: TestFramework
+  alias Aecore.Chain.Worker, as: Chain
+  alias Aecore.Tx.Pool.Worker, as: Pool
+
+  setup do
+    TestFramework.start_link(%{})
+    Chain.clear_state()
```
cytadela8: You should use new way of cleaning blockchain state, but I don't think you use the blockchain state at all so you might not need this.

cytadela8:
Also, all tests fail for me with messages similar to this:
```
 1) test namings test (MultiNodeSyncTest)
     test/multi_node_sync_test.exs:83
     ** (EXIT from #PID<0.581.0>) an exception was raised:
         ** (ArgumentError) argument error                                                                                                                                  
             :erlang.port_command(#Port<0.13578>, "{:ok, peer_info} = Client.get_info(\"localhost:4002\")\n")                                                               
             (aetestframework) lib/worker.ex:493: Aetestframework.Worker.handle_call/3                                                                                      
             (stdlib) gen_server.erl:636: :gen_server.try_handle_call/4                                                                                                     
             (stdlib) gen_server.erl:665: :gen_server.handle_msg/6                                                                                                          

12:43:31.458 [error] GenServer Aetestframework.Worker terminating
** (ArgumentError) argument error                                                                                                                                           
    :erlang.port_command(#Port<0.13578>, "{:ok, peer_info} = Client.get_info(\"localhost:4002\")\n")                                                                        
    (aetestframework) lib/worker.ex:493: Aetestframework.Worker.handle_call/3                                                                                               
    (stdlib) gen_server.erl:636: :gen_server.try_handle_call/4                                                                                                              
    (stdlib) gen_server.erl:665: :gen_server.handle_msg/6                                                                                                                   
Last message (from #PID<0.581.0>): {:sync_two_nodes, "node1", "node2"}
```

Do you know why?

0legIv:
@cytadela8 Can you run firstly `iex -S mix phx.server`, and after that `mix test`, I am not sure but probably there is a problem that there is no _build/dev directory when you run `mix test` after you cloned the branch

apps/aetestframework/lib/worker.ex
```diff
@@ -0,0 +1,607 @@
...
+
+  @doc """
+    Updates top block state for 2 nodes and compares them.
+    If top block of each node is equal - they are synced
+  """
+  @spec compare_nodes_by_top_block(String.t(), String.t()) :: :synced | :not_synced
+  def compare_nodes_by_top_block(node_name1, node_name2) do
+    get_node_top_block(node_name1)
+    get_node_top_block(node_name2)
+    GenServer.call(__MODULE__, {:compare_nodes_by_top_block, node_name1, node_name2})
```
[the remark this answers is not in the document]
0legIv: Yes, will change it

0legIv:
@cytadela8 And about the generic way of executing the code on node and getting the return value, I was researching this and I found out that it is not possible, because the return value is always string and it has some additional unnecessary elements like `iex>`, that's why the result has to be filtered somehow. Probably I am wrong, and there is a way to do this, maybe it is possible using another library instead of Port

cytadela8:
@0legIv You could do for example: `Base.encode64(:erlang.term_to_binary(function_to_invoke))` and then regexp to match contents between `"`

cytadela8:
If that would match something else then we could try things like `{:return, Base.encode64(:erlang.term_to_binary(function_to_invoke))}` and regexp between `{:return, "` and `"}`

0legIv:
@cytadela8 Yes, great idea, changed it

cytadela8:
@0legIv Nice! Only the setup in test left. (see previous comment)

thepiwo:
whats the status here? is it finished, so could you review or should we close for now?

0legIv:
@thepiwo Yes it's finished, I just need to fix the problem that it cytadela wrote in the previous comments and, I'm waiting for another reviews.

thepiwo:
It would be good that our tests actually show coverage, so we can check what parts of code are tested and what are not

0legIv:
@thepiwo Yes, I can make a new issue for this

thepiwo:
@cytadela8 can you re-review?

apps/aetestframework/test/multi_node_sync_test.exs
cytadela8: > You should use new way of cleaning blockchain state, but I don't think you use the blockchain state at all so you might not need this.

This comment still applies

cytadela8: > You should use new way of cleaning blockchain state, but I don't think you use the blockchain state at all so you might not need this.

This comment still applies

apps/aetestframework/test/multi_node_sync_test.exs
```diff
@@ -0,0 +1,135 @@
...
```
cytadela8: > You should use new way of cleaning blockchain state, but I don't think you use the blockchain state at all so you might not need this.

This comment still applies

0legIv: Yes, I removed this as I don't use the chainstate at all in the general node

cytadela8:
All of my previous comments are fixed, but merging probably introduced something that makes tests fail.

PR #553, opened 2018-08-24
[GH-#527] Adjusted naming system

### LINKED ISSUE #527: Adjust naming system

DanielaIvanova:
resolves #527


apps/aecore/test/aecore_naming_test.exs
```diff
@@ -378,12 +391,12 @@ defmodule AecoreNamingTest do
     first_name_claim = NamingStateTree.get(naming_state_2, claim_hash_1)
 
     assert {:ok, first_name_claim.hash.value} == NameUtil.normalized_namehash("test.aet")
-    assert first_name_claim.owner.value == elem(Keys.keypair(:sign), 0)
+    assert first_name_claim.owner == elem(Keys.keypair(:sign), 0)
 
     assert first_name_claim.status == :claimed
-    assert first_name_claim.pointers == []
+    assert first_name_claim.pointers == "[]"
```
cytadela8: Is it supposed to be like that? (a string=`[]` instead of a an empty list)?

DanielaIvanova: Yes, it is supposed to be like this because we are not using `Poison.encode([])`. We are just hard coding `"[]"`

apps/aecore/lib/aecore/naming/name.ex
```diff
@@ -1,81 +1,70 @@
-defmodule Aecore.Naming.NameClaim do
+defmodule Aecore.Naming.Name do
```
thepiwo: why is the name changed?

apps/aecore/lib/aecore/naming/name_util.ex
```diff
@@ -28,32 +28,39 @@ defmodule Aecore.Naming.NameUtil do
   def normalize_name(name), do: name |> :idna.utf8_to_ascii() |> to_string()
 
   @spec namehash(String.t()) :: binary()
-  defp namehash(name) do
+  def namehash(name) do
```
thepiwo: I assume only the normalized_namehash should be uses, thus this function was private

apps/aecore/lib/aecore/naming/name_util.ex
```diff
@@ -28,32 +28,39 @@ defmodule Aecore.Naming.NameUtil do
...
   end
 
   @spec validate_normalized_name(String.t()) :: :ok | {:error, String.t()}
   defp validate_normalized_name(name) do
     allowed_registrar =
       GovernanceConstants.name_registrars()
-      |> Enum.any?(fn registrar -> String.ends_with?(name, registrar) end)
+      |> Enum.any?(fn registrar ->
+        String.ends_with?(name, registrar) &&
+          Enum.count(String.split(name, GovernanceConstants.split_name_symbol())) == 2
```
thepiwo: what is the 2 used for? can you extract this as a named constant?

apps/aecore/lib/aecore/naming/name.ex
DanielaIvanova: We found this name more compatible with epoch node

thepiwo: okay

apps/aecore/lib/aecore/naming/tx/name_claim_tx.ex
```diff
@@ -62,9 +60,9 @@ defmodule Aecore.Naming.Tx.NameClaimTx do
       validate_name |> elem(0) == :error ->
         {:error, "#{__MODULE__}: #{validate_name |> elem(1)}: #{inspect(name)}"}
 
-      byte_size(name_salt) != NameClaim.get_name_salt_byte_size() ->
+      !is_integer(name_salt) ->
         {:error,
-         "#{__MODULE__}: Name salt bytes size not correct: #{inspect(byte_size(name_salt))}"}
+         "#{__MODULE__}: Name salt is not correct: #{inspect(byte_size(name_salt))}, should be integer"}
```
d-velev: rempve byte_size()

apps/aecore/lib/aecore/naming/name_claim.ex
```diff
@@ -0,0 +1,128 @@
+defmodule Aecore.Naming.NameClaim do
```
thepiwo: is this supposed to be re-added?

DanielaIvanova: No, I removed it

thepiwo:
@cytadela8 @d-velev can you re-review?

PR #557, opened 2018-08-24
[GH-#512] Transaction pool sync

### LINKED ISSUE #512: Transaction pool sync
Pool sync was changed in epoch at some point, we have to implement this to make sync work since we would receive messages we can't handle

d-velev:
Comparison of pools between epoch nodes is done with MPTs so for now we tell the epoch node that we have an empty tree in order to get its transactions (this means that syncing isn't optimized, I'll try to fix this after the 31st). There might be some messy stuff leftover which was needed to match the genesis hash, I'll clean that up soon.

closes #512


apps/aecore/lib/aecore/chain/block.ex
```diff
@@ -24,11 +24,39 @@ defmodule Aecore.Chain.Block do
     @version
   end
 
+  # @spec genesis_header() :: Header.t()
+  # defp genesis_header do
+  #   header = Application.get_env(:aecore, :pow)[:genesis_header]
+  #   struct(Header, header)
+  # end
+
+  def genesis_header do
```
gspasov: Why create this function if there is one already doing this in the [Genesis module](https://github.com/aeternity/elixir-node/blob/master/apps/aecore/lib/aecore/chain/genesis.ex#L71)?

thepiwo:
what is the status here?

apps/aecore/lib/aecore/chain/block.ex
d-velev: As I said there is some stuff that needs to be removed, I'll take care of it

gspasov: I didn't know you are still working on this and aware of the issue in hand. I guess there was some miscommunication on my part.

PR #562, opened 2018-08-28
[GH-556] Adjustment to Oracles

### LINKED ISSUE #556: Adjust Oracle system

Artur64:
resolves #556


apps/aecore/lib/aecore/oracle/tx/oracle_response_tx.ex
```diff
@@ -94,7 +94,7 @@ defmodule Aecore.Oracle.Tx.OracleResponseTx do
     updated_interaction_objects = %{
       interaction_objects
       | response: tx.response,
-        expires: interaction_objects.expires + interaction_objects.response_ttl,
+        expires: interaction_objects.response_ttl + block_height,
```
thepiwo: so response expiry is relative to current block height?

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
```diff
@@ -192,7 +193,14 @@ defmodule Aecore.Oracle.OracleStateTree do
     case PatriciaMerkleTree.lookup(tree, key) do
       {:ok, serialized} ->
         {:ok, deserialized} = Serialization.rlp_decode_anything(serialized)
-        deserialized
+
+        case deserialized do
+          %Oracle{} ->
+            %{deserialized | owner: Identifier.create_identity(key, :oracle)}
```
thepiwo: is the owner supposed to be `:oracle`?

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
```diff
@@ -192,7 +193,14 @@ defmodule Aecore.Oracle.OracleStateTree do
...
+
+          _ ->
```
d-velev: what exactly is this case

apps/aecore/lib/aecore/oracle/tx/oracle_response_tx.ex
DanielaIvanova: We adjusted like epoch does. `expires` is sum of `response_ttl` and `block_height`

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
DanielaIvanova: Yes, it supposed to be like that, same as `epoch`

apps/aecore/lib/aecore/oracle/oracle_state_tree.ex
DanielaIvanova: Can be `%OracleQuery{}` as well.

PR #567, opened 2018-08-29
[GH-524] Rewrite Sync module

### LINKED ISSUE #451: fix nakamoto consensus
allow chain to be overwritten by longer ones

### LINKED ISSUE #524: Rewrite Sync module
Sync module needs to be rewritten so that we can be able to sync with Epoch node.
There are differences in how sync is handled by epoch atm.
This rewriting of the module will additionally make the sync faster (close or even as fast as Epoch does it).

### LINKED ISSUE #560: Bug when trying to send new tx to the synced node
How to reproduce:
1. Run two nodes: `make iex-1`, `make iex-2`
2. Sync them
3. Add a new tx to the pool in one of the nodes, for example spend_tx:
```
{pubkey, _} = Keys.keypair :sign
{:ok, tx} = Account.spend(pubkey, 20, 10, <<"payload">>)
Pool.add_transaction(tx)
```

Error:
```
[error] GenServer Aecore.Peers.Jobs terminating
** (UndefinedFunctionError) function Aecore.Peers.PeerConnection.send_new_tx/2 is undefined or private. Did you mean one of:

      * send_new_block/2

    (aecore) Aecore.Peers.PeerConnection.send_new_tx(%Aecore.Tx.SignedTx{data: %Aecore.Tx.DataTx{fee: 10, nonce: 1, payload: %Aecore.Account.Tx.SpendTx{amount: 20, payload: "payload", receiver: %Aecore.Chain.Identifier{type: :account, value: <<131, 38, 207, 89, 116, 83, 32, 200, 7, 116, 133, 153, 178, 250, 54, 237, 148, 103, 50, 110, 16, 15, 131, 194, 232, 119, 169, 120, 55, 221, 106, 81>>}, version: 1}, senders: [%Aecore.Chain.Identifier{type: :account, value: <<131, 38, 207, 89, 116, 83, 32, 200, 7, 116, 133, 153, 178, 250, 54, 237, 148, 103, 50, 110, 16, 15, 131, 194, 232, 119, 169, 120, 55, 221, 106, 81>>}], ttl: 0, type: Aecore.Account.Tx.SpendTx}, signatures: [<<193, 42, 204, 148, 167, 107, 166, 162, 186, 59, 102, 23, 157, 147, 139, 124, 194, 166, 22, 71, 73, 100, 1, 14, 61, 207, 84, 21, 228, 140, 119, 213, 42, 2, 119, 54, 194, 208, 117, 254, 69, 215, 97, 239, 252, 61, 183, ...>>]}, #PID<0.559.0>)
    (aecore) lib/aecore/peers/sync.ex:485: Aecore.Peers.Sync.process_jobs/1
    (aecore) lib/aecore/peers/jobs.ex:81: Aecore.Peers.Jobs.handle_cast/2
    (stdlib) gen_server.erl:616: :gen_server.try_dispatch/4
    (stdlib) gen_server.erl:686: :gen_server.handle_msg/6
Last message: {:"$gen_cast", {:dequeue, :sync_jobs}}
State: %{queues: [sync_jobs: 1]}
```

rdimitrovbg:
Resolves #524 resolves #560 resolves #451


apps/aecore/lib/aecore/sync/chain.ex
```diff
@@ -0,0 +1,141 @@
...
+  end
+
+  @spec merge(t(), t(), list()) :: t()
+  defp merge([], [], acc) do
+    acc
+    |> Enum.sort()
+    |> Enum.reverse()
+  end
+
+  defp merge([], [e2 | chain_2], acc) do
```
thepiwo: can you give the e2 a semantic name?

apps/aecore/lib/aecore/sync/chain.ex
```diff
@@ -0,0 +1,141 @@
...
+    merge([], chain_2, [e2 | acc])
+  end
+
+  defp merge([e1 | chain_1], chain_2, acc) do
```
thepiwo: can you give the e1 a semantic name?

thepiwo: can you add some documentation to the merge function?

apps/aecore/lib/aecore/sync/chain.ex
```diff
@@ -0,0 +1,141 @@
...
+    case Enum.member?(chain_2, e1) do
+      true ->
+        new_chain_2 = Enum.filter(chain_2, fn elem -> elem != e1 end)
```
thepiwo: can you give elem a semantic name?

apps/aecore/lib/aecore/sync/chain.ex
```diff
@@ -0,0 +1,141 @@
...
+
+  def find_hash_at_height(height, [_ | chain]), do: find_hash_at_height(height, chain)
+
+  ## If there is a task with chain_id equal to the given chain,
+  ## merge the data between the chain in the task and the given chain
+  @spec add_chain_info(t(), Sync.t()) :: Sync.t()
+  def add_chain_info(%Chain{chain_id: chain_id} = chain, sync) do
+    case Task.get_sync_task(chain_id, sync) do
+      {:ok, st = %Task{chain: chain_1}} ->
+        st1 = struct(st, chain: merge_chains(chain, chain_1))
```
thepiwo: can you give st1 a semantic name?

apps/aecore/lib/aecore/sync/sync.ex
```diff
@@ -0,0 +1,756 @@
...
+    Jobs.init_queues()
+    Process.flag(:trap_exit, true)
+    {:ok, state}
+  end
+
+  def handle_call({:sync_in_progress, peer_id}, _from, state) do
+    {:reply, peer_in_sync?(state, peer_id), state}
+  end
+
+  def handle_call({:known_chain, %Chain{chain_id: cid} = chain, new_chain_info}, _from, state) do
```
thepiwo: can you give the cid a semantic name?

apps/aecore/lib/aecore/sync/sync.ex
```diff
@@ -0,0 +1,756 @@
...
+    {new_chain, state1} =
```
thepiwo: can you give the state1 a semantic name?

apps/aecore/lib/aecore/sync/sync.ex
```diff
@@ -0,0 +1,756 @@
...
+      case new_chain_info do
+        :none ->
+          {chain, state}
+
+        %Chain{chain_id: c} when c == cid ->
```
thepiwo: can you give the c a semantic name?

apps/aecore/lib/aecore/sync/sync.ex
```diff
@@ -0,0 +1,756 @@
...
+
+        %Chain{chain_id: _cid} ->
+          {chain, Chain.add_chain_info(new_chain_info, state)}
+      end
+
+    {res, new_state} = sync_task_for_chain(new_chain, state1)
+    {:reply, res, new_state}
+  end
+
+  def handle_call({:update_sync_task, update, st_id}, _from, state) do
```
thepiwo: can you give the st_id a semantic name?

apps/aecore/lib/aecore/sync/sync.ex
```diff
@@ -0,0 +1,756 @@
...
+  end
+
+  @spec sync_task_for_chain(Chain.t(), t()) ::
+          {:inconclusive, Chain.t(), {:get_header, chain_id(), peer_id(), height()}}
+          | {{:existing, Task.task_id()}, t()}
+          | {{:new, Chain.t(), Task.task_id()}, t()}
+  def sync_task_for_chain(chain, %Sync{sync_tasks: sts} = state) do
+    case Task.match_tasks(chain, sts, []) do
+      :no_match ->
+        st = %Task{id: st_id} = Task.init_sync_task(chain)
```
thepiwo: can you give the st a semantic name?

apps/aecore/lib/aecore/sync/sync.ex
```diff
@@ -0,0 +1,756 @@
...
+    {{:get_block, pick_height, pick_hash}, st}
+  end
+
+  def get_next_work_item(%Task{} = st) do
+    Logger.info("#{__MODULE__}: Take a break, nothing to do: #{inspect(st)}")
+    {:take_a_break, st}
+  end
+
+  def do_handle_worker({:new_worker, peer_id, pid}, %Task{workers: ws} = st) do
+    case Enum.filter(ws, fn {p_id, _} -> p_id == peer_id end) do
```
thepiwo: can you give the ws, st, p_id a semantic name?

apps/aecore/lib/aecore/sync/sync.ex
```diff
@@ -0,0 +1,756 @@
...
+        |> do_terminate_worker(st)
+        |> Task.set_sync_task(state)
+
+      [] ->
+        state
+    end
+  end
+
+  def do_terminate_worker(pid, %Task{workers: ws} = st) do
+    [{peer, _}] = Enum.filter(ws, fn {_, p} -> p == pid end)
```
thepiwo: can you give the ws, p, pid a semantic name?

apps/aecore/lib/aecore/sync/sync.ex
```diff
@@ -0,0 +1,756 @@
...
+
+  def do_start_sync(peer_id, remote_hash) do
+    if sync_in_progress?(peer_id) do
+      Logger.info("#{__MODULE__}: Already syncing with #{inspect(peer_id)}")
+    else
+      do_start_sync1(peer_id, remote_hash)
+    end
+  end
+
+  defp do_start_sync1(peer_id, remote_hash) do
```
thepiwo: can you give the do_start_sync1 a semantic name?

apps/aecore/lib/aecore/sync/sync.ex
```diff
@@ -0,0 +1,756 @@
...
+
+  defp identify_chain({:new, %Chain{chain: [target | _]}, task}) do
+    Logger.info(
+      "#{__MODULE__}: Starting new sync task #{inspect(task)}, target is: #{inspect(target)}"
+    )
+
+    {:ok, task}
+  end
+
+  defp identify_chain({:inconclusive, chain, {:get_header, ch_id, peers, n}}) do
```
thepiwo: can you give the ch_id, n a semantic name?

apps/aecore/lib/aecore/sync/sync.ex
```diff
@@ -0,0 +1,756 @@
...
+    {:ok, min, agreed_hash}
+  end
+
+  ## Ping logic makes sure they always agree on genesis header (height 0)
+  ## We look for the block that is bloth on remote highest chain and in our local
+  ## chain, connected to genesis (may be on a fork, but that fork
+  ## has now more dofficulty than our highest chain (otherwise we would not sync))
+  ##
+  ## agreed_hash is hash at height min (genesis hash)
+  defp agree_on_height(peer_id, rhash, rh, lh, max, min, agreed_hash) when rh == lh do
```
thepiwo: can you give the rhash, rh, lh, max, min a semantic name?

apps/aecore/lib/aecore/sync/sync.ex
```diff
@@ -0,0 +1,756 @@
...
+        err
+    end
+  end
+
+  def do_forward_block(block, peer_id) do
+    PeerConnection.send_new_block(block, peer_id)
+  end
+
+  def do_forward_tx(tx, peer_id) do
+    PeerConnection.send_new_tx(peer_id, tx)
```
0legIv: Does this function exist `PeerConnection.send_new_tx`? I tried to test it running separate nodes and the same error appears:
https://github.com/aeternity/elixir-node/issues/560

apps/aecore/lib/aecore/sync/chain.ex
```diff
@@ -0,0 +1,141 @@
...
+
+  @type t :: %Chain{chain_id: chain_id(), peers: list(), chain: list(chain())}
+
+  defstruct chain_id: nil,
+            peers: [],
+            chain: []
+
+  @spec init_chain(peer_id(), Header.t()) :: t()
+  def init_chain(peer_id, header) do
+    init_chain(Kernel.make_ref(), [peer_id], header)
```
d-velev: is Kernel added not to confuse these functions for local ones?

apps/aecore/lib/aecore/sync/sync.ex
```diff
@@ -0,0 +1,756 @@
...
+    {{:fill_pool, last_hash, target_hash}, st}
+  end
+
+  def get_next_work_item(
+        %Task{pool: [{_, _, {_, _}} | _] = pool, adding: add, pending: pend} = st
+      ) do
+    {to_be_added, new_pool} = split_pool(pool)
+
+    cond do
+      add === [] ->
```
d-velev: why === instead of ==

apps/aecore/lib/aecore/sync/sync.ex
gspasov: Yes, now we saw that the function is no longer available since PR #557. We will fix this now.

apps/aecore/lib/aecore/sync/chain.ex
gspasov: Yes

gspasov:
In most of the cases where we add element to the end of the list it cannot be changed to a prepend. Or even if it is possible it will make everything much more confusing. In any case, most of the times this is done, we are not adding elem to such a big list, so that it has a big impact on execution time.

apps/aecore/lib/aecore/sync/sync.ex
0legIv: And one little remark - the arguments are not in the right order, should be PeerConnection.send_new_tx(tx, peer_id)

thepiwo:
@0legIv @rdimitrovbg @gspasov so this resolves #560 as well?

thepiwo:
@0legIv @rdimitrovbg @gspasov I added it to the PR description as "resolves"

apps/aecore/lib/aecore/sync/sync.ex
```diff
@@ -130,24 +130,24 @@ defmodule Aecore.Sync.Sync do
...
+    new_state = Task.do_update_sync_task(state, task_id, {:error, peer_id})
     {:reply, :abort_work, new_state}
   end
 
-  def handle_call({:next_work_item, st_id, peer_id, last_result}, _from, state) do
-    state1 = handle_last_result(state, st_id, last_result)
-    {reply, state2} = get_next_work_item(state1, st_id, peer_id)
-    {:reply, reply, state2}
+  def handle_call({:next_work_item, task_id, peer_id, last_result}, _from, state) do
+    updated_state = handle_last_result(state, task_id, last_result)
```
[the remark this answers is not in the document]
gspasov: This cannot be done. We don't know at a given time what `handle_last_result()` will return, thus we cannot say specifically how the state has been updated.

apps/aecore/lib/aecore/sync/sync.ex
```diff
@@ -130,24 +130,24 @@ defmodule Aecore.Sync.Sync do
...
+    {reply, new_state} = get_next_work_item(updated_state, task_id, peer_id)
```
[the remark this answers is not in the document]
gspasov: Same goes here, `get_next_work_item()` could return multiple types of reply, so we can't say how the state has been changed.

apps/aecore/lib/aecore/sync/sync.ex
```diff
@@ -407,26 +413,26 @@ defmodule Aecore.Sync.Sync do
...
         state
     end
   end
 
-  def do_terminate_worker(pid, %Task{workers: ws} = st) do
-    [{peer, _}] = Enum.filter(ws, fn {_, p} -> p == pid end)
-    Logger.info("#{__MODULE__}: Terminating worker: #{pid} for worker: #{peer}")
-    %Task{st | workers: Enum.filter(ws, fn {_, p} -> p != pid end)}
+  def do_terminate_worker(worker_pid, %Task{workers: workers} = task) do
+    [{peer, _}] = Enum.filter(workers, fn {_, pid} -> pid == worker_pid end)
```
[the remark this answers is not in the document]
gspasov: pid is ment for process id

apps/aecore/lib/aecore/sync/sync.ex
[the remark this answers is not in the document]
thepiwo: but this could be described with something like state_applied_last_result or similar

apps/aecore/lib/aecore/sync/sync.ex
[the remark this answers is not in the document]
thepiwo: `state_applied_next_work_item` could describe this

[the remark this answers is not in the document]
thepiwo: as this does actually describe what was changed to the item even if you don't know whats inside your next work item

apps/aecore/lib/aecore/sync/sync.ex
```diff
@@ -457,11 +463,11 @@ defmodule Aecore.Sync.Sync do
     if sync_in_progress?(peer_id) do
       Logger.info("#{__MODULE__}: Already syncing with #{inspect(peer_id)}")
     else
-      do_start_sync1(peer_id, remote_hash)
+      do_start_sync_internal(peer_id, remote_hash)
     end
   end
 
-  defp do_start_sync1(peer_id, remote_hash) do
+  defp do_start_sync_internal(peer_id, remote_hash) do
```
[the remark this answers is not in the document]
thepiwo: the naming problem is in  `do_start_sync`, as it eventually doesn't start sync. If you fix the naming of this to e.g. `ensure_sync_running` or something that indicates that it would not start again if already running there is no need for `_internal`

thepiwo:
@gspasov @rdimitrovbg is there a reasoning why you did choose a `do_` semantic for function names?

thepiwo:
@gspasov could we stick to what we use right now, to be consistent in the project? for major code style changes we should probably discuss with the whole team

apps/aecore/lib/aecore/sync/chain.ex
```diff
@@ -131,18 +131,18 @@ defmodule Aecore.Sync.Chain do
     |> Enum.reverse()
   end
 
-  defp merge([], [elem2 | chain_2], acc) do
-    merge([], chain_2, [elem2 | acc])
+  defp merge([], [chain2 | chain_list_2], acc) do
```
thepiwo: maybe I don't understand what this function does, could you explain what a chain2 element is?

gspasov: Chain1 and Chain2 are both the same type of element, with this function we just merge both chains to one chain. We get all the different elements from chain1 and chain2 and add the elements to the acc.

gspasov:
@thepiwo This does not change any major code style. This is just something that is used in this particular case we have. And there haven't been other function with same name before in the project to use this.

apps/aecore/lib/aecore/sync/chain.ex
thepiwo: That does not answer what an element is

thepiwo:
@gspasov ok, keep it like this, but we should look for other instances in the code where this style should be applied or sub-functions should be private

thepiwo:
If the miner is running in elixir, mining a spendtx from elixir, epoch will not accept this block

apps/aecore/lib/aecore/sync/chain.ex
```diff
@@ -0,0 +1,148 @@
...
+  Implements all the functions regarding the Chain structure of the SyncTask
+  """
+
+  alias Aecore.Chain.Header
+  alias Aecore.Chain.BlockValidation
+  alias Aecore.Sync.Task
+  alias __MODULE__
+
+  @type peer_id :: pid()
+  @type chain_id :: reference()
```
thepiwo: what is `reference`?

apps/aecore/lib/aecore/sync/chain.ex
```diff
@@ -0,0 +1,148 @@
...
+  @type height :: non_neg_integer()
+  @type hash :: binary()
```
thepiwo: what hash is it? don't we have more specific typespecs for block_hash, tx_hash or whatever this should be somewhere?

apps/aecore/lib/aecore/sync/chain.ex
```diff
@@ -0,0 +1,148 @@
...
+  @type chain :: %{height: height(), hash: hash()}
+
+  @type t :: %Chain{chain_id: chain_id(), peers: list(), chain: list(chain())}
```
thepiwo: peers is list() of what type?

apps/aecore/lib/aecore/sync/chain.ex
gspasov: [Basic elixir types](https://hexdocs.pm/elixir/typespecs.html#basic-types)

thepiwo: the link does not explain this well

gspasov: This is used as unique identifier. Elixir has such type, named reference(), and we could simply make one using `Kernel.make_ref()`.

thepiwo: what does data inside look

gspasov: The data is it's own type - reference -> `#Reference<0.53780266.3535798273.175449>`

thepiwo: Okay, I'll have to do more research myself

apps/aecore/lib/aecore/peers/peer_connection.ex
```diff
@@ -148,6 +150,13 @@ defmodule Aecore.Peers.PeerConnection do
     |> send_msg_no_response(pid)
   end
 
+  @spec send_new_tx(SignedTx.t(), pid()) :: :ok | :error
+  def send_new_tx(tx, pid) when is_pid(pid) do
```
grzegorz225: Add pattern matching - only SignedTx should be allowed here:
```
def send_new_tx(%SignedTx{} = tx, pid) when is_pid(pid) do
...
def send_new_tx(_, _), do: :error
```

apps/aecore/lib/aecore/sync/chain.ex
```diff
@@ -0,0 +1,130 @@
...
+  @spec merge_chains(Chain.t(), Chain.t()) :: Chain.t()
+  def merge_chains(%Chain{chain_id: chain_id, peers: peers_1, chain: chain_1}, %Chain{
+        chain_id: chain_id,
+        peers: peers_2,
+        chain: chain_2
+      }) do
+    peers =
+      (peers_1 ++ peers_2)
+      |> Enum.sort()
+      |> Enum.uniq()
```
grzegorz225: The peers are already sorted -> We can use this to merge and deduplicate the peers in O(n) instead of O(n log n)

apps/aecore/lib/aecore/sync/chain.ex
```diff
@@ -0,0 +1,130 @@
...
+        Task.set_sync_task(task_with_merged_chain, sync)
+
+      {:error, :not_found} ->
+        sync
+    end
+  end
+
+  @doc """
+  Get the next known header_hash at a height bigger than N; or
+  if no such hash exist, the header_hash at the highest known height.
```
grzegorz225: There are public functions in this module which miss documentation.

apps/aecore/lib/aecore/sync/sync.ex
```diff
@@ -0,0 +1,843 @@
...
+  end
+
+  def handle_last_result(%Task{} = task, {:post_blocks, :ok}), do: %Task{task | adding: []}
+
+  def handle_last_result(
+        %Task{adding: adds, pending: pends, pool: pool, chain: chain} = task,
+        {:post_blocks, {:error, block_from_peer_id, block_height}}
+      ) do
+    ## Put back the blocks we did not manage to post, and schedule
+    ## failing block for another retraival
```
grzegorz225: Fix typo in comment

apps/aecore/lib/aecore/sync/chain.ex
gspasov: If the two variables of peers_1 and peers_2 are sorted when concatenating them they won't be anymore: `[2, 3] ++ [1, 4] = [2, 3, 1, 4]`. So I don't really understand your point here.

apps/aecore/lib/aecore/peers/peer_connection.ex
```diff
@@ -489,8 +501,13 @@ defmodule Aecore.Peers.PeerConnection do
 
   def rlp_decode(@get_header_by_height, encoded_get_header_by_height) do
     # vsn should be addititonaly decoded with :binary.decode_unsigned
-    [_vsn, height] = ExRLP.decode(encoded_get_header_by_height)
-    %{height: :binary.decode_unsigned(height)}
+    [
+      <<@get_header_by_height_version>>,
+      height,
+      top_hash
+    ] = ExRLP.decode(encoded_get_header_by_height)
```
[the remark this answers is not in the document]
gspasov: Yes, this should be fixed in a different PR, because it's not specifically related to Sync.

apps/aecore/lib/aecore/sync/sync.ex
```diff
@@ -0,0 +1,843 @@
...
+  defp agree_on_height(_peer_id, _r_header_hash, _rheight, _lheight, min, min, agreed_header_hash) do
+    {:ok, min, agreed_header_hash}
+  end
+
+  ## Ping logic makes sure they always agree on genesis header (height 0)
+  ## We look for the block that is both on remote highest chain and in our local
+  ## chain, connected to genesis (may be on a fork, but that fork
+  ## has now more difficulty than our highest chain (otherwise we would not sync))
+  ## agreed_hash is hash at height min (genesis hash)
+  defp agree_on_height(
```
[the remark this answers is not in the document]
gspasov: It is only one version of this function. The different bodies of it are because of the guards and pattern-matching. Maybe it is confusing because the explanation of the function is on top of the second body, will move it to the top one.

apps/aecore/lib/aecore/sync/sync.ex
```diff
@@ -0,0 +1,843 @@
...
+  defp fill_pool(peer_id, start_header_hash, target_header_hash, task_id) do
+    case PeerConnection.get_n_successors(
+           start_header_hash,
+           target_header_hash,
+           @max_headers_per_chunk,
+           peer_id
+         ) do
+      {:ok, %{hashes: []}} ->
+        update_sync_task({:done, peer_id}, task_id)
+        Logger.info("#{__MODULE__}: Sync done (according to #{inspect(peer_id)})")
```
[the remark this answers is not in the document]
gspasov: I think the return value is not important here, but I'll reverse them.

apps/aecore/lib/aecore/peers/peer_connection.ex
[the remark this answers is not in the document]
thepiwo: can you propose a task to adjust this?

apps/aeutil/lib/list.ex
```diff
@@ -0,0 +1,35 @@
...
+    |> Enum.sort()
+    |> Enum.reverse()
+  end
+
+  defp merge_descending([], [head2 | list_2], acc) do
+    merge_descending([], list_2, [head2 | acc])
+  end
+
+  defp merge_descending([head1 | list_1], list_2, acc) do
+    case Enum.member?(list_2, head1) do
```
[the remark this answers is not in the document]
gspasov: If I do it this way, we lose the generic part of the func, because the head in our case is a map `%{height: height, hash: header_hash}`. Yes, I could match it, but then no other two list will match this scenario. So maybe I should move it to the `Aecore.Sync.Chain` module again, if I'll do it this way.

apps/aecore/lib/aecore/sync/sync.ex
```diff
@@ -0,0 +1,843 @@
...
+
+        do_work_on_sync_task(peer_id, task_id, res)
+
+      :abort_work ->
+        Logger.info("#{__MODULE__}: #{self()} Aborting sync work against #{inspect(peer_id)}")
+        schedule_ping(peer_id)
+    end
+  end
+
+  defp post_blocks([]), do: :ok
```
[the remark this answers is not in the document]
gspasov: If you are referring to the `Sync` module, I don't see such functions?

apps/aecore/lib/aecore/peers/peer_connection.ex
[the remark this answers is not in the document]
cytadela8: Why in different PR? I think we could just check here if `ExRLP.decode(encoded_get_header_by_height)` matches `[<<@get_header_by_height_version>>, height, top_hash]`.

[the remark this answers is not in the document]
gspasov: Because as you said, there are multiple cases where this should be fixed.

apps/aecore/lib/aecore/sync/sync.ex
```diff
@@ -0,0 +1,843 @@
...
+  @spec sync_in_progress?(peer_id) :: true | false
+  def sync_in_progress?(peer_id) do
+    GenServer.call(__MODULE__, {:sync_in_progress, peer_id})
+  end
+
+  @spec known_chain(Chain.t(), Chain.t() | :none) ::
+          {:inconclusive, Chain.t()}
+          | {:existing, Task.task_id()}
+          | {:new, Chain.t(), Task.task_id()}
+  def known_chain(chain, extra_info) do
```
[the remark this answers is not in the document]
thepiwo: @gspasov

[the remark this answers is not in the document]
gspasov: I added [doc](https://github.com/aeternity/elixir-node/pull/567/commits/3c2ad1efe5c8030602c3009ba142eb92abcc8f2e#diff-fc5e317d2d5645e116620d8c1f2de703R61)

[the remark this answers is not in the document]
thepiwo: ah okay, in this case the comment can be marked as resolved in github

apps/aeutil/lib/list.ex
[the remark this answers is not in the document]
cytadela8: But `Enum.sort` that is currently used also uses the `<` ordering (I think), so there should be no difference between the current code and my proposal.

apps/aecore/lib/aecore/sync/sync.ex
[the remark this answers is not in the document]
cytadela8: Just the execution flow between the two bodies is a little bit confusing for me, but it's ok to leave it as is.

apps/aecore/lib/aecore/sync/chain.ex
grzegorz225: No, you don't need to concatenate the list - all you need to do is to perform the merging step from the classic merge sort algorithm and add some more checks to make sure that while merging you aren't including duplicates. Here is my code(with tests) for performing this operation in linear time: https://gist.github.com/grzegorz225/431bf22243f032843d13c0ab03a2396d
On my computer it runs 3-4 times faster for random lists containing 10k elements than the naive algorithm.

gspasov: Okay, but as far as I know, we are not going to have more than 30 peers connected to us. So this operation will actually happen for 30 elements. I don't think that it makes much sense to change it for such a little speed change.

gspasov: I don't know if there is much reason in sorting them tbo. But have in mind that most of the times this elements will not be more than 10-20. We can only connect to 30 peers AFAIK.

thepiwo:
@gspasov @rdimitrovbg @grzegorz225 whats the status here?

apps/aeutil/lib/list.ex
```diff
@@ -3,41 +3,45 @@ defmodule Aeutil.List do
...
   """
 
-  @doc """
-  Merge two list in descending order
-  """
-  @spec merge_descending(list(), list()) :: list()
-  def merge_descending(list1, list2) do
-    merge(list1, list2, [])
+  @spec merge(list(), list()) :: list()
+  def merge(l1, l2) do
```
grzegorz225: Change to a more semantic correct name to avoid confusion with the ++ operator and the merge_descending function for instance: merge_sorted_lists_without_duplicates. Add docs saying that the resulting list is sorted and the inputs need to be sorted as well.

apps/aecore/lib/aecore/sync/chain.ex
```diff
@@ -127,4 +124,39 @@ defmodule Aecore.Sync.Chain do
 
     header_hash
   end
+
+  @doc """
+  Merge two chains in descending order
+  """
+  @spec merge_descending(list(chain()), list(chain())) :: list(chain())
+  def merge_descending(list1, list2) do
```
grzegorz225: Change to a more semantic correct name, for instance: merge_descending_chains. This function does not work for general lists and is only used in this module, maybe this should be a private function? You can generalize this operation and add it to the ListUtils module by adding an additional argument which decodes the items in the list. For instance:
def merge_descending(list1, list2, getter_fun \\ &(&1)) do
merge(list1, list2, [], getter_fun)
end
...
  defp merge(
         [hd1 | rest1] = list1,
         [hd2 | rest2] = list2,
         acc, getter_fun
       ) do
    height1 = getter_fun(hd1)
    height2 = getter_fun(hd2)
    ...

Then instead of merge_descending(chain_1, chain_2) you could write:
ListUtils.merge_descending(chain_1, chain_2, fn %{height: height} -> height end)

apps/aeutil/lib/list.ex
gspasov: When I gave it a bit more thought I decided to return the old functionality for few reasons:
1. The `peers` elements are not sorted to begin with, there is no functionality in chain of sync that does so;
2. When a `Chain` is initialized, we only have in the list of peers one peer_id, which is the peer from where the chain is coming;
3. When we do call `merge_chains`, what we do is take our already familiar chain and try matching it to an incoming one, if they do match, we merge them, so when we are merging the peers we are actually adding to our familiar chain just another peer;

For example if our familiar chain has already merged a few times and in the peers it has 3 peer_ids [100, 111, 1002], when we are merging to incoming chain what happens is this -> [100, 111, 1002] ++ [440]. Basically we are adding one element to the list, and we sort this list, because it's needed. For me this makes the merging function unneeded and only over-complicates the logic.

apps/aecore/lib/aecore/sync/chain.ex
gspasov: I don't see why I'd like this to be generic, when it is used only in the Chain module, will make it private though.

gspasov:
Sync is ready for merging

Issue #568, opened 2018-08-30
SignedTx is not compatible with epoch


grzegorz225:
A signed Transaction in the serialized form must have sorted signatures.
https://github.com/aeternity/epoch/blob/v0.16.0/apps/aetx/src/aetx_sign.erl#L133

Our signature verification currently assumes that the signatures have a 1-1 correspondence to the senders pubkey's. This will be resolved together with #533.

gspasov:
Isn't this issue resolved with PR #572 ? If so it should be closed.

thepiwo:
@grzegorz225 ?

PR #572, opened 2018-08-31
[GH-571] Adjusted calculate_txs_hash logic

### LINKED ISSUE #571: epoch does not accept blocks from elixir with spendtx included
If the miner is running in elixir, mining a spendtx from elixir, epoch will not accept this block

epoch reports Failed to add synced block 3: malformed_txs_hash

tested based on #524 code

DanielaIvanova:
resolves #571

### DIFF
--- a/apps/aecore/lib/aecore/chain/block_validation.ex
+++ b/apps/aecore/lib/aecore/chain/block_validation.ex
@@ -6,13 +6,13 @@ defmodule Aecore.Chain.BlockValidation do
   alias Aecore.Pow.Cuckoo
   alias Aecore.Chain.{Block, Header, Genesis}
   alias Aecore.Tx.SignedTx
-  alias Aecore.Tx.DataTx
   alias Aecore.Chain.Chainstate
   alias Aecore.Chain.Target
   alias Aeutil.Hash
   alias Aecore.Chain.Chainstate
   alias Aecore.Governance.GovernanceConstants
   alias Aeutil.PatriciaMerkleTree
+  alias Aeutil.Serialization
 
   @type tree :: :gb_merkle_trees.tree()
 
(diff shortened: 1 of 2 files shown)


apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -134,13 +134,16 @@ defmodule Aecore.Chain.BlockValidation do
...
-    Enum.reduce(txs, PatriciaMerkleTree.new(:txs), fn tx, trie ->
-      encoded_tx = tx.data |> DataTx.rlp_encode()
-      PatriciaMerkleTree.enter(trie, encoded_tx |> Hash.hash(), encoded_tx)
-    end)
+    build_merkle_tree(txs, 0, PatriciaMerkleTree.new(:txs))
+  end
+
+  defp build_merkle_tree([], _position, tree), do: tree
+
+  defp build_merkle_tree([signed_tx | list_txs], position, tree) do
```
grzegorz225: We should pattern match to make sure that the tree only contains SignedTx. For instance:
```
defp build_merkle_tree([%SignedTx{} = signed_tx, | list_txs], position, tree) do
```

apps/aecore/lib/aecore/tx/signed_tx.ex
```diff
@@ -253,7 +253,7 @@ defmodule Aecore.Tx.SignedTx do
   def encode_to_list(%SignedTx{} = tx) do
     [
       :binary.encode_unsigned(@version),
-      tx.signatures,
+      Enum.sort(tx.signatures),
```
grzegorz225: This change will break our signature verification code as our code currently assumes that the signatures are not sorted and have a 1-1 correspondence to sender's pubkeys.
If we serialize and then deserialize the SignedTx then the signature verification will fail when multiple signatures are present. The signatures also need to be sorted for the JSON api.
Here is the code to resolve issue #568 (the code misses typespecs and docs - I will add them soon): https://github.com/aeternity/elixir-node/commit/99a2f367b4254359a095f1dd86c9ebe1278ee3a0

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -134,13 +134,16 @@ defmodule Aecore.Chain.BlockValidation do
...
-
   def build_merkle_tree(txs) do
-    Enum.reduce(txs, PatriciaMerkleTree.new(:txs), fn tx, trie ->
-      encoded_tx = tx.data |> DataTx.rlp_encode()
-      PatriciaMerkleTree.enter(trie, encoded_tx |> Hash.hash(), encoded_tx)
-    end)
+    build_merkle_tree(txs, 0, PatriciaMerkleTree.new(:txs))
+  end
+
+  defp build_merkle_tree([], _position, tree), do: tree
```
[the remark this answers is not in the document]
DanielaIvanova: The function is private.

apps/aecore/lib/aecore/tx/signed_tx.ex
DanielaIvanova: We will remove sorting function and we will leave everything here as it was. So then you can resolve it in your PR.

PR #586, opened 2018-09-05
[GH-533] Channel adjustments

### LINKED ISSUE #492: Channel transactions
Change from current simple model of off-chain state handling to transaction model.

### LINKED ISSUE #533: ChannelStateOffChain serialization
Epoch has a very complicated ChannelStateOffChain serialization for transactions. Target of this issue is understanding what's exactly going on and implementing it in our code.

### LINKED ISSUE #568: SignedTx is not compatible with epoch

grzegorz225:
This PR resolves #533, resolves #492, resolves #568.


apps/aecore/lib/aecore/tx/signed_tx.ex
```diff
@@ -220,48 +197,99 @@ defmodule Aecore.Tx.SignedTx do
...
+        {:ok, _} ->
+          true
+        :error ->
+          false
+      end
     end
   end
 
+  @spec many_signatures_check(list(binary()), binary(), list(Keys.pubkey())) :: boolean()
+  defp many_signatures_check(signatures, data_binary, [pubkey | remaining_pubkeys]) do
```
d-velev: check_multiple_signatures sounds better

apps/aecore/lib/aecore/channel/channel_offchain_update.ex
```diff
@@ -0,0 +1,128 @@
+defmodule Aecore.Channel.ChannelOffchainUpdate do
```
cytadela8: To follow existing patterns please rename to OffChain (capital C) and change filename accordingly (`git mv`)

cytadela8: This applies also to other files.

apps/aecore/lib/aecore/channel/channel_state_on_chain.ex
```diff
@@ -45,23 +50,27 @@ defmodule Aecore.Channel.ChannelStateOnChain do
     :responder_amount,
     :lock_period,
     :slash_close,
-    :slash_sequence
+    :slash_sequence,
+    :state_hash,
+    :channel_reserve
   ]
 
   use ExConstructor
```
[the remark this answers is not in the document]
cytadela8: This module defines a structure (has `defstruct`). Isn't this needed for all such modules?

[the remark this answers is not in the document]
d-velev: exconstructor just gives you a new() function which takes a map with the arguments and creates a structure, it's a library

apps/aecore/lib/aecore/channel/channel_transaction.ex
```diff
@@ -0,0 +1,212 @@
...
+  alias Aecore.Tx.DataTx
+  alias Aecore.Channel.ChannelOffchainTx
+
+  @typedoc """
+  Data structures capable of mutating the offchain chainstate off an state channel
+  """
+  @type channel_tx ::
+          Aecore.Channel.ChannelOffchainTx
+          | Aecore.Channel.Tx.ChannelCreateTx
+          #| Aecore.Channel.Tx.ChannelWidhdrawTx
```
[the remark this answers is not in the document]
grzegorz225: These are placeholders for the withdraw and deposit transactions which will be implemented in a future PR.

apps/aecore/lib/aecore/channel/updates/channel_deposit_update.ex
```diff
@@ -0,0 +1,85 @@
...
+  Serializes ChannelDepositUpdate. The serialization was changed in later versions of epoch.
+  """
+  @spec encode_to_list(ChannelDepositUpdate.t()) :: list(binary())
+  def encode_to_list(
+        %ChannelDepositUpdate{
+          from: from,
+          amount: amount
+        })
+  do
+    [from, from, amount]
```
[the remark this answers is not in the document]
grzegorz225: This is needed to be compatible with epoch 0.16

In epoch 0.16 the offchain updates are not serializable objects(this was changed in later versions of epoch). In order to serialize the updates epoch 0.16 serializes each update as a tuple: {update_tag(), binary(), binary(), non_neg_integer()}.
One can clearly see that we do not need two binary fields for withdraw and deposit updates but as our serialization format is fixed we need to include the public key twice. Here is the serialization in epoch 0.16: https://github.com/aeternity/epoch/blob/v0.16.0/apps/aechannel/src/aesc_offchain_state.erl#L163

apps/aecore/lib/aecore/poi/poi_proof.ex
```diff
@@ -0,0 +1,231 @@
...
+  def encode_to_list(%PoiProof{root_hash: :empty}) do
+    []
+  end
+
+  def encode_to_list(%PoiProof{root_hash: root_hash, db: db}) do
+    contents =
+      db
+      |> Enum.map(fn {key, value} -> [key, ExRLP.decode(value)] end)
+      |> Enum.sort_by(fn [key, _] -> key end) #the specification requires keys to be sorted in serialized POI's
+    [[root_hash, contents]]
```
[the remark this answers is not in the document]
grzegorz225: Yes, this is the serialization used by epoch 0.16.
I discovered this when I noticed that the aecore_poi_epoch_compatibility_and_functional_test was failing because my code was not able to produce the same serialization as epoch. After I inspected the rlp decoded data it turned out that the proofs in epoch are enclosed in a double list.
I wont resolve this conversation here as I'm not sure why epoch uses a double list here.

apps/aecore/lib/aecore/channel/channel_state_tree.ex
```diff
@@ -55,13 +58,22 @@ defmodule Aecore.Channel.ChannelStateTree do
...
   def root_hash(tree) do
     PatriciaMerkleTree.root_hash(tree)
   end
+
+  @spec normalize_key(id()) :: ChannelStateOnChain.id()
+  def normalize_key(%Identifier{type: :channel, value: key}) do
+      key
+  end
+
+  def normalize_key(key) when is_binary(key) do
```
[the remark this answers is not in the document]
grzegorz225: @thepiwo I don't see a problem here, could you elaborate. The id type is defined as Identifier.t() or ChannelStateOnChain.id(). ChannelStateOnChain.id() is currently defined as binary().

apps/aecore/lib/aecore/channel/channel_transaction.ex
```diff
@@ -0,0 +1,212 @@
...
+    false
+  end
+
+  @doc """
+  Retrieves the sequence of the offchain state after applying the channel transaction to the state channel
+  """
+  @spec get_sequence(signed_tx() | channel_tx()) :: non_neg_integer()
+  def get_sequence(tx) do
+    structure = unpack_tx(tx)
+    structure.__struct__.get_sequence(structure)
```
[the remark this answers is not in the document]
grzegorz225: I removed these getter functions.

apps/aecore/lib/aecore/channel/channel_transaction.ex
```diff
@@ -0,0 +1,212 @@
+defmodule Aecore.Channel.ChannelTransaction do
```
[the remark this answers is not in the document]
grzegorz225: @thepiwo @cytadela8 
I've removed the getters from this module. I've left the setters for now.

apps/aecore/test/aecore_channels_test.exs
```diff
@@ -208,29 +168,160 @@ defmodule AecoreChannelTest do
     assert PatriciaMerkleTree.trie_size(Chain.chain_state().channels) == 0
   end
 
+  @tag :channels
```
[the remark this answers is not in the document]
grzegorz225: Some of the code added here(withdraw and deposit updates) can not be tested right now without the onchain part by @cytadela8 which will be added in a future PR.

grzegorz225:
@thepiwo  @cytadela8 @d-velev 
PR is ready for re-review.

apps/aecore/lib/aecore/channel/channel_off_chain_tx.ex
```diff
@@ -0,0 +1,249 @@
...
+
+  @doc """
+  Decodes the embedded payload of ChannelSoloCloseTx, ChannelSlashTx, ChannelSnapshotTx
+  """
+  @spec decode_from_payload(binary()) :: {:ok, ChannelOffChainTx.t()} | :empty | error()
+  def decode_from_payload(<<>>) do
+    {:ok, :empty}
+  end
+
+  def decode_from_payload([@signed_tx_tag, @version, [sig1, sig2], encoded_tx]) do
```
cytadela8: This has to be reconsidered. There is no function currently to deserialize and serialize a ChannelOffChainTx with signatures.

Such a function should maybe be part of SignedTx as a ChannelOffChainTx with signatures is an ChannelOffChainTx inside SignedTx.

Please make sure it works for following binary: 
```Bits.decode58("G6DYbwbT9jsa4M7dKjQAw1bSwcPN3TcaCKtxBBAk28QaXfEugae4HGVNxSHSrSd59vPCLAssWmX8vN48LWwTypFq2FYy8AVAev68Q5WerLhcsCViNAtgNdzdeGhRMYHBHxYwnPdMRjKThcQAogdqjrtnMKnwqVE6awwDPYPmRjsDT3JW8HY8m43SWB9QeiauG3EZx8KnjejyXuNZNPGVjFiGnjQCAhwVb5nPXuhs4KXgTKjB5w1Huch5wJg3HUzyQXhuoCk4gn7vcQvjX7P4rthpo1CMNfJj6WXFowKpVzUCQrVrTYUMHzvbYuymzh1CdcmfgJouxHQmDSHXJwRW6Uw6uwRY192oKMbhsanuLjRJC9YB8SzSLdrSu828c6QLHnWqPedsHwW")```

**Comment by cytadela8** [comment]
Ready for review! @thepiwo

**Comment by cytadela8** [comment]
`test/aehttpclient_test.exs:13` test succeeds locally, but fails on master due to different order of txs in Pool

**Comment by thepiwo** [comment]
@cytadela8 could you make the test so it doesn't depend on ordering of transactions?

File: apps/aecore/lib/aecore/account/account.ex [hunk 12]
```diff
@@ -18,8 +18,8 @@ defmodule Aecore.Account.Account do
 
   @version 1
 
-  @typedoc "Reason of the error"
-  @type reason :: String.t()
+  @typedoc "Error with reason"
+  @type error :: {:error, binary()}
```
thepiwo: why is the reason not `String.t()`?

apps/aecore/lib/aecore/account/account.ex
```diff
@@ -343,6 +343,20 @@ defmodule Aecore.Account.Account do
     %Account{account_state | balance: new_balance}
   end
 
+  @doc """
+  Adds balance to the given Account state. No exception version.
+  """
+  @spec apply_transfer(Account.t(), non_neg_integer(), integer()) :: {:ok, Account.t()} | error()
```
thepiwo: why is this function added? if this is a variant of another one, we should use one variant for all use-cases

apps/aecore/lib/aecore/account/account.ex
```diff
@@ -343,6 +343,20 @@ defmodule Aecore.Account.Account do
...
+  def apply_transfer(account_state, _block_height, amount) do
+    new_balance = account_state.balance + amount
+
+    if new_balance >= 0 do
+      {:ok, %Account{account_state | balance: new_balance}}
+    else
+      {:error, "#{__MODULE__}: Negative balance"}
```
thepiwo: error message should include some debug data

apps/aecore/lib/aecore/account/account.ex
```diff
@@ -352,6 +366,15 @@ defmodule Aecore.Account.Account do
     %Account{account_state | nonce: new_nonce}
   end
 
+  @spec apply_nonce(Account.t(), integer()) :: {:ok, Account.t()} | error()
+  def apply_nonce(%Account{nonce: current_nonce} = account_state, new_nonce) do
+    if current_nonce < new_nonce do
+      {:ok, %Account{account_state | nonce: new_nonce}}
+    else
+      {:error, "#{__MODULE__}: Invalid nonce"}
```
thepiwo: error message should include some debug data

apps/aecore/lib/aecore/account/account.ex
```diff
@@ -352,6 +366,15 @@ defmodule Aecore.Account.Account do
     %Account{account_state | nonce: new_nonce}
   end
 
+  @spec apply_nonce(Account.t(), integer()) :: {:ok, Account.t()} | error()
```
thepiwo: why is this function added?

apps/aecore/lib/aecore/account/account_state_tree.ex
```diff
@@ -14,6 +14,8 @@ defmodule Aecore.Account.AccountStateTree do
   @typedoc "Hash of the tree"
   @type hash :: binary()
 
+  @type error :: {:error, binary()}
```
thepiwo: why is this binary() not String.t()?

apps/aecore/lib/aecore/account/account_state_tree.ex
```diff
@@ -44,6 +46,23 @@ defmodule Aecore.Account.AccountStateTree do
     put(tree, key, fun.(get(tree, key)))
   end
 
+  @spec safe_update(
```
thepiwo: same question as above, why do we need this function here?

apps/aecore/lib/aecore/account/tx/spend_tx.ex
```diff
@@ -121,10 +119,8 @@ defmodule Aecore.Account.Tx.SpendTx do
         %{},
         block_height,
         %SpendTx{amount: amount, receiver: %Identifier{value: receiver}},
-        %DataTx{} = data_tx
+        %DataTx{senders: [%Identifier{value: sender}]}
```
thepiwo: I assume this fails if multiple senders are present

apps/aecore/lib/aecore/account/tx/spend_tx.ex
```diff
@@ -147,14 +143,11 @@ defmodule Aecore.Account.Tx.SpendTx do
...
-        accounts,
-        %{},
-        _block_height,
-        %SpendTx{amount: amount},
-        %DataTx{fee: fee} = data_tx
-      ) do
-    %Account{balance: balance} = AccountStateTree.get(accounts, DataTx.main_sender(data_tx))
+  def preprocess_check(accounts, %{}, _block_height, %SpendTx{amount: amount}, %DataTx{
+        fee: fee,
+        senders: [%Identifier{value: sender}]
```
thepiwo: I assume this fails if multiple senders are present

apps/aecore/lib/aecore/channel/channel_off_chain_update.ex
```diff
@@ -0,0 +1,164 @@
...
+      :ok
+    end
+  end
+
+  @doc """
+  Runs preprocess checks for an update which was signed by the foreign peer in the channel.
+  """
+  @spec half_signed_preprocess_check(update_types(), map()) :: :ok | error()
+  def half_signed_preprocess_check(update, opts) do
+    module = update.__struct__
```
thepiwo: can you explain this?

apps/aecore/lib/aecore/channel/channel_state_peer.ex
```diff
@@ -46,118 +52,362 @@ defmodule Aecore.Channel.ChannelStatePeer do
...
-      highest_state: state,
-      highest_signed_state: state,
+      channel_id: channel_id,
       channel_reserve: channel_reserve
     }
+
+    Enum.reduce_while(offchain_tx_list_from_oldest, {:ok, initial_state}, fn tx, {:ok, state} ->
+      case process_fully_signed_tx(tx, state) do
+        {:ok, _} = new_acc ->
+          {:cont, new_acc}
```
thepiwo: what is `:cont` maybe use a more semantic name

apps/aecore/lib/aecore/channel/channel_state_peer.ex
```diff
@@ -46,118 +52,362 @@ defmodule Aecore.Channel.ChannelStatePeer do
...
+        err
+    end
+  end
+
+  @spec half_signed_mutual_onchain_tx_from_spec(
+          ChannelStatePeer.t(),
+          non_neg_integer(),
+          non_neg_integer(),
+          module(),
+          map(),
```
thepiwo: use a more expressive type alias (maybe needed for other occurrences as well)

apps/aecore/lib/aecore/channel/channel_state_peer.ex
```diff
@@ -46,118 +52,362 @@ defmodule Aecore.Channel.ChannelStatePeer do
...
+      {:error, _} = err ->
+        err
+    end
+  end
+
+  @spec half_signed_mutual_onchain_tx_from_spec(
+          ChannelStatePeer.t(),
+          non_neg_integer(),
+          non_neg_integer(),
+          module(),
```
thepiwo: what is the module type? maybe use a more semantic type name

thepiwo:
@cytadela8 @grzegorz225 In general this PR looks good, but is really hard to review, as it is so big, next time we should have split this in smaller tasks and separate PRs, not longer than maybe ~500 lines. This would have been good to have a proper estimate of timing as well, as this PR was "will be ready today" for the last 2 weeks.

thepiwo:
Did you check for epoch compatibility in sync?

apps/aecore/lib/aecore/account/tx/spend_tx.ex
cytadela8: Number of senders shuld be checked in `validate` and it's enforced by encoding

apps/aecore/lib/aecore/channel/channel_state_peer.ex
cytadela8: This is for `reduce_while`. `reduce_while` compared to `reduce` stops also when there are no more items to process or supplied function returns `{:halt, value}`

cytadela8:
I checked for epoch compatibility in normal scenario. Now that offchaintx serialization is fixed I will check solo close and slash.

apps/aecore/lib/aecore/account/tx/spend_tx.ex
cytadela8: Number of senders shuld be checked in `validate` and it's enforced by encoding

apps/aecore/lib/aecore/channel/channel_state_peer.ex
thepiwo: ah okay, I didn't know

thepiwo:
Can you report back here, after you checked this?

cytadela8:
>     sometimes use, other times @behaviour is used
I don't see anywhere where it shuld be change. We use behaviour when a module is simply implementing some behaviour. We use `use` when we want to invoke a macro module (Aecore.Util.Serializable is such a module. It adds a `rlp_encode`, `rlp_decode` functions to all Serializable structures modules.)
>     sig, sig1, sig2 could be more expressive
Signatures are sorted (by bytes, not ordered by party) so there isn't much to add to the name. It renamed to `signature1` and `signature2`.
>     inconsistency between binary() and String.t() for error reasoning
Fixed
>     error reasoning often misses debug data
Added more info to errors.
>     remove getter/setter functions
Could you provide some examples? Functions in `Aecore.Channel.ChannelTransaction` provide an abstraction layer over different types of txs and aren't simple getters and setters.

cytadela8:
I left some comments to be explained by @grzegorz225 as they don't have very simple fixes and are regarding to his changes.

thepiwo:
@cytadela8 

> test/aehttpclient_test.exs:13
> est succeeds locally, but fails on master due to different order of txs in Pool


> @cytadela8 could you make the test so it doesn't depend on ordering of transactions?

should we create an issue for this in the overall project as this happens on other branches as well?

cytadela8:
I will create an issue and make a PR with my short solution (94ef096).

apps/aecore/lib/aecore/account/account.ex
grzegorz225: @thepiwo The `apply_transfer!` variant uses exceptions to handle errors and is used in the on chain transaction processing stack. The off chain updates need to change the state of accounts and originally used the version with exceptions(https://github.com/aeternity/elixir-node/pull/586/commits/d8fb2596a57872dabbeae88348bd1750e48fb689#diff-850c2fc5dd4f310c3b6b7fa770e16236L128). Handling exceptions is slow so it was suggested by @cytadela8 to create an exception-less version. 

We can consider to go back to the version which uses exceptions or refactor the onchain transactions so they switch to the exception-less version(this can be a separate issue).

apps/aecore/lib/aecore/account/account.ex
grzegorz225: Same as `apply_transfer`

apps/aecore/lib/aecore/account/account_state_tree.ex
grzegorz225: Same as apply_transfer

apps/aecore/lib/aecore/channel/channel_off_chain_update.ex
grzegorz225: @thepiwo When one party initializes an update of the channel state and sends an signed transaction to the other party, the other party needs to run some checks to determine if the transaction is not malformed or malicious. All incoming transaction are processed by the same logic in channels which currently looks like this:
1) Run checks determining if the updates in the incoming transaction make sense (transfers are in the right direction, transfer of an positive amount of tokens etc). This step works similar to the validate callback in the on chain transaction validation stack.
2) Check metadata such as signatures, sequence of updates, channel id etc.
3) Run the updates
4) Check if the resulting state hash agrees with the one provided in the transaction
5) Accept the transaction by signing it and mutating the state

This function is an callback for performing the step 1) of the workflow. Maybe this function should have a more self explanatory name (for instance: validate_incoming_halfsigned_transaction)?

grzegorz225:
Responded to comments and made small changes.

apps/aecore/lib/aecore/channel/channel_off_chain_update.ex
thepiwo: the "preprocess_check" naming is okay, but what does the update.__struct__ do?

apps/aecore/lib/aecore/account/account.ex
thepiwo: I prefer the exception-less functions over the existing ones, could we use them for all cases? Otherwise we could keep it like it is, but refactor so the exception-less version gets called by the `!` one.

apps/aecore/lib/aecore/channel/channel_off_chain_update.ex
grzegorz225: @thepiwo Each structure in elixir has a special field called ```__struct__  ``` which contains the module defining the structure. This works exactly like the type field in ```DataTx```.

grzegorz225:
@thepiwo @cytadela8 All comments/issues were addressed. Could you take a look at the current state of the PR?

apps/aecore/lib/aecore/channel/channel_off_chain_update.ex
```diff
@@ -116,41 +125,42 @@ defmodule Aecore.Channel.ChannelOffChainUpdate do
...
-  def apply_updates(chainstate, updates, channel_reserve) do
-    Enum.reduce_while(
-      updates,
-      {:ok, chainstate},
-      &apply_single_update_to_chainstate(&1, &2, channel_reserve)
-    )
+  def update_offchain_chainstate(chainstate, object) do
+    module = object.__struct__
+    {:ok, module.update_offchain_chainstate!(chainstate, object)}
+  catch
```
cytadela8: We don't catch programming errors.

thepiwo:
@cytadela8 can you approve now?
@d-velev @DanielaIvanova @Artur64 can you maybe review as well?

thepiwo:
@grzegorz225 @cytadela8 can you resolve the conflicts, so we can merge this before next week?

cytadela8:
Yes

PR #592, opened 2018-09-07
[GH-590] Adjust the Multinode Test Framework to check the pool and mined txs

### LINKED ISSUE #590: Adjust the Multinode Test Framework to check the pool and mined txs

0legIv:
resolves #590

### DIFF
--- a/apps/aecore/test/aecore_channels_test.exs
+++ b/apps/aecore/test/aecore_channels_test.exs
@@ -110,7 +110,7 @@ defmodule AecoreChannelTest do
   end
 
   @tag :channels
-  @tag timeout: 240_000
+  @tag timeout: 340_000
   test "create channel, transfer twice, slash with old, slash with corrent and settle", ctx do
     id = create_channel(ctx)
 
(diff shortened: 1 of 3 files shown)


apps/aecore/test/aecore_channels_test.exs
```diff
@@ -110,7 +110,7 @@ defmodule AecoreChannelTest do
   end
 
   @tag :channels
-  @tag timeout: 240_000
+  @tag timeout: 340_000
```
thepiwo: why do these tests need so much time? (cc @grzegorz225 @cytadela8 )

apps/aetestframework/lib/worker.ex
```diff
@@ -224,6 +224,19 @@ defmodule Aetestframework.Worker do
     send_command(node_name, "Oracle.respond(query_id, \"boolean\", 5)")
   end
 
+  def get_latest_tx_type(node_name) do
+    send_command(node_name, "%Block{txs: txs} = Chain.top_block")
+    send_command(node_name, "[tx] = txs")
+    send_command(node_name, "encoded_type = :erlang.term_to_binary(tx.data.type)")
```
thepiwo: why is term_to_binary and then Base.encode32 used?

0legIv: Because the result that is received from the node is string and Base.encode32 is needed because it encodes term_to_binary result to the string in order to send it back to the node

thepiwo: thy

PR #596, opened 2018-09-07
[GH-591] Make max txs per block be equal to epoch node

### LINKED ISSUE #591: Adjust the limit for transactions in a block
edit: as @gspasov said below, the limit should be changed from 100 to 10946

gspasov:
resolves #591


apps/aecore/lib/aecore/governance/governance_constants.ex
```diff
@@ -29,6 +29,8 @@ defmodule Aecore.Governance.GovernanceConstants do
 
   @name_claim_burned_fee 3
 
+  @max_txs_per_block 10946
```
cytadela8: This is some duplication. Max txs per block should be defined in GovernaneConstants OR in config. Not both.

gspasov: Forgot to delete it.

0legIv:
@thepiwo I removed them just for now, there are some errors in the Multi node test framework now, we are investigating them with @gspasov

thepiwo:
thank you @0legIv

apps/aetestframework/test/multi_node_sync_test.exs
```diff
@@ -39,9 +36,7 @@ defmodule MultiNodeSyncTest do
   test "spend_tx test" do
     TestFramework.mine_sync_block("node1")
     TestFramework.spend_tx("node1")
-    assert TestFramework.get_pool_tx_count("node1") == 1
```
cytadela8: Why do we remove all those asserts?

cytadela8: How are we sure now that the txs get mined?

thepiwo:
@gspasov this is ready, but just waiting for https://github.com/aeternity/elixir-node/pull/617 right?

gspasov:
Yes, since it has some already unwanted changes to the TestFramework

thepiwo:
thank you @gspasov

Issue #352, opened 2018-04-19
refactor serialization to be explicit and in every structure


thepiwo:
every structure that needs to be serialized and deserialized should have it's own functions to do so, included. Additional there should be no default serialization, every field should be handled explicitly.

cytadela8:
(See changes started in PR #355)

PR #606, opened 2018-09-11
[GH-352] Fix merge issue in oracles

cytadela8:
This fixes a merge issue introduced by #352 in oracles.


cytadela8:
I'm not sure actually if it's correct now. I see previously thre was a line: `%DataTx{type: OracleResponseTx} = data -> DataTx.rlp_encode(data)`. Can somebody take a look at this, please?

PR #608, opened 2018-09-11
[GH-584] Improved function specs

### LINKED ISSUE #584: Improve function specs
Go over specs and check if they can be more specific, fill in missing specs for public functions

DanielaIvanova:
resolves #584


DanielaIvanova:
ok, I will check

thepiwo:
@d-velev @gspasov what is your opinion on the `start_link` specs? I don't think they are needed to have.

apps/aecore/lib/aecore/tx/pool/worker.ex
```diff
@@ -18,6 +18,7 @@ defmodule Aecore.Tx.Pool.Worker do
 
   @type tx_pool :: map()
 
+  @spec start_link(any()) :: :ignore | {:error, any()} | {:ok, pid()}
```
thepiwo: this looks better than the huge spec used in the other `start_link`

DanielaIvanova: I can remove all `start_link` specs if you think that is better without them. I really don't like it!

thepiwo: thank you

PR #611, opened 2018-09-11
[GH-610] MerklePatriciaTree bug temporary workaround

### LINKED ISSUE #565: Problem in oracle PMT root_hash in the chain_state
**Steps to reproduce the problem:**
Be sure that the keys are the same in both nodes.
1. Register an oracle in elixir node.
2. Register an oracle with the same data in epoch node.
**Note:** The oracle_tree root_hashes should be same in both nodes.
3. Register a query in elixir node.
4. Register a query in epoch node with the same data.
**Note:** The oracle_tree root_hashes should be same in both nodes.
After that register response in both nodes with the same data and you can see that the root_hashes are still the same in both trees.
The problem is coming when we extend the oracle.
5. Make a transaction to `extend` the oracles in both nodes with same `ttl` and `fee`
At this moment the oracle_tree root_hashes are **different**.
The most interesting thing is that when we inspect the trees all keys and values are equal.

**Useful information**
In epoch we can inspect the tree:
```
aeu_mtrees:to_list(Tree).
```
In elixir we can inspect the tree:
```
Chain.chain_state.oracles.oracle_tree |> PatriciaMerkleTree.print_trie([output: :as_pair, deserialize: false]) 
```

### LINKED ISSUE #610: MerklePatriciaTree workaround
There seems to be a very simple workaround for MerklePatriciaTree bugs.

https://github.com/aeternity/elixir-merkle-patricia-tree/issues/13

cytadela8:
MerklePatriciaTree has a bug that may be complicated to solve, but has easy workaround.

This bug showed itself in https://github.com/aeternity/elixir-node/issues/565. This workaround will allow further work on making our node compatible with epoch till we solve elixir-merkle-patricia-tree issues.

MerklePatriciaTree issues:
- https://github.com/aeternity/elixir-merkle-patricia-tree/issues/10
- https://github.com/aeternity/elixir-merkle-patricia-tree/issues/13

This resolves #565, resolves #610


cytadela8:
If we merge this, I will add an blocked issue: "Remove temporary workaround" so we don't forget.

thepiwo:
@cytadela8 could you explain how this workaround works? as I see, reconstructing the whole tree each time?

cytadela8:
@thepiwo yes, that's what it does. It works, because updating entries in MerklePatriciaTree triggers some bug in the library. We didn't find any problems when each key is inserted only once. I opened an issue on elixir-merkle-patricia-trie repo.

PR #617, opened 2018-09-13
[GH-602] Refactoring Multinode sync test Framework and it's tests

### LINKED ISSUE #602: Handle messages accordingly - Multisync node Test Framework
The way the messages are handles atm in the Multisync Framework, makes the Tests unstable. They could pass or could fail, without any obvious reason. This is due to the asynchronous calls the `Port` makes. Will need to make sure that we are receiving the correct messages until we check for some validity of the received data.

gspasov:
resolves #602


thepiwo:
looks good in general

apps/aetestframework/lib/worker.ex
```diff
@@ -1,599 +1,161 @@
...
-
-  def get_top_block(node) do
-    GenServer.call(__MODULE__, {:get_top_block, node})
-  end
-
-  def get_top_block_hash(node) do
-    GenServer.call(__MODULE__, {:get_top_block_hash, node})
-  end
+  defstruct port_id: nil, node_port: nil, sync_port: nil
+  use ExConstructor
```
thepiwo: is ExConstructor needed here?

apps/aetestframework/lib/worker.ex
```diff
@@ -1,599 +1,161 @@
...
-      node_name,
-      "{:ok, tx} = Account.claim(\"test.aet\", 123, 10)"
-    )
-
-    send_command(node_name, "Pool.add_transaction(tx)")
+  @spec call_with_delay(String.t(), reference(), atom(), atom(), non_neg_integer()) :: any
+  def call_with_delay(cmd, fun, match_by, node, delay) do
+    :timer.sleep(delay)
+    new_timeout = @default_timeout + delay
+    fun.(cmd, match_by, node, new_timeout)
```
thepiwo: nice abstraction 👍

apps/aetestframework/lib/worker.ex
```diff
@@ -1,599 +1,161 @@
...
-      Port.close(val.process_port)
+  def handle_call(:delete_nodes, _from, state) do
+    Enum.each(state, fn {_node, %{port_id: port_id, node_port: port}} ->
+      Port.command(port_id, ":erlang.halt()\n")
+      Port.close(port_id)
       path = String.replace(System.cwd(), ~r/(?<=elixir-node).*$/, "") <> "/apps/aecore/priv/"
-      File.rm_rf(path <> "aewallet_#{val.port}")
-      File.rm_rf(path <> "peerkeys_#{val.port}")
-      File.rm_rf(path <> "rox_db_#{val.port}")
+      File.rm_rf(path <> "aewallet_#{port}")
```
thepiwo: we should move this to explicitly "test" named directories so this doesn't delete data used in development

apps/aetestframework/lib/worker.ex
```diff
@@ -1,599 +1,161 @@
...
-            oracle_interaction_objects: %{}
-          })
-
-        {:reply, :ok, new_state}
-    end
+  @doc """
+  Checking if the port is busy
+  """
+  def busy_port?(port) do
+    :os.cmd('lsof -i -P -n | grep -w #{port}') != []
```
thepiwo: lsof should be added to dependencies, with reference to the test framework

apps/aetestframework/test/multi_node_sync_test.exs
```diff
@@ -37,114 +40,425 @@ defmodule MultiNodeSyncTest do
...
+
+    # Check that node4 has 10 tokens
+    assert node4_pub
+           |> get_balance_cmd()
+           |> TestFramework.call_with_delay(&TestFramework.get/4, :balance, :node4, 1000) == 10
+
+    TestFramework.delete_all_nodes()
+  end
+
+  ## Helper functions
```
thepiwo: I think the helper functions should be moved to a separate file

apps/aetestframework/lib/worker.ex
```diff
@@ -1,599 +1,161 @@
...
-    Sending command in a given node.
-  """
-  @spec send_command(String.t(), String.t()) :: :ok | :unknown_node
-  def send_command(node_name, cmd) do
-    # adding \n to cmd to imitate pressing enter in iex shell
-    cmd = cmd <> "\n"
-    GenServer.call(__MODULE__, {:send_command, node_name, cmd}, 10_000)
+  @spec new_node(atom(), non_neg_integer()) :: :ok
+  def new_node(node_name, iex_num) do
+    GenServer.call(__MODULE__, {:new_node, node_name, iex_num}, 10_000)
```
d-velev: I see the timeout is passed as an argument in other places, can you do the same here or just have it as a constant?

apps/aetestframework/lib/worker.ex
gspasov: It's used to make the node "objects"

apps/aetestframework/lib/worker.ex
gspasov: But when we are getting data for the keys, we are looking into these folders for already created keys. If we move it to somewhere else, it won't read from the correct place and possibly will create new keys in the /priv dir anyways

apps/aetestframework/lib/worker.ex
gspasov: `lsof` is a unix command, what do you mean by adding it to the dependencies?

thepiwo: lsof needs to be installed on the system this is supposed to run on

apps/aetestframework/lib/worker.ex
thepiwo: but somewhere we set these directories to be used, so we can add some "test" string in the path

apps/aetestframework/lib/worker.ex
thepiwo: okay 👍

apps/aetestframework/lib/worker.ex
gspasov: I can add this to the README.md file, that it is required in order for the Multinode sync test Framework to work.

thepiwo: Yes

gspasov:
Made the test re-run themselves if they fail. This is done 5 times. If the check is still not correct we assume there is something wrong. Although there is some kind of problem with NamingTxs. They are not shared correctly between the nodes. (Some nodes take them, some not). Will investigate this.

Also now checking top_block_header_hash of all nodes, not only just 2 picked nodes. (Done for more secure check)

apps/aetestframework/lib/worker.ex
```diff
@@ -1,599 +1,188 @@
...
-      node_name,
-      "{:ok, revoke} = Account.name_revoke(transfer_to_pub, transfer_to_priv, \"test.aet\", 10, next_nonce)"
-    )
-
-    send_command(node_name, "Pool.add_transaction(revoke)")
+  def verify_with_delay(valid?, execute_times) do
+    if valid?.() do
+      true
+    else
+      :timer.sleep(1000)
```
cytadela8: Maybe lower the sleep to 100ms? Running check more frequently should lower the time it takes for a test to pass.

thepiwo:
@gspasov @0legIv is there still something to to here, except adjusting to comments?

gspasov:
Yes, and maybe the timeouts should be picked up, because apparently Travis runs things slowly. No problems on my machine.

Issue #543, opened 2018-08-16
Channel transactions serialization adjustments


cytadela8:
Adjustment of channel transactions serialization. 
This excludes ChannelStateOffChain serialization required by `ChannelSlashTx` and `ChannelSoloCloseTx`.

thepiwo:
@cytadela8 this is resolved by #623 right?

thepiwo:
ah it is not, it will be resolved with #586

cytadela8:
This was resolved by #669

PR #623, opened 2018-09-18
[GH-533][GH-543] Channels serialization and tests

cytadela8:
This fixes all Channel serialization and add tests for most parts. More tests are still needed, but we (I and @grzegorz225) want to merge master to our work so first we want to merge our changes together.

Changes in this PR:
- possibility to get senders of Tx from chainstate (required for ChannelCloseMutalTx)
- adjusted serialization of all channel staratures
- mix format
- serialization tests with data from epoch


cytadela8:
Everything passes, but one of tests timesout for some reason. Will fix conflicts and investigate later.

cytadela8:
I'm repeatedly getting this error in CI:
```
14:21:59.774 [error] ** (MatchError) no match of right hand side value: {:error, "HTTPPoison Error %HTTPoison.Error{id: nil, reason: :econnrefused}"}
14:22:01.206 [error] 14:22:01.201 [error] Can't add ourself
  1) test namings test (MultiNodeSyncTest)
test/multi_node_sync_test.exs:82
Assertion with == failed
code:assert :synced == TestFramework.compare_nodes_by_top_block_hash("node1", "node4")
left::synced
right::not_synced
stacktrace:
       test/multi_node_sync_test.exs:112: (test)
```

apps/aecore/lib/aecore/chain/identifier.ex
```diff
@@ -10,6 +10,7 @@ defmodule Aecore.Chain.Identifier do
 
   alias __MODULE__
   defstruct type: :undefined, value: ""
+  use ExConstructor
```
grzegorz225: Is this used somewhere?

apps/aecore/lib/aecore/channel/channel_off_chain_tx.ex
```diff
@@ -215,13 +219,21 @@ defmodule Aecore.Channel.ChannelOffChainTx do
...
       ]) do
-    {:ok, channel_id} = Identifier.decode_from_binary(encoded_channel_id)
-    %ChannelOffChainTx{
-      channel_id: channel_id,
-      sequence: :binary.decode_unsigned(sequence),
-      updates: Enum.map(encoded_updates, &ChannelOffChainUpdate.from_list/1),
-      state_hash: state_hash
-    }
+    with {:ok, %Identifier{type: :channel, value: channel_id}} <-
+           Identifier.decode_from_binary(encoded_channel_id) do
```
grzegorz225: The check for the expected type should be done inside Identifier -> in this way we avoid duplicating the type check code in every tx that de-serializes identifiers.

apps/aecore/lib/aecore/channel/channel_state_on_chain.ex
```diff
@@ -249,30 +292,42 @@ defmodule Aecore.Channel.ChannelStateOnChain do
...
-       slash_close: :binary.decode_unsigned(slash_close),
-       slash_sequence: :binary.decode_unsigned(slash_sequence),
-       state_hash: state_hash,
-       channel_reserve: :binary.decode_unsigned(channel_reserve)
-     }}
+    responder_amount =
+      :binary.decode_unsigned(total_amount) - :binary.decode_unsigned(initiator_amount)
+
+    with {:ok, %Identifier{type: :account, value: initiator_pubkey}} <-
+           Identifier.decode_from_binary(encoded_initiator_pubkey),
```
grzegorz225: Same here

apps/aecore/lib/aecore/channel/tx/channel_close_mutal_tx.ex
```diff
@@ -51,6 +50,19 @@ defmodule Aecore.Channel.Tx.ChannelCloseMutalTx do
   @spec get_chain_state_name :: atom()
   def get_chain_state_name, do: :channels
 
+  def chainstate_senders?(), do: true
```
grzegorz225: Could we deduce this from DataTx alone? For instance:
```
chainstate_senders?(data_tx), do: DataTx.senders(data_tx) == []
```

apps/aecore/lib/aecore/tx/data_tx.ex
```diff
@@ -194,12 +198,23 @@ defmodule Aecore.Tx.DataTx do
     end
   end
 
+  @spec chainstate_senders?(DataTx.t()) :: boolean()
+  def chainstate_senders?(%DataTx{type: type}) do
+    type.chainstate_senders?()
```
grzegorz225: This can be implemented without dispatching a function call.
```
senders == []
```

apps/aecore/lib/aecore/tx/data_tx.ex
```diff
@@ -194,12 +198,23 @@ defmodule Aecore.Tx.DataTx do
...
+    type.senders_from_chainstate(payload, chainstate)
+  end
+
   @doc """
   Validates the transaction without considering state
   """
   @spec validate(DataTx.t(), non_neg_integer()) :: :ok | {:error, String.t()}
   def validate(
         %DataTx{fee: fee, type: type, senders: senders} = tx,
+        # FIXME
```
grzegorz225: Is this a placeholder for a future PR?

apps/aecore/lib/aecore/tx/signed_tx.ex
```diff
@@ -50,7 +50,7 @@ defmodule Aecore.Tx.SignedTx do
 
   @spec validate(SignedTx.t(), non_neg_integer()) :: :ok | {:error, String.t()}
   def validate(%SignedTx{data: data} = tx, block_height) do
-    if signatures_valid?(tx) do
+    if DataTx.chainstate_senders?(data) || signatures_valid?(tx, DataTx.senders(data)) do
```
grzegorz225: Will this break anything? Skipping the signature check when the pubkeys are not available does not seem right. Could the tx verification logic pass the chainstate here so the verification could proceed?

apps/aecore/lib/aecore/tx/signed_tx.ex
```diff
@@ -77,8 +82,7 @@ defmodule Aecore.Tx.SignedTx do
      - tx: The transaction data that it's going to be signed
      - priv_key: The priv key to sign with
   """
-  @spec sign_tx(DataTx.t() | SignedTx.t(), binary()) ::
-          {:ok, SignedTx.t()} | {:error, String.t()}
+  @spec sign_tx(DataTx.t() | SignedTx.t(), binary()) :: {:ok, SignedTx.t()} | {:error, String.t()}
```
grzegorz225: {:error, String.t()} is repeated a lot. Maybe define a error() type?

thepiwo:
@cytadela8 the multinode issues are supposed to be resolved with https://github.com/aeternity/elixir-node/pull/617

apps/aecore/lib/aecore/chain/identifier.ex
cytadela8: It was added like everywhere. Will remove as we are changing approach.

apps/aecore/lib/aecore/channel/tx/channel_close_mutal_tx.ex
cytadela8: Current way is safer. `chainstate_senders?` should depend on type. If due to some kind of bug no senders are included then SignedTx.validate can fail nicely.

I think your proposed solution although potentially easier to implement is too "hack-y". It's a little bit like "we encode the information that senders are from chainstate (and function to get them is expected in transaction module) by setting senders to `[]`". I think that's hacky.

apps/aecore/lib/aecore/tx/data_tx.ex
cytadela8: `Chain.top_height()` shouldn't be used as block_height ever.

apps/aecore/lib/aecore/tx/signed_tx.ex
cytadela8: Chainstate is not available by definition in `validate` functions. Preprocess check is for checks requiring chainstate.

apps/aecore/lib/aecore/tx/data_tx.ex
cytadela8: I just opened an issue for that #626

thepiwo:
@cytadela8 multinode tests are now disabled on master, so they should pass here, after merging

apps/aecore/lib/aecore/tx/signed_tx.ex
cytadela8: I think I would leave such changes to some refactoring later as this applies to most of our codebase. I'm open to introducing such changes to channels in channels PR, but we shouldn't refactor whole codebase in channels PR.

thepiwo:
@cytadela8 @grzegorz225 so should we merge this

Issue #574, opened 2018-09-03
Modify the test framework to work with the Epoch


0legIv:
Minimal automated test to check if epoch does sync with elixir-node:
1. Download epoch release
2. Start, connect
3. Do all kinds of transactions in elixir
4. Check if same block hash at height if mined the txs

PR #630, opened 2018-09-25
[GH-545] added aeternity erlang compatibility doc

### LINKED ISSUE #545: Write step-by-step guide on how to check consensus to be epoch 0.16 compatible
add a docs/epoch-compatibility.md with `## epoch 0.16` headline and description

thepiwo:
resolves #545

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,9 @@
 # Aeternity Elixir Full Node
 This is an elixir full node implementation of the aeternity specification.
 
+Compatibility to the erlang aeternity implementation is documented in [docs/aeternity-erlang-compatibility.md](docs/aeternity-erlang-compatibility.md).
+
+
 ## Getting started
 
 ### Required packages
--- a/docs/aeternity-erlang-compatibility.md
+++ b/docs/aeternity-erlang-compatibility.md
@@ -1,0 +1,201 @@
+# aeternity Erlang implementation compatibility
+
+## 0.16.0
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -3,6 +3,9 @@
 # Aeternity Elixir Full Node
 This is an elixir full node implementation of the aeternity specification.
 
+Compatibility to the erlang aeternity implementation is documented in [docs/aeternity-erlang-compatibility.md](docs/aeternity-erlang-compatibility.md).
```
grzegorz225: I think this should be written more like:

Tests for checking compatibility to the epoch aeternity implementation are documented in...

docs/aeternity-erlang-compatibility.md
```diff
@@ -0,0 +1,201 @@
+# aeternity Erlang implementation compatibility
```
grzegorz225: Shouldn't we make a test framework for this?
We could download epoch via the makefile and create a wrapper for ExUnit to directly invoke and test code running in epoch.

thepiwo: Yes, there is already and issue for this https://github.com/aeternity/elixir-node/issues/574

README.md
thepiwo: this should be more of a documentation with future versions just stating changes, without more specific cases. The sample cases are just for outsiders to quickly get up and running to check these the documented cases.

PR #631, opened 2018-09-26
[GH-626] removed top_block reference from validation

### LINKED ISSUE #626: Remove top block references from transactions
`Aecore.Chain.Worker` module is for storing current state of chain and information about consensus (top block, longest chain, current chainstate), but `Aecore.Chain.Worker.top_block()` is used as a default value for `block_height` parameter in `Aecore.Tx.DataTx.validate`.

In general when block is validated we should only use previous block and previous chainstate to do that. In transactions and block validation logic there should be no references to current top block or anything like that.

thepiwo:
resolves #626


cytadela8:
I see Chain is also referenced in OracleQueryTx and OracleResponseTx. Would you like to create an separate issue for that?

thepiwo:
@cytadela8 I will, thank you!

PR #649, opened 2018-10-02
[GH-640] Added check if txs minimum fees are met when validating a block

### LINKED ISSUE #640: Check if txs minimum fees are met when validating a block

DanielaIvanova:
resolves #640

### DIFF
--- a/apps/aecore/lib/aecore/chain/block_validation.ex
+++ b/apps/aecore/lib/aecore/chain/block_validation.ex
@@ -119,7 +119,7 @@ defmodule Aecore.Chain.BlockValidation do
 
   @spec check_is_minimum_fee_met?(list()) :: boolean()
   def check_is_minimum_fee_met?(txs) do
-    Enum.all?(txs, fn tx -> Pool.is_minimum_fee_met?(tx, :pool) end)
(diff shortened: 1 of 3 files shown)


apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -113,6 +117,11 @@ defmodule Aecore.Chain.BlockValidation do
     Enum.map(txs, fn tx -> :ok == SignedTx.validate(tx) end)
   end
 
+  @spec check_is_minimum_fee_met?(list()) :: boolean()
+  def check_is_minimum_fee_met?(txs) do
+    Enum.all?(txs, fn tx -> Pool.is_minimum_fee_met?(tx, :pool) end)
```
cytadela8: Why isn't identifier (second argument to `is_minimum_fee_met?`) `:validation`?

DanielaIvanova: I will change to `:validation`

thepiwo:
regarding the commented issue I proposed a later task that will fix this https://github.com/aeternity/elixir-node/issues/659

PR #653, opened 2018-10-03
[GH-652] Automatically added spend_tx and naming txs to the pool upon creation

### LINKED ISSUE #652: Automatically add spend_tx and naming txs to the pool upon creation

DanielaIvanova:
resolves #652

### DIFF
--- a/apps/aecore/lib/aecore/account/account.ex
+++ b/apps/aecore/lib/aecore/account/account.ex
@@ -11,6 +11,7 @@ defmodule Aecore.Account.Account do
   alias Aecore.Naming.{NameCommitment, NameUtil}
   alias Aecore.Naming.Tx.{NamePreClaimTx, NameClaimTx, NameUpdateTx, NameTransferTx, NameRevokeTx}
   alias Aecore.Tx.{DataTx, SignedTx}
+  alias Aecore.Tx.Pool.Worker, as: Pool
   alias Aeutil.{Bits, Serialization}
 
   require Logger
@@ -323,7 +324,9 @@ defmodule Aecore.Account.Account do
         ) :: {:ok, SignedTx.t()} | {:error, reason()}
   def build_tx(payload, tx_type, sender, sender_prv, fee, nonce, ttl \\ 0) do
     tx = DataTx.init(tx_type, payload, sender, fee, nonce, ttl)
-    SignedTx.sign_tx(tx, sender, sender_prv)
+    {:ok, tx} = SignedTx.sign_tx(tx, sender, sender_prv)
+    Pool.add_transaction(tx)
+    tx
   end
 
   @doc """
--- a/apps/aecore/test/aecore_chain_state_test.exs
+++ b/apps/aecore/test/aecore_chain_state_test.exs
@@ -37,10 +37,10 @@ defmodule AecoreChainstateTest do
   test "chain state", wallet do
     init_accounts_state = Chain.chain_state().accounts
 
-    {:ok, signed_tx1} =
+    signed_tx1 =
       Account.spend(wallet.b_pub_key, wallet.b_priv_key, wallet.a_pub_key, 1, 1, 2, <<"payload">>)
 
-    {:ok, signed_tx2} =
+    signed_tx2 =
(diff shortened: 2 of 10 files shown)


apps/aecore/test/aecore_chain_state_test.exs
```diff
@@ -37,11 +38,9 @@ defmodule AecoreChainstateTest do
...
     init_accounts_state = Chain.chain_state().accounts
 
-    {:ok, signed_tx1} =
-      Account.spend(wallet.b_pub_key, wallet.b_priv_key, wallet.a_pub_key, 1, 1, 2, <<"payload">>)
-
-    {:ok, signed_tx2} =
-      Account.spend(wallet.c_pub_key, wallet.c_priv_key, wallet.a_pub_key, 2, 1, 2, <<"payload">>)
+    Account.spend(wallet.b_pub_key, wallet.b_priv_key, wallet.a_pub_key, 1, 1, 2, <<"payload">>)
+    Account.spend(wallet.c_pub_key, wallet.c_priv_key, wallet.a_pub_key, 2, 1, 2, <<"payload">>)
+    txs = Pool.get_pool() |> Map.values()
```
thepiwo: While testing the miner is running in the background, so they could be gone already, the old way looked better

apps/aecore/lib/aecore/account/account.ex
```diff
@@ -320,10 +321,11 @@ defmodule Aecore.Account.Account do
           non_neg_integer(),
           non_neg_integer(),
           non_neg_integer()
-        ) :: {:ok, SignedTx.t()} | {:error, reason()}
+        ) :: :ok | :error
   def build_tx(payload, tx_type, sender, sender_prv, fee, nonce, ttl \\ 0) do
     tx = DataTx.init(tx_type, payload, sender, fee, nonce, ttl)
-    SignedTx.sign_tx(tx, sender, sender_prv)
+    {:ok, tx} = SignedTx.sign_tx(tx, sender, sender_prv)
+    Pool.add_transaction(tx)
```
thepiwo: You can still return the tx

apps/aecore/test/aecore_naming_test.exs
```diff
@@ -412,14 +387,13 @@ defmodule AecoreNamingTest do
...
-    {:ok, transfer} = Account.name_transfer("test.aet", transfer_to_pub, 5)
-    Pool.add_transaction(transfer)
+    Account.name_transfer("test.aet", transfer_to_pub, 5)
     Miner.mine_sync_block_to_chain()
 
     naming_state_4 = Chain.chain_state().naming
 
     assert 1 == naming_state_4 |> PatriciaMerkleTree.all_keys() |> Enum.count()
-    hash_1 = transfer.data.payload.hash.value
+    {:ok, hash_1} = NameUtil.normalized_namehash("test.aet")
```
thepiwo: The data from the original transaction should be used, so in case there is an issue there we will catch it here

apps/aecore/lib/aecore/account/account.ex
DanielaIvanova: Ok, I'll return tx

thepiwo: @DanielaIvanova the comments apply for all occurrences of this in the PR

apps/aecore/test/aecore_chain_state_test.exs
cytadela8: I'm pretty sure the miner isn't running in background on tests, but I agree.

thepiwo: okay, I am not to sure about this, but it should be adjusted anyways

apps/aecore/test/aecore_naming_test.exs
```diff
@@ -38,7 +38,7 @@ defmodule AecoreNamingTest do
   @tag :naming
   test "test naming workflow", setup do
     Miner.mine_sync_block_to_chain()
-    {:ok, pre_claim} = Account.pre_claim("test.aet", 123, 5)
+    pre_claim = Account.pre_claim("test.aet", 123, 5)
     Pool.add_transaction(pre_claim)
```
d-velev: It's not necessary to add the transaction anymore right?

PR #663, opened 2018-10-08
[GH-656] Adjusted code duplication between state trees

### LINKED ISSUE #656: Code duplication between state trees
Lots of code is duplicated between different `*StateTree` modules. We should make a module using mechanics like in `Aecore.Util.Serializable` to remove the duplication.

DanielaIvanova:
resolves #656

### DIFF
--- a/apps/aecore/lib/aecore/account/account_state_tree.ex
+++ b/apps/aecore/lib/aecore/account/account_state_tree.ex
@@ -2,7 +2,7 @@ defmodule Aecore.Account.AccountStateTree do
   @moduledoc """
   Top level account state tree.
   """
-  use Aecore.Util.StateTrees
+  use Aecore.Util.StateTrees, [:accounts, Aecore.Account.Account]
 
   alias Aecore.Account.Account
   alias Aecore.Chain.Identifier
@@ -12,9 +12,6 @@ defmodule Aecore.Account.AccountStateTree do
 
   @typedoc "Accounts tree"
   @type accounts_state :: Trie.t()
-
-  @spec tree_type :: atom()
-  def tree_type, do: :accounts
 
   @spec get(accounts_state(), Keys.pubkey()) :: Account.t()
   def get(tree, key) do
@@ -28,10 +25,5 @@ defmodule Aecore.Account.AccountStateTree do
         id = Identifier.create_identity(key, :account)
         %Account{acc | id: id}
     end
-  end
-
-  @spec update(accounts_state(), Keys.pubkey(), (Account.t() -> Account.t())) :: accounts_state()
-  def update(tree, key, fun) do
-    put(tree, key, fun.(get(tree, key)))
   end
 end
--- a/apps/aecore/lib/aecore/account/account_state_tree.ex
+++ b/apps/aecore/lib/aecore/account/account_state_tree.ex
@@ -13,9 +13,6 @@ defmodule Aecore.Account.AccountStateTree do
   @typedoc "Accounts tree"
   @type accounts_state :: Trie.t()
 
-  @typedoc "Hash of the tree"
-  @type hash :: binary()
-
   @spec get(accounts_state(), Keys.pubkey()) :: Account.t()
   def get(tree, key) do
     case PatriciaMerkleTree.lookup(tree, key) do
--- a/apps/aecore/lib/aecore/account/account_state_tree.ex
+++ b/apps/aecore/lib/aecore/account/account_state_tree.ex
@@ -2,6 +2,8 @@ defmodule Aecore.Account.AccountStateTree do
   @moduledoc """
   Top level account state tree.
   """
+  use Aecore.Util.StateTrees
+
   alias Aecore.Account.Account
   alias Aecore.Chain.Identifier
   alias Aecore.Keys
@@ -14,10 +16,8 @@ defmodule Aecore.Account.AccountStateTree do
   @typedoc "Hash of the tree"
   @type hash :: binary()
 
-  @spec init_empty() :: accounts_state()
-  def init_empty do
-    PatriciaMerkleTree.new(:accounts)
-  end
+  @spec name() :: atom()
+  def name(), do: :accounts
 
   @spec put(accounts_state(), Keys.pubkey(), Account.t()) :: accounts_state()
   def put(tree, key, value) do
--- a/apps/aecore/lib/aecore/account/account_state_tree.ex
+++ b/apps/aecore/lib/aecore/account/account_state_tree.ex
@@ -19,12 +19,6 @@ defmodule Aecore.Account.AccountStateTree do
   @spec name() :: atom()
   def name(), do: :accounts
 
-  @spec put(accounts_state(), Keys.pubkey(), Account.t()) :: accounts_state()
-  def put(tree, key, value) do
-    serialized_account_state = Account.rlp_encode(value)
-    PatriciaMerkleTree.enter(tree, key, serialized_account_state)
-  end
-
   @spec get(accounts_state(), Keys.pubkey()) :: Account.t()
   def get(tree, key) do
     case PatriciaMerkleTree.lookup(tree, key) do
@@ -42,15 +36,5 @@ defmodule Aecore.Account.AccountStateTree do
   @spec update(accounts_state(), Keys.pubkey(), (Account.t() -> Account.t())) :: accounts_state()
   def update(tree, key, fun) do
     put(tree, key, fun.(get(tree, key)))
-  end
-
-  @spec has_key?(accounts_state(), Keys.pubkey()) :: boolean()
-  def has_key?(tree, key) do
-    PatriciaMerkleTree.lookup(tree, key) != :none
-  end
-
-  @spec root_hash(accounts_state()) :: hash()
-  def root_hash(tree) do
-    PatriciaMerkleTree.root_hash(tree)
   end
 end
--- a/apps/aecore/lib/aecore/account/account_state_tree.ex
+++ b/apps/aecore/lib/aecore/account/account_state_tree.ex
@@ -25,7 +25,8 @@ defmodule Aecore.Account.AccountStateTree do
     end
   end
 
-  @spec process_struct(Account.t(), binary(), accounts_state()) :: Account.t()
+  @spec process_struct(Account.t(), binary(), accounts_state()) ::
+          Account.t() | {:error, String.t()}
   def process_struct(%Account{} = deserialized_value, key, _tree) do
     id = Identifier.create_identity(key, :account)
     %Account{deserialized_value | id: id}
--- a/apps/aecore/lib/aecore/account/account_state_tree.ex
+++ b/apps/aecore/lib/aecore/account/account_state_tree.ex
@@ -13,6 +13,9 @@ defmodule Aecore.Account.AccountStateTree do
   @typedoc "Accounts tree"
   @type accounts_state :: Trie.t()
 
+  @spec tree_type :: atom()
+  def tree_type, do: :accounts
+
   @spec get(accounts_state(), Keys.pubkey()) :: Account.t()
   def get(tree, key) do
     case PatriciaMerkleTree.lookup(tree, key) do
--- a/apps/aecore/lib/aecore/account/account_state_tree.ex
+++ b/apps/aecore/lib/aecore/account/account_state_tree.ex
@@ -16,9 +16,6 @@ defmodule Aecore.Account.AccountStateTree do
   @typedoc "Hash of the tree"
   @type hash :: binary()
 
-  @spec name() :: atom()
-  def name(), do: :accounts
-
   @spec get(accounts_state(), Keys.pubkey()) :: Account.t()
   def get(tree, key) do
     case PatriciaMerkleTree.lookup(tree, key) do
--- a/apps/aecore/lib/aecore/account/account_state_tree.ex
+++ b/apps/aecore/lib/aecore/account/account_state_tree.ex
@@ -16,14 +16,23 @@ defmodule Aecore.Account.AccountStateTree do
   @spec get(accounts_state(), Keys.pubkey()) :: Account.t()
   def get(tree, key) do
     case PatriciaMerkleTree.lookup(tree, key) do
-      :none ->
-        Account.empty()
-
       {:ok, account_state} ->
         {:ok, acc} = Account.rlp_decode(account_state)
+        process_struct(acc, key, tree)
 
-        id = Identifier.create_identity(key, :account)
-        %Account{acc | id: id}
+      :none ->
+        Account.empty()
     end
+  end
+
+  @spec process_struct(Account.t(), binary(), accounts_state()) :: Account.t()
+  def process_struct(%Account{} = deserialized_value, key, _tree) do
+    id = Identifier.create_identity(key, :account)
+    %Account{deserialized_value | id: id}
+  end
+
+  def process_struct(deserialized_value, _key, _tree) do
+    {:error,
+     "#{__MODULE__}: Invalid data type: #{deserialized_value.__struct__} but expected %Account{}"}
   end
 end
--- a/apps/aecore/lib/aecore/channel/channel_state_tree.ex
+++ b/apps/aecore/lib/aecore/channel/channel_state_tree.ex
@@ -28,7 +28,7 @@ defmodule Aecore.Channel.ChannelStateTree do
   end
 
   @spec process_struct(ChannelStateOnChain.t(), ChannelStateOnChain.id(), channel_state()) ::
-          ChannelStateOnChain.t()
+          ChannelStateOnChain.t() | {:error, String.t()}
   def process_struct(%ChannelStateOnChain{} = deserialized_value, _key, _tree) do
     deserialized_value
   end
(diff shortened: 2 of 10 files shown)


apps/aecore/lib/aecore/util/state_trees.ex
```diff
@@ -0,0 +1,95 @@
...
+      def has_key?(tree, key) do
+        PatriciaMerkleTree.lookup(tree, key) != :none
+      end
+
+      @spec delete(Trie.t(), binary()) :: Trie.t()
+      def delete(tree, key) do
+        PatriciaMerkleTree.delete(tree, key)
+      end
+
+      defoverridable get: 2
```
thepiwo: what is this for?

apps/aecore/lib/aecore/util/state_trees.ex
```diff
@@ -0,0 +1,95 @@
...
+                identified_id = Identifier.create_identity(key, :contract)
+                store_id = Contract.store_id(%{deserialized_value | id: identified_id})
+
+                %Contract{
+                  deserialized_value
+                  | id: identified_id,
+                    store: ContractStateTree.get_store(store_id, tree)
+                }
+
+              _ ->
```
thepiwo: do we use the default here? otherwise an error message would be better, or not handling the case at all

DanielaIvanova: We are using `case do` because of the identifiers(in the structures that we need), as epoch does. Where we don't need identifiers we just return the deserialized value.

thepiwo: ah okay!

apps/aecore/lib/aecore/util/state_trees.ex
DanielaIvanova: If Some module implements its own function `get` with 2 arguments, it will use this function. If not will use the one defined in the macro.

apps/aecore/lib/aecore/util/state_trees.ex
```diff
@@ -0,0 +1,95 @@
...
+        PatriciaMerkleTree.enter(tree, key, serialized_state)
+      end
+
+      @spec get(Trie.t(), binary()) :: :none | map()
+      def get(tree, key) do
+        case PatriciaMerkleTree.lookup(tree, key) do
+          {:ok, serialized_value} ->
+            {:ok, deserialized_value} = Serialization.rlp_decode_anything(serialized_value)
+
+            case deserialized_value do
```
cytadela8: This `case ... do` prevents `Aecore.Util.StateTrees` from being abstract. You can extract it to a function and each ***StateTree module will implement modification of corresponding structure.

apps/aecore/lib/aecore/util/state_trees.ex
```diff
@@ -0,0 +1,95 @@
...
+      def delete(tree, key) do
+        PatriciaMerkleTree.delete(tree, key)
+      end
+
+      defoverridable get: 2
+    end
+  end
+
+  # New tree definitions should be described and handled here
+  def tree_type(AccountStateTree), do: :accounts
```
cytadela8: `tree_type` can also be implemented in corresponding modules. You call such a function as follows: `__MODULE__.tree_type()`

apps/aecore/lib/aecore/account/account_state_tree.ex
```diff
@@ -43,14 +31,4 @@ defmodule Aecore.Account.AccountStateTree do
   def update(tree, key, fun) do
```
cytadela8: Update can also be moved to `Aecore.Util.StateTree`.

apps/aecore/lib/aecore/util/state_trees.ex
```diff
@@ -0,0 +1,95 @@
...
+      @typedoc "Hash of the tree"
+      @type hash :: binary()
+
+      @spec init_empty :: Trie.t()
+      def init_empty do
+        PatriciaMerkleTree.new(StateTrees.tree_type(__MODULE__))
+      end
+
+      @spec put(Trie.t(), binary(), map()) :: Trie.t()
+      def put(tree, key, value) do
```
grzegorz225: There is no type safety here. Imagine that we put an oracle into the channels trie and then try to fetch it.

apps/aecore/lib/aecore/util/state_trees.ex
```diff
@@ -0,0 +1,95 @@
...
+  alias Aecore.Account.AccountStateTree
+  alias Aecore.Contract.CallStateTree
+  alias Aecore.Contract.ContractStateTree
+  alias Aecore.Channel.ChannelStateTree
+  alias Aecore.Naming.NamingStateTree
+
+  @moduledoc """
+  Abstract module defining functions for known Tree types
+  """
+  defmacro __using__(_) do
```
grzegorz225: The __using__ macro can have parameters. This can be used here to remove a lot of code and guarantee type safety. 
For instance by declaring the __using__ macro as:
```
defmacro __using__(tree_type, stored_type) when is_atom(tree_type) and is_module(stored_type) do
```
We can then write for instance:
```
PatriciaMerkleTree.new(tree_type)
```
instead of
```
PatriciaMerkleTree.new(StateTrees.tree_type(__MODULE__))
```
or more importantly:
```
{:ok, %stored_type{} = deserialized_value} = Serialization.rlp_decode_only(serialized_value, stored_type)
```
instead of:
```
{:ok, deserialized_value} = Serialization.rlp_decode_anything(serialized_value)
```
Then to use the new code you write:
```
use Aecore.Util.StateTrees, :non_existant_tree, RandomStoredType 
```

This approach together with unquoting/quoting code will make this module pretty small.
Please define an overwridable callback which returns an default value for an unknown key (the default implementation will return an error). 

As @cytadela8 said bellow make an callback to setup the deserialized value. For instance:
```
def setup_value(identifier, decoded_value) do
raise "Provide an custom implementation"
end

defoverridable setup_value:2
```

DanielaIvanova:
@cytadela8 @grzegorz225 
In our approach and the way we implemented there is no need to implement `behavior` and `callback` 
When we write `use Aecore.Util.StateTrees` to some module we inject the functions from the macro to the module. We don't need any specific functions to be implemented in that module.

DanielaIvanova:
@thepiwo 
At the current state, we think that we are ready.

apps/aecore/lib/aecore/contract/contract_state_tree.ex
```diff
@@ -50,30 +44,32 @@ defmodule Aecore.Contract.ContractStateTree do
...
-        identified_id = Identifier.create_identity(key, :contract)
-
-        store_id = Contract.store_id(%{deserialized | id: identified_id})
-
-        %Contract{
-          deserialized
-          | id: identified_id,
-            store: get_store(store_id, contract_tree)
-        }
+  def get_store(store_id, tree) do
```
thepiwo: can you add `@spec` here?

apps/aecore/lib/aecore/util/state_trees.ex
```diff
@@ -0,0 +1,88 @@
...
+      def update(tree, key, fun) do
+        put(tree, key, fun.(get(tree, key)))
+      end
+
+      @spec delete(Trie.t(), binary()) :: Trie.t()
+      def delete(tree, key) do
+        PatriciaMerkleTree.delete(tree, key)
+      end
+
+      defoverridable get: 2, update: 3
```
thepiwo: can you add a comment here with what you answered me on my initial question about this?

apps/aecore/lib/aecore/util/state_trees.ex
```diff
@@ -0,0 +1,88 @@
...
+      end
+
+      @spec get(Trie.t(), binary()) :: map() | :none | {:error, String.t()}
+      def get(tree, key) do
+        case PatriciaMerkleTree.lookup(tree, key) do
+          {:ok, serialized_value} ->
+            {:ok, deserialized_value} = Serialization.rlp_decode_anything(serialized_value)
+
+            if StateTrees.valide_store_type?(deserialized_value, unquote(stored_type)) do
+              __MODULE__.process_struct(deserialized_value, key, tree)
```
cytadela8: The `process_struct` function is required to be implemented by ***StateTrees. There should be a callback specification that it's required or you can add a default.

DanielaIvanova: You are right! You will fix this. Thank you

thepiwo:
@cytadela8 can you approve?

apps/aecore/lib/aecore/account/account_state_tree.ex
```diff
@@ -11,46 +13,26 @@ defmodule Aecore.Account.AccountStateTree do
...
 
-  @spec update(accounts_state(), Keys.pubkey(), (Account.t() -> Account.t())) :: accounts_state()
-  def update(tree, key, fun) do
-    put(tree, key, fun.(get(tree, key)))
-  end
-
-  @spec has_key?(accounts_state(), Keys.pubkey()) :: boolean()
-  def has_key?(tree, key) do
-    PatriciaMerkleTree.lookup(tree, key) != :none
+  @spec process_struct(Account.t(), binary(), accounts_state()) :: Account.t()
```
grzegorz225: ''' process_struct ''' can error in case the pattern match fails. Either change the type spec to:
```
@spec process_struct(Account.t(), binary(), accounts_state()) :: Account.t() | {:error, String.t()}
```
or change the alternate case to:
```
throw {:error, "#{__MODULE__}: Invalid data type: #{deserialized_value.__struct__} but expected %Account{}"} 
```

apps/aecore/lib/aecore/channel/channel_state_tree.ex
```diff
@@ -55,13 +27,14 @@ defmodule Aecore.Channel.ChannelStateTree do
...
+  @spec process_struct(ChannelStateOnChain.t(), ChannelStateOnChain.id(), channel_state()) ::
+          ChannelStateOnChain.t()
+  def process_struct(%ChannelStateOnChain{} = deserialized_value, _key, _tree) do
+    deserialized_value
   end
 
-  @spec root_hash(channel_state()) :: hash()
-  def root_hash(tree) do
-    PatriciaMerkleTree.root_hash(tree)
+  def process_struct(deserialized_value, _key, _tree) do
```
grzegorz225: Same as above.

apps/aecore/lib/aecore/contract/call_state_tree.ex
```diff
@@ -36,25 +30,18 @@ defmodule Aecore.Contract.CallStateTree do
...
-    end
+  @spec construct_call_tree_id(binary(), binary()) :: binary()
+  def construct_call_tree_id(contract_id, call_id) do
+    <<contract_id.value::binary, call_id::binary>>
   end
 
-  @spec root_hash(calls_state()) :: hash()
-  def root_hash(calls_tree) do
-    PatriciaMerkleTree.root_hash(calls_tree)
+  @spec process_struct(Call.t(), binary(), calls_state()) :: Call.t()
```
grzegorz225: Same as above

apps/aecore/lib/aecore/contract/contract_state_tree.ex
```diff
@@ -50,30 +44,37 @@ defmodule Aecore.Contract.ContractStateTree do
...
   end
 
-  @spec root_hash(contracts_state()) :: hash()
-  def root_hash(contract_tree) do
-    PatriciaMerkleTree.root_hash(contract_tree)
+  defp get_store(store_id, tree) do
+    keys = PatriciaMerkleTree.all_keys(tree)
+    store_id_bit_size = (@contract_key_size + 1) * 8
+
+    Enum.reduce(keys, %{}, fn key, store_acc ->
```
grzegorz225: @DanielaIvanova
We can have a lot of contracts on the blockchain. Ethereum currently has more than a million of then (https://etherscan.io/accounts/c).

This code has a time complexity of O(N) which means that this code will be unusable in production -> Each O(log(N)) lookup will degrade to O(N).

The operation implemented here corresponds in fact to taking an subset of the whole trie which can be done in O(logN). Open an issue to implement this operation in our MerklePatriciaLibrary - we will need to implement a module providing a read/write view onto a subset of a Merkle Patricia Trie - Please note that we cannot simply lookup all the keys in a sub tree and create a new map from this as this could be potentially time consuming. 

@cytadela8 @thepiwo I would suggest that we merge this PR to master but open an issue about this so we do not forget to rewrite this. 

@DanielaIvanova Add a comment to the code stating that this snipped has a time complexity od O(n) and needs to be rewritten.

apps/aecore/lib/aecore/naming/name.ex
```diff
@@ -23,7 +23,7 @@ defmodule Aecore.Naming.Name do
   @typedoc "Structure of the Name Transaction type"
   @type t :: %Name{
           hash: binary(),
-          owner: Keys.pubkey(),
+          owner: Identifier.t(),
```
grzegorz225: This will break serialization of the name structure:
https://github.com/aeternity/elixir-node/blob/c76f802d029419e90ab86d841948c1bc3d4349fa/apps/aecore/lib/aecore/naming/name.ex#L94

apps/aecore/lib/aecore/naming/name_commitment.ex
```diff
@@ -17,7 +17,7 @@ defmodule Aecore.Naming.NameCommitment do
   @typedoc "Structure of the NameCommitment Transaction type"
   @type t :: %NameCommitment{
           hash: binary(),
-          owner: Keys.pubkey(),
+          owner: Identifier.t(),
```
grzegorz225: Same as above -> This will break the serialization:
https://github.com/aeternity/elixir-node/blob/c76f802d029419e90ab86d841948c1bc3d4349fa/apps/aecore/lib/aecore/naming/name_commitment.ex#L75

apps/aecore/lib/aecore/util/state_trees.ex
```diff
@@ -0,0 +1,97 @@
...
+
+      @typedoc "Hash of the tree"
+      @type hash :: binary()
+
+      @spec init_empty :: Trie.t()
+      def init_empty do
+        PatriciaMerkleTree.new(unquote(tree_type))
+      end
+
+      @spec put(Trie.t(), binary(), map()) :: Trie.t() | {:error, String.t()}
```
grzegorz225: Fix typespec -> map is too general.

apps/aecore/lib/aecore/util/state_trees.ex
```diff
@@ -0,0 +1,97 @@
...
+          PatriciaMerkleTree.enter(tree, key, serialized_state)
+        else
+          {:error,
+           "#{__MODULE__}: Invalid value type: #{value.__struct__} but expected value type #{
+             unquote(stored_type)
+           }"}
+        end
+      end
+
+      @spec get(Trie.t(), binary()) :: map() | :none | {:error, String.t()}
```
grzegorz225: Same as above

apps/aecore/lib/aecore/util/state_trees.ex
```diff
@@ -0,0 +1,97 @@
...
+      def get(tree, key) do
+        case PatriciaMerkleTree.lookup(tree, key) do
```
grzegorz225: Why not use ``` with ``` to remove nesting? The deserialization of the value can error - we should handle this.

apps/aecore/lib/aecore/util/state_trees.ex
```diff
@@ -0,0 +1,97 @@
...
+      def root_hash(tree) do
+        PatriciaMerkleTree.root_hash(tree)
+      end
+
+      @spec has_key?(Trie.t(), binary()) :: boolean()
+      def has_key?(tree, key) do
+        PatriciaMerkleTree.lookup(tree, key) != :none
+      end
+
+      @spec update(Trie.t(), binary(), (any() -> any())) :: Trie.t()
```
grzegorz225: More specific typespec would be nice.

apps/aecore/lib/aecore/util/state_trees.ex
```diff
@@ -0,0 +1,97 @@
...
+  @doc "Takes structure, key and tree. Returns structure or error"
+  @callback process_struct(map(), binary(), Trie.t()) :: map() | {:error, String.t()}
+
+  @spec valid_store_type?(map(), atom() | list()) :: boolean()
+  def valid_store_type?(deserialized_value, store_type) when is_atom(store_type) do
+    deserialized_value.__struct__ == store_type
+  end
+
+  def valid_store_type?(deserialized_value, store_type) when is_list(store_type) do
+    Enum.any?(store_type, fn type -> deserialized_value.__struct__ == type end)
```
grzegorz225:
```
deserialized_value.__struct__ in store_type
```

apps/aecore/lib/aecore/account/account_state_tree.ex
DanielaIvanova: Good note!

apps/aecore/lib/aecore/contract/contract_state_tree.ex
DanielaIvanova: This code is already in the master. I will add the comment . This code was written @cheezus1

grzegorz225: @thepiwo I opened the issue - https://github.com/aeternity/elixir-node/issues/668

thepiwo:
@grzegorz225 I think all your comments were resolved

PR #679, opened 2018-10-16
[GH-645] Identifier types are now checked when processing transactions

### LINKED ISSUE #645: DataTx does not enforce Identifier types
In &DataTx.init_binary/6 the types of identifiers are not checked. This could lead to proper deserialization of transactions having senders of invalid types(for instance a SpendTx with an oracle as the sender).

d-velev:
closes #645 
channels txs don't have these checks in their validation yet


apps/aecore/lib/aecore/account/tx/spend_tx.ex
```diff
@@ -83,6 +83,9 @@ defmodule Aecore.Account.Tx.SpendTx do
         %DataTx{senders: senders}
       ) do
     cond do
+      !Identifier.valid?(receiver, :account) ->
+        {:error, "#{__MODULE__}: Invalid receiver identifier"}
```
thepiwo: can you add some debug data here

d-velev: yup, I've missed that one

DanielaIvanova:
@d-velev 
`respond_tx` is not working with synced `epoch` node!

thepiwo:
@d-velev are all transactions now working in sync with epoch?

PR #680, opened 2018-10-16
[GH-598] Introducing a Rox DB supplier.

### LINKED ISSUE #598: Rox DB not working if Peristence GenServer crashes
So if the Persistence GenServer, crashes, it is being restarted by it's Supervisor.
Problem is that the way RoxDB is designed, at least in the library that we use, we cannot open it again until it is closed (at least that's what I have found). 
But we cannot manually close the DB. By design it is made to be automatically closed when the BEAM VM garbage collects it: 

Quoting:
`The database will automatically be closed when the BEAM VM releases it for garbage collection.`

So if we try to open the RoxDB again, we get the following message: 
`{:error, "IO error: While lock file: /home/gspasov/1work/aeternity/elixir-node/test/LOCK: No locks available"}`

This means that we cannot get the reference of the DB, not of the families, until the VM garbage collects it, i.e. until we restart the project.

This faces us with the question of how do we deal with this issue. For me there are 2 options:
 - Go around the problem and figure out a workaround (which is not a solution in my opinion). For me this will be to use another GenServer to keep the state of the DB in 2 places, so if one of them crashes we still have the db and families references;
 - Maybe use another DB?

Artur64:
To resolve this issue , we tried following solutions:
1. **ETS tables**. 
First of all, we tried the idea of ETS tables, offered by Georgi Spasov/Dimitar Velev in the discussion of this issue. It was implemented but then we faced a problem: When we create an ETS table, it will be linked to a parent process, which in our case doesn't really resolves the problem with failing and losing all references/data needed for RoxDB. As the process dies, ETS table dies with it.
2. **Garbage collection**
Next idea was the garbage collection , as it was mentioned in official RoxDB documentation of [open/3 function](https://hexdocs.pm/rox/Rox.html#open/3) : 
> The database will automatically be closed when the BEAM VM releases it for **garbage** collection.

But as we tried many ways to collect the "garbage", for instance: At the beginning of Persistence init  process/end of it , tried to clean the memory, defining [a terminate callback](https://hexdocs.pm/elixir/GenServer.html#c:terminate/2) , which will be invoked before the server crashes, a "cleanup" before exiting the application.  
Didn't help us too, as we suppose it will collect garbage only for a given PID of a process , or if no parameters are used - the default is a self PID of a server calling this function. So we came up with  a conclusion that as a RoxDB a some kind of "NIF", "native" process. It is completely isolated from our project, so if we would like to "garbage-collect" the DB in order to close it, we will have either to enforce BEAM VM clean up this process or if it's not running on BEAM VM we should then just restart this process.

3. **Accessing, restarting isolated process**.
So this is the last solution we tried to implement. After all the research we made we tried to somehow track and restart or  kill-start  the RoxDB process again. We made many attempts in order to find it but got no success as the process is probably started outside VM , even is not, the process doesn't have a name which we can address to.

In this PR, we decided to implement and propose a simple GenServer which will store and get stored references from its state in order to "supply" the Persistence module in case of inner fails.

resolves #598


apps/aecore/lib/aecore.ex
```diff
@@ -7,6 +7,7 @@ defmodule Aecore do
 
   def start(_type, _args) do
     children = [
+      Aecore.Persistence.SupplierWorker.SupplierSupervisor,
```
thepiwo: Normally we use just `Supervisor` as name in the module directories

apps/aecore/lib/aecore/persistence/supplier.ex
```diff
@@ -0,0 +1,37 @@
...
+    GenServer.start_link(__MODULE__, %{}, name: __MODULE__)
+  end
+
+  @spec store_references(map()) :: {:error, String.t()} | :ok
+  def store_references(%{db: _, families_map: _} = references) do
+    GenServer.call(__MODULE__, {:store_references, references})
+  end
+
+  def store_references(_) do
+    {:error, "#{__MODULE__}: Error, invalid database/references information given"}
```
thepiwo: can you add debug data here, ~~given~~

d-velev:
Should we really have backup for a scenario where the persistence worker crashes? If there is some issue that causes it to crash maybe we should fix that instead of having this, what if the supplier crashes as well?

thepiwo:
@d-velev, we need a way to be on the save side here, It could be there is a bug in the node and all of a sudden all nodes in the network crash and make the database would be unusable. This solution looks like a good workaround for now. But lets await response from @cytadela8

PR #690, opened 2018-10-19
[GH-602] Testframework stability fixes

cytadela8:
This is a PR **into** branch GH-602.

This should hopefully fix stability problems with testframework. I managed to make tests run faster and reliably on my laptop.

Changes:
- a little bit less checks (maybe we should revert this, but I think it's ok)
- compare only top hashes instead of full blocks
- concurent node starting
- force kill nodes instead of gracefull stop
- harcode port numbers instead of detecting open ports
- properly checking if transaction is in pool (we have sometimes mined transactions in pool, becouse pool is synced and different nodes might get new blocks at different time)
- NEW - use mock miner (MIX_ENV=test)

TODO (as part of different PR):
- checking if transactions get mined
- consider lowering number of nodes to 3
- reuse nodes (reset nodes between tests in some way)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ iex-node:
 	@PERSISTENCE_PATH=apps/aecore/priv/rox_db_400$(NODE_NUMBER)/ PEER_KEYS_PATH=apps/aecore/priv/peerkeys_400$(NODE_NUMBER)/ SIGN_KEYS_PATH=apps/aecore/priv/signkeys_400$(NODE_NUMBER)/ PORT=400$(NODE_NUMBER) SYNC_PORT=300$(NODE_NUMBER) iex -S mix phx.server
 
 iex-test-node:
-	@PERSISTENCE_PATH=apps/aecore/priv/test_rox_db_400$(NODE_NUMBER)/ PEER_KEYS_PATH=apps/aecore/priv/test_peerkeys_400$(NODE_NUMBER)/ SIGN_KEYS_PATH=apps/aecore/priv/test_signkeys_400$(NODE_NUMBER)/ PORT=400$(NODE_NUMBER) SYNC_PORT=300$(NODE_NUMBER) iex -S mix phx.server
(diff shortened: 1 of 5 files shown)


cytadela8:
It seems killing instead of gracefully closing nodes doesn't work as expected in Travis. I will investigate later. I somebody knows a good way of forcefully closing a node while closing ports properly let me know!

cytadela8:
*I made 10 empty commits to see what (if any) Travis errors we will encounter.*

thepiwo:
@cytadela8 you can also manually trigger builds on travis directly

thepiwo:
@cytadela8 webhooks are currently disabled from github, so travis is not notified. I will start a few test manually

thepiwo:
https://travis-ci.org/aeternity/elixir-node/builds/444604784#L679

https://travis-ci.org/aeternity/elixir-node/builds/444604780#L680

cytadela8:
The node starting problem should be fixed. https://travis-ci.org/aeternity/elixir-node/builds/444604784#L679 is an actuall bug in sync I think.

cytadela8:
Just run 5 builds manually without any errors. I'm getting now: "Oh no! You’ve exceeded the limit for triggering builds, please wait a while before trying again."

thepiwo:
I did start a few as well

https://travis-ci.org/aeternity/elixir-node/builds

only one failed

https://travis-ci.org/aeternity/elixir-node/builds/444610454#L683

same issue as before, so we should be fine

cytadela8:
@thepiwo The new checking "is transaction added to Pool" tests only node2. Most of transactions are added to node2 so most of the tests can't fail becouse of pool sync. This is one of the tests that adds transactions to node1 so it sometimes fails, becouse pool isn't synced (sometimes) correctly.

thepiwo:
@cytadela8 I propose we keep it like this for now, if you don't have an idea how to solve this in mind right now

cytadela8:
I don't have any idea how to solve this. Probably some unknown bug in Pool sync has to be fixes.

cytadela8:
Yes

Issue #693, opened 2018-10-23
Check if code is formatted in travis


grzegorz225:
We should only accept code formatted with ```mix format```

grzegorz225:
Closed by https://github.com/aeternity/elixir-node/pull/694

PR #694, opened 2018-10-23
[GH-693] Check if code is formatted in travis

grzegorz225:
We should only accept code formatted via ```mix format```


thepiwo:
@aeternity/elixir what is your opinion, should we reintroduce the check?

PR #710, opened 2018-10-26
[GH-709] Remove reference to Chain.chain_state() in ContractCallTx

### LINKED ISSUE #709: Remove reference to Chain.chain_state() in contract call tx

d-velev:
closes #709

### DIFF
--- a/apps/aecore/lib/aecore/contract/tx/contract_call_tx.ex
+++ b/apps/aecore/lib/aecore/contract/tx/contract_call_tx.ex
@@ -208,7 +208,7 @@ defmodule Aecore.Contract.Tx.ContractCallTx do
 
   @spec preprocess_check(
           Chainstate.accounts(),
-          tx_type_state(),
+          Chainstate.t(),
           non_neg_integer(),
           ContractCallTx.t(),
           DataTx.t(),
--- a/apps/aecore/lib/aecore/contract/tx/contract_call_tx.ex
+++ b/apps/aecore/lib/aecore/contract/tx/contract_call_tx.ex
@@ -9,7 +9,6 @@ defmodule Aecore.Contract.Tx.ContractCallTx do
   alias Aecore.Governance.GovernanceConstants
   alias Aecore.Account.{Account, AccountStateTree}
   alias Aecore.Tx.DataTx
-  alias Aecore.Chain.Worker, as: Chain
   alias Aecore.Chain.{Identifier, Chainstate}
   alias Aecore.Contract.{Contract, Call, CallStateTree, Dispatch, ContractStateTree}
   alias Aecore.Tx.Transaction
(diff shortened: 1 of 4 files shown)


apps/aecore/lib/aecore/tx/data_tx.ex
```diff
@@ -288,7 +288,14 @@ defmodule Aecore.Tx.DataTx do
         context \\ :transaction
       ) do
     accounts_state = chainstate.accounts
-    tx_type_state = Map.get(chainstate, type.get_chain_state_name(), %{})
+    chain_state_name = type.get_chain_state_name()
+
+    tx_type_state =
+      if Enum.member?([:contracts, :calls], chain_state_name) do
+        chainstate
```
thepiwo: you require the full chain state for this, right?

apps/aecore/lib/aecore/contract/tx/contract_call_tx.ex
```diff
@@ -217,16 +216,14 @@ defmodule Aecore.Contract.Tx.ContractCallTx do
         ) :: :ok | {:error, String.t()}
   def preprocess_check(
         accounts,
-        _calls,
+        chainstate,
```
DanielaIvanova: The second argument supposed to be "Structure that holds specific transaction info in the chainstate". I am not sure if this is correct. You can check [here](https://github.com/aeternity/elixir-node/blob/master/apps/aecore/lib/aecore/tx/transaction.ex#L51).

apps/aecore/lib/aecore/tx/data_tx.ex
thepiwo: :+1:

apps/aecore/lib/aecore/contract/tx/contract_call_tx.ex
thepiwo: Yes @DanielaIvanova is right, the `@spec` of the behaviour is broken with this

d-velev: yes I know but we need the entire chainstate for this one, not only calls

thepiwo: yes, but as you do it won't work with the spec

d-velev: right

d-velev:
@DanielaIvanova The reason is that we don't want to take the current chainstate from Chain.chain_state, we want to work with the copy of the chainstate from when we know it was the correct one

apps/aecore/lib/aecore/contract/tx/contract_call_tx.ex
```diff
@@ -208,7 +208,7 @@ defmodule Aecore.Contract.Tx.ContractCallTx do
 
   @spec preprocess_check(
           Chainstate.accounts(),
-          tx_type_state(),
+          Chainstate.t(),
```
thepiwo: this is still not the correct spec, as the behaviour callbacks state `tx_type_state()`

PR #711, opened 2018-10-26
[GH-707] Adjust contract_call_tx function

### LINKED ISSUE #707: Adjust contract_call_tx function
Currently, contract call transaction is not being mined by Epoch's node, which means, that there is no compatibility to epoch.

cheezus1:
Call contract tx is now being successfully synced with epoch as well.

Resolves #707

### DIFF
--- a/apps/aecore/lib/aecore/contract/tx/contract_call_tx.ex
+++ b/apps/aecore/lib/aecore/contract/tx/contract_call_tx.ex
@@ -65,7 +65,7 @@ defmodule Aecore.Contract.Tx.ContractCallTx do
     :gas,
     :gas_price,
     :call_data,
-    :call_stack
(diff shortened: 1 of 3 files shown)


cheezus1:
Yes, that was the purpose.
What I had missed is that the REVERT opcode wasn't terminating the execution, which caused more gas to be used on our node. There were some values that were not deserialized as well.

thepiwo:
@DanielaIvanova could you provide the commands you test with?

thepiwo:
@cheezus1 Tests are still failing

cheezus1:
We can merge this now. Contract tests will be re-enabled after issue #712

thepiwo:
@cheezus1 please disable the test using the disabled tag, not commenting the file out...

Issue #714, opened 2018-10-29
Don't accept transactions with non consecutive nonce


grzegorz225:
While testing compatibility of withdraw and deposit tx with my tool(https://github.com/aeternity/elixir-node/issues/713) I noticed that epoch fails to mine the ChannelCreateTx.

The ChannelCreateTx is properly gossiped to every node in the network. In every elixir node Miner.candidate contains the ChannelCreateTx as expected. The problem is that when epoch tries to mine a block the ChannelCreateTx is not being included in the block as it does not validate.

The epoch logs contain the reason why ChannelCreateTx is not mined:
```
2018-10-29 23:58:01.319 [debug] <0.5788.0>@aec_trees:apply_txs_on_state_trees:276 Tx {aetx,channel_create_tx,aesc_create_tx,{channel_create_tx,{id,account,<<177,200,2,169,77,161,103,38,251,213,241,29,89,5,39,70,169,38,227,34,65,209,163,113,54,48,187,86,18,9,228,159>>},100,{id,account,<<86,153,201,210,211,197,179,33,24,213,195,169,97,199,198,169,43,93,55,18,202,186,118,60,237,231,25,44,73,177,224,214>>},100,5,2,0,1,<<216,210,41,252,196,75,135,247,38,209,40,199,125,2,87,26,206,201,146,15,201,5,232,243,206,90,139,218,124,255,42,72>>,2}} cannot be applied due to an error account_nonce_too_low
``` 
The nonce of the transaction is 2. The transaction in question is:
```
%Aecore.Tx.SignedTx{
    data: %Aecore.Tx.DataTx{
      fee: 1,
      nonce: 2,
      payload: %Aecore.Channel.Tx.ChannelCreateTx{
        channel_reserve: 5,
        initiator_amount: 100,
        locktime: 2,
        responder_amount: 100,
        state_hash: <<216, 210, 41, 252, 196, 75, 135, 247, 38, 209, 40, 199,
          125, 2, 87, 26, 206, 201, 146, 15, 201, 5, 232, 243, 206, 90, 139,
          218, 124, 255, 42, 72>>
      },
      senders: [
        %Aecore.Chain.Identifier{
          type: :account,
          value: <<177, 200, 2, 169, 77, 161, 103, 38, 251, 213, 241, 29, 89, 5,
            39, 70, 169, 38, 227, 34, 65, 209, 163, 113, 54, 48, 187, 86, 18, 9,
            228, 159>>
        },
[...]

PR #715, opened 2018-10-30
[GH-714] Check if nonce is consecutive

grzegorz225:
Epoch treats transactions with non consecutive nonce's as invalid transactions.

### DIFF
--- a/apps/aecore/lib/aecore/tx/data_tx.ex
+++ b/apps/aecore/lib/aecore/tx/data_tx.ex
@@ -300,7 +300,7 @@ defmodule Aecore.Tx.DataTx do
     tx_type_preprocess_check =
       type.preprocess_check(accounts_state, tx_type_state, block_height, payload, tx, context)
 
-    expected_tx_nonce = Account.nonce(chainstate.accounts, main_sender(tx, chainstate)) + 1
+    current_nonce = Account.nonce(chainstate.accounts, main_sender(tx, chainstate))
 
     cond do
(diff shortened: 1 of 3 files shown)


apps/aecore/lib/aecore/tx/data_tx.ex
```diff
@@ -310,9 +312,12 @@ defmodule Aecore.Tx.DataTx do
            block_height
          }"}
 
-      Account.nonce(chainstate.accounts, main_sender(tx, chainstate)) >= tx.nonce ->
+      cur_nonce >= tx.nonce ->
```
thepiwo: I assume these checks can be implemented as one

grzegorz225:
@thepiwo Check was refactored

apps/aecore/lib/aecore/tx/data_tx.ex
```diff
@@ -300,7 +300,7 @@ defmodule Aecore.Tx.DataTx do
     tx_type_preprocess_check =
       type.preprocess_check(accounts_state, tx_type_state, block_height, payload, tx, context)
 
-    cur_nonce = Account.nonce(chainstate.accounts, main_sender(tx, chainstate))
+    expected_tx_nonce = Account.nonce(chainstate.accounts, main_sender(tx, chainstate)) + 1
```
thepiwo: I'd split math and getter here and have the maths in the check itself

grzegorz225:
@thepiwo Refactored

PR #717, opened 2018-10-31
[GH-716] moved environment variable reading from config to code

### LINKED ISSUE #716: Move environment variable read from config to code

thepiwo:
This was needed to start the release with a different SYNC_PORT than epoch.

resolves #716

### DIFF
--- a/apps/aecore/lib/aecore/pow/pow.ex
+++ b/apps/aecore/lib/aecore/pow/pow.ex
@@ -1,6 +1,6 @@
 defmodule Aecore.Pow.Pow do
   @moduledoc """
-  A meta Proof of work alorithm that invokes proper allgorithm for current environment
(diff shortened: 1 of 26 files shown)


thepiwo:
Using environment variables as usual. Either setting them using `export VAR=val` or when running `VAR=Val ./executable`

apps/aecore/lib/aecore/pow/pow.ex
```diff
@@ -1,6 +1,6 @@
 defmodule Aecore.Pow.Pow do
   @moduledoc """
-  A meta Proof of work alorithm that invokes proper allgorithm for current environment
+  A meta Proof of work alorithm that invokes proper allgorithm for current environment.ex
```
grzegorz225: Typo

Issue #722, opened 2018-11-01
Spellcheck our entire code base


grzegorz225:
I made a bash one liner to spell check the entire code base using Aspell.
We should add this script to master and run it from time to time.

grzegorz225:
Running spell_check_diff_master.sh shows properly the typos in an interactive manner: 
```
./apps/aecore/lib/aecore/pow/cuckoo.ex:3          A librarye providing Cuckoo Cycle PoW generation\
and verification.
./apps/aecore/lib/aecore/pow/mock.ex:3    Provides a mock proofd of work that always succeeds     \
generation and uses predefined integers as proof. In validation checks that integers are as       \
expected
```

PR #723, opened 2018-11-01
[GH-722] Spellcheck our entire code base

grzegorz225:
I've spell checked our entire code base using aspell and removed a lot of typos.

The PR includes bash scripts for spellchecking the entire code base and spellchecking the difference between the current and master branch. The scripts require that aspell(interactive spell checker) and pcregrep(grep with Perl-compatible regular expressions) is installed.


thepiwo:
Can you move the scripts in a scripts folder?

PR #728, opened 2018-11-03
[GH-721] Miner crash workaround

### LINKED ISSUE #721: Miner crashes sometimes
When using the lean16 miner our miner process sometimes crashes.
On a clean chain the miner crashes after mining 10-20 blocks.
When I tried using the mean28-avx2 miner the same problem occurs.

grzegorz225:
This PR will prevent the miner from crashing on long runs.
I've tried to find the underlying reason for the crashes but I couldn't find it. It seems that very rarely in the multi threaded native miner a race condition occurs which crashes the miner. I've added the functionality of ignoring these crashes.

resolves #721

### DIFF
--- a/apps/aecore/lib/aecore/miner/worker.ex
+++ b/apps/aecore/lib/aecore/miner/worker.ex
@@ -94,13 +94,25 @@ defmodule Aecore.Miner.Worker do
   end
 
   defp mine_sync_block(
-         {:error, reason},
+         {:error, :no_solution},
          %Block{header: %Header{nonce: nonce} = header} = cblock
-       )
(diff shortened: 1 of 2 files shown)


apps/aecore/lib/aecore/miner/worker.ex
```diff
@@ -96,6 +102,19 @@ defmodule Aecore.Miner.Worker do
     mine_sync_block(Pow.generate(cheader), cblock)
   end
 
+  defp mine_sync_block(
```
d-velev: Is this case needed when we have the check for != {:error, :miner_was_stopped} above?

PR #784, opened 2018-11-13
[GH-735] Adjusted commitment hash function

### LINKED ISSUE #735: Adjust commitment hash calculations in naming system
To be Hash(NameHash(name) + name_salt) instead of Hash(Hash(name + name_salt))

Artur64:
Resolves #735 . Should be tested and merged after #782 is merged.

### DIFF
--- a/apps/aecore/lib/aecore/naming/name_commitment.ex
+++ b/apps/aecore/lib/aecore/naming/name_commitment.ex
@@ -46,7 +46,7 @@ defmodule Aecore.Naming.NameCommitment do
   def commitment_hash(name, name_salt) when is_integer(name_salt) do
     case NameUtil.normalize_and_validate_name(name) do
       {:ok, normalized_name} ->
-        hash_name = Hash.hash(normalized_name)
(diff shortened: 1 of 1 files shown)


thepiwo (commented):
Did you check if the commitment hash is the same as in epoch? For the old version it was the same as in epoch 0.16. but now the logic for hash creation is changed. Is this as intended?

Artur64:
@thepiwo  I'm going to test hashes after all blockers are merged in master, as i mentioned before in the description of this PR.

cytadela8:
I think it would be nice if you add a simple test that creates a hash using our code of (for example) "exampledomain.test" and compares it against a hash you obtained manually in epoch. This way we make sure everything will stay working as expected even if we make some changes to Naming. 

This is of course to be done after the issue is unblocked.

Artur64:
@thepiwo Commitment hashes being produced by our and Epoch's node are equal, so i'm ready with this PR.

DanielaIvanova (approved):

thepiwo (approved):

Issue #736, opened 2018-11-05
Adjust miner reward chainstate calculation


d-velev:
The reward for the current leader is the following: coinbase + 0.6 * sum of fees in the micro blocks in previous generation + 0.4 * sum of fees in current generation

thepiwo:
Can you add information on how to adjust

cytadela8:
I think this issue might be more deffictult then it seems. New rewards calculation requires access to previous block. Also if I remember correctly bitcoin-ng requires locking of rewards.

I belive this issue intersects with "Adjust block validation..."

d-velev:
@cytadela8 I think it will just require some adjustments to calculating reward in chainstate, validation checks if the chainstate is calculated correctly so it shouldn't be touched right? I guess we will need to keep track of when to apply rewards as well

d-velev:
blocked by #742 (requires getting previous generation from chain)

thepiwo:
closed by https://github.com/aeternity/elixir-node/pull/798

Issue #727, opened 2018-11-02
Adjust block validation to work with key/micro blocks


d-velev:

thepiwo:
Can you add information on what to adjust?

thepiwo:
@d-velev I assume this was closed by #792 ? If not, reopen

PR #786, opened 2018-11-14
[GH-726] NG mining

### LINKED ISSUE #726: Adjust candidate block generation/mining
Key block candidates should be generated and mined while the node isn't the current leader, micro blocks should be produced when the node is leading

### LINKED ISSUE #776: Adjust genesis block
Should be a key block

d-velev:
closes #726 closes #776 
This also adjusts the genesis block.
things to be done later on: #736 #727 and tests to be refactored when everything needed is added


thepiwo:
With what issues will this be fixed? If these are actual "features" its okay to do later, if this is within the scope of this feature it should be fixed here

d-velev:
@thepiwo block validation and miner reward calculation adjustments are a separate issue, also tests need to be redone in some way so that transactions pass without any balance, I think this is a minimal solution that doesn't change too many things, with more of the issues resolved at some point I'll probably do another issue that fixes tests and removes all old code

thepiwo:
@d-velev could you reference these issues here, so everyone can see

cytadela8 (commented):
Just a question.

apps/aecore/lib/aecore/chain/chainstate.ex
```diff
@@ -64,27 +64,29 @@ defmodule Aecore.Chain.Chainstate do
...
-          {:ok, updated_chainstate} ->
-            {:cont, updated_chainstate}
-
-          {:error, reason} ->
-            {:halt, reason}
-        end
-      end)
-
-    final_chainstate =
-      calculate_miner_reward_chain_state(txs, updated_chainstate, block_height, miner)
```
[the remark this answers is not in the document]
cytadela8: Ok.

cytadela8 (approved):

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -224,31 +243,29 @@ defmodule Aecore.Chain.Worker do
     end
   end
 
-  @spec add_block(Block.t()) :: :ok | {:error, String.t()}
-  def add_block(%Block{header: %Header{prev_hash: prev_hash}} = block) do
+  @spec add_block(Block.t(), boolean()) :: :ok | {:error, String.t()}
+  def add_block(
+        %{header: %{prev_hash: prev_hash, prev_key_hash: prev_key_hash, height: height}} = block,
```
grzegorz225: No type safety here. Add "when block.__struct__ in [KeyBlock, MicroBlock]"
The typespec is wrong.

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -224,31 +243,29 @@ defmodule Aecore.Chain.Worker do
...
+      add_validated_block(block, new_chain_state, loop_micro_block)
     else
       err -> err
     end
   end
 
-  @spec add_validated_block(Block.t(), Chainstate.t()) :: :ok
-  defp add_validated_block(%Block{} = block, chain_state) do
-    GenServer.call(__MODULE__, {:add_validated_block, block, chain_state})
+  @spec add_validated_block(Block.t(), Chainstate.t(), boolean()) :: :ok
```
grzegorz225: Typespec

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -338,19 +355,47 @@ defmodule Aecore.Chain.Worker do
 
   def handle_call(
         {:add_validated_block,
-         %Block{
-           header: %Header{prev_hash: prev_hash, height: height, target: target} = header,
-           txs: txs
-         } = new_block, new_chain_state},
+         %{
+           header: %{prev_hash: prev_hash, height: height, time: time} = header
```
grzegorz225: Add type guard check.

gorbak25 (changes requested):
Partial review.

apps/aecore/lib/aecore/chain/worker.ex
d-velev: I don't think there's a need for explicit type checking in this case, we do checks in other functions/fragments that are part of this

apps/aecore/lib/aecore/util/header.ex
```diff
@@ -0,0 +1,28 @@
+defmodule Aecore.Util.Header do
+  @moduledoc """
+  Header utilities
+  """
+  alias Aecore.Chain.KeyHeader
+  alias Aecore.Chain.MicroHeader
+
+  @spec hash(KeyHeader.t() | MicroHeader.t()) :: binary()
+  def hash(header) do
+    case header do
```
grzegorz225: This case can be removed by specifying two clauses in the hash function, eg:
```
def hash(%KeyHeader{}) do
...
end
def hash(%MicroHeader()) do
...
end
...
```
Or by writing:
 ``` 
def hash(header) when header.__struct__ in [KeyHeader, MicroHeader] do
    header.__struct__.hash(header)
end
```

File: apps/aecore/lib/aecore/util/header.ex [hunk 6]
```diff
@@ -0,0 +1,28 @@
...
+      %KeyHeader{} ->
+        KeyHeader.hash(header)
+
+      %MicroHeader{} ->
+        MicroHeader.hash(header)
+    end
+  end
+
+  def top_key_block_hash(prev_header) do
+    case prev_header do
```
grzegorz225: Same as above

gorbak25 (commented):
Looks good.

apps/aecore/lib/aecore/chain/chainstate.ex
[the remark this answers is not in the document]
thepiwo: what issue is it, can you reference here for all to see?

apps/aecore/lib/aecore/util/header.ex
d-velev: Code reduction is always nice, will change it 😆 I dislike the second one

gorbak25 (changes requested):
Could you make sure that the code compiles without warnings?
https://travis-ci.org/aeternity/elixir-node/builds/455887419#L600

cytadela8:
@grzegorz225 This is "In progress" code. A lot of work is done here and some warnings are due to unimplemented features. (They are warning for example that something isn't used were it should be. I think proper way of fixing such warning is by implementing NG)

gorbak25 (approved):

DanielaIvanova (approved):

PR #788, opened 2018-11-15
[GH-737] Added functionality to get a gas price for all transactions 

### LINKED ISSUE #737: Add gas field to all transactions
All transactions should have a gas field in it.

Artur64:
Resolves #737 .

### DIFF
--- a/apps/aecore/lib/aecore/contract/tx/contract_call_tx.ex
+++ b/apps/aecore/lib/aecore/contract/tx/contract_call_tx.ex
@@ -254,11 +254,6 @@ defmodule Aecore.Contract.Tx.ContractCallTx do
     DataTx.standard_deduct_fee(accounts, block_height, data_tx, fee)
   end
 
-  @spec gas :: non_neg_integer()
-  def gas do
-    Constants.call_gas_price_multiplier() * GovernanceConstants.block_gas_limit()
-  end
-
   @spec encode_to_list(ContractCallTx.t(), DataTx.t()) :: list()
   def encode_to_list(
         %ContractCallTx{
--- a/apps/aecore/lib/aecore/contract/tx/contract_call_tx.ex
+++ b/apps/aecore/lib/aecore/contract/tx/contract_call_tx.ex
@@ -254,8 +254,8 @@ defmodule Aecore.Contract.Tx.ContractCallTx do
     DataTx.standard_deduct_fee(accounts, block_height, data_tx, fee)
   end
 
-  @spec gas() :: non_neg_integer()
-  def gas() do
+  @spec gas :: non_neg_integer()
+  def gas do
     Constants.call_gas_price_multiplier() * GovernanceConstants.block_gas_limit()
   end
 
--- a/apps/aecore/lib/aecore/contract/tx/contract_call_tx.ex
+++ b/apps/aecore/lib/aecore/contract/tx/contract_call_tx.ex
@@ -14,6 +14,7 @@ defmodule Aecore.Contract.Tx.ContractCallTx do
   require Aecore.Contract.ContractConstants, as: Constants
 
   @version 1
+  @gas_price_multiplier 30
 
   @typedoc "Reason of the error"
   @type reason :: String.t()
@@ -327,6 +328,11 @@ defmodule Aecore.Contract.Tx.ContractCallTx do
   @spec is_minimum_fee_met?(DataTx.t(), tx_type_state(), non_neg_integer()) :: boolean()
   def is_minimum_fee_met?(%DataTx{fee: fee}, _chain_state, _block_height) do
     fee >= GovernanceConstants.minimum_fee()
+  end
+
+  @spec gas_price :: non_neg_integer()
+  def gas_price do
+    GovernanceConstants.default_tx_gas_price() * @gas_price_multiplier
   end
 
   defp run_contract(
--- a/apps/aecore/lib/aecore/contract/tx/contract_call_tx.ex
+++ b/apps/aecore/lib/aecore/contract/tx/contract_call_tx.ex
@@ -254,6 +254,11 @@ defmodule Aecore.Contract.Tx.ContractCallTx do
     DataTx.standard_deduct_fee(accounts, block_height, data_tx, fee)
   end
 
+  @spec gas() :: non_neg_integer()
+  def gas() do
+    Constants.call_gas_price_multiplier() * GovernanceConstants.block_gas_limit()
+  end
+
   @spec encode_to_list(ContractCallTx.t(), DataTx.t()) :: list()
   def encode_to_list(
         %ContractCallTx{
--- a/apps/aecore/lib/aecore/governance/governance_constants.ex
+++ b/apps/aecore/lib/aecore/governance/governance_constants.ex
@@ -2,6 +2,8 @@ defmodule Aecore.Governance.GovernanceConstants do
   @moduledoc """
   Module containing governance constants
   """
+  alias Aecore.Contract.Tx.ContractCallTx
+  alias Aecore.Contract.Tx.ContractCreateTx
 
   @number_of_blocks_for_target_recalculation 10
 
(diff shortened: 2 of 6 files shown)


apps/aecore/lib/aecore/contract/tx/contract_call_tx.ex
```diff
@@ -254,6 +254,11 @@ defmodule Aecore.Contract.Tx.ContractCallTx do
     DataTx.standard_deduct_fee(accounts, block_height, data_tx, fee)
   end
 
+  @spec gas :: non_neg_integer()
+  def gas do
+    Constants.call_gas_price_multiplier() * GovernanceConstants.block_gas_limit()
```
cytadela8: What does this function return? Is it the gas price of transaction? The calculation seems weird.

apps/aecore/lib/aecore/tx/transaction.ex
```diff
@@ -7,14 +7,21 @@ defmodule Aecore.Tx.Transaction do
...
+
       @behaviour Transaction
 
       @spec chainstate_senders?() :: boolean()
       def chainstate_senders?() do
         false
       end
 
-      defoverridable chainstate_senders?: 0
+      @spec gas :: non_neg_integer()
```
cytadela8: You could probably document here what the `gas` function returns.

apps/aecore/lib/aecore/contract/tx/contract_call_tx.ex
Artur64: As it was stated in the issue's, PR's, function's  name:

> [**Add get base tx gas function** to all transactions](https://github.com/aeternity/elixir-node/issues/737)
>  [Added **functionality to get a gas price** for all transactions](https://github.com/aeternity/elixir-node/pull/788)
> [def **gas** do](https://github.com/aeternity/elixir-node/pull/788/files/7fc1ee12191921a20062288705abf82fe48e4cea#diff-4d3873e694a8fe0632d13c6224d6aea5R258) ...

It surely **returns**: 
> the gas price of transaction

Regarding calculation: Personally, I don't see any problems with it, if you don't like `*` (multiplier) operation, you are free to replace it with the implementation, for instance, involving `+` operation.

cytadela8: I find it weird that gas price of transaction is bigger then block_gas_limit. For example:

for ContractCallTx it's: `Constants.call_gas_price_multiplier() * GovernanceConstants.block_gas_limit()`
where `call_gas_price_multiplier` is 30 so this results with value bigger then block_gas_limit.

apps/aecore/lib/aecore/tx/transaction.ex
thepiwo: returning the limit as gas price for each transaction doesn't seem to make sense

thepiwo (changes requested):

Artur64:
@thepiwo , @cytadela8 
As it is implemented in Epoch:

Artur64:
I will make an adjustments, it seems i mixed up everything.

apps/aecore/lib/aecore/governance/governance_constants.ex
```diff
@@ -42,6 +44,31 @@ defmodule Aecore.Governance.GovernanceConstants do
 
   @max_txs_per_block 10_946
 
+  @default_tx_gas_price 15_000
+
+  @allowed_txs_types [
```
thepiwo: change the name to be more semantic "allowed for what?"

thepiwo: if any new transaction type has the need to be added here, you should add it to the dev documentation

apps/aecore/lib/aecore/governance/governance_constants.ex
```diff
@@ -101,5 +128,22 @@ defmodule Aecore.Governance.GovernanceConstants do
...
   def max_txs_per_block, do: @max_txs_per_block
 
+  @spec block_gas_limit :: non_neg_integer()
   def block_gas_limit, do: @block_gas_limit
+
+  @spec default_tx_gas_price :: non_neg_integer()
+  def default_tx_gas_price, do: @default_tx_gas_price
+
+  @spec tx_base_gas(atom()) :: non_neg_integer()
+  def tx_base_gas(type) when type in @allowed_txs_types do
```
thepiwo: I think this makes more sense, where the fee calculation was previously in code. Its not a constant.

thepiwo (changes requested):

apps/aecore/lib/aecore/governance/governance_constants.ex
```diff
@@ -46,7 +44,7 @@ defmodule Aecore.Governance.GovernanceConstants do
 
   @default_tx_gas_price 15_000
 
-  @allowed_txs_types [
+  @known_tx_types [
```
thepiwo: why is this "governance" constant? I think of this more related to the transactions calculation.

thepiwo: "known..." doesn't explain well what its used for and why we would need it

apps/aecore/lib/aecore/tx/transaction.ex
```diff
@@ -18,10 +19,14 @@ defmodule Aecore.Tx.Transaction do
 
       @spec gas_price :: non_neg_integer()
       def gas_price do
-        GovernanceConstants.tx_base_gas(__MODULE__)
+        if Enum.member?(GovernanceConstants.get_valid_txs_type(), __MODULE__) do
+          GovernanceConstants.default_tx_gas_price()
+        else
+          GovernanceConstants.block_gas_limit()
```
thepiwo: my old comment regarding this still applies

apps/aecore/lib/aecore/governance/governance_constants.ex
Artur64: Could you suggest a proper name then?

apps/aecore/lib/aecore/tx/transaction.ex
Artur64: Could you clarify your thoughts on this? I don't really get it

apps/aecore/lib/aecore/governance/governance_constants.ex
thepiwo: right now its only used for gas calculation, so maybe `gas_calculation_tx_types`

apps/aecore/lib/aecore/tx/transaction.ex
thepiwo: "returning the limit as gas price for each transaction doesn't seem to make sense"

Why is the `block_gas_limit` returned if we don't know this transaction type?

apps/aecore/lib/aecore/governance/governance_constants.ex
Artur64: Thank you for your clarifications. Well, when I was defining this constant, I also thought about that it could be reused in somewhere else too , that's why I used more generic name for it. I will change it.

apps/aecore/lib/aecore/tx/transaction.ex
Artur64: This is how Epoch node [handles it](https://github.com/aeternity/epoch/blob/v0.25.0/apps/aecore/src/aec_governance.erl#L121). They allow "unknown" transactions,  but the gas price for it is, by default: `block gas limit`. They might handle it somewhere , but I don't think that this in the scope of this issue.

thepiwo: Okay, thanks for clarifying

thepiwo (approved):

DanielaIvanova (approved):

apps/aecore/lib/aecore/tx/transaction.ex
d-velev: How would you get an unknown transaction unless you write the code yourself? Is this really needed?

PR #794, opened 2018-11-20
[GH-733] Implemented garbage collection function for transactions pool 

### LINKED ISSUE #733: Implement garbage collection function for transactions pool
Implement garbage collection function, which deletes expired by ttl transactions from the pool

DanielaIvanova:
resolves #733
@cytadela8 
We are having some problems with multi node sync test. Could you help us?


cytadela8:
@DanielaIvanova The problem is weird. I looked at it briefly, confirmed it and didn't find any reason for it. I will have to look deeper into it.

cytadela8:
@DanielaIvanova The problem is you call `Pool.garbage_collection` inside `Chain.handle_call(:add_validated_block`, but `Pool.garbage_collection` calls `Chain.top_height`. `Chain.top_height` can't finish till the previous call finishes, but the previous call can't finish till `Pool.garbage_collection` finishes...

cytadela8:
I didn't trigger the bug when testing your code for some reason, but this should be it.

cytadela8:
To fix this you can change `Pool.garbage_collect` into a cast so it will be executed in parallel and won't block `Chain.handle_call(:add_validated_block`

DanielaIvanova:
@cytadela8 
We call `Pool.garbage_collection(height)` inside `Chain.handle_call(:add_validated_block)` and we are passing  the height of the incoming block, **not** `Chain.top_height()`. We switched to `handle_cast` but the problem still persists.

DanielaIvanova:
We tested with 3 synced nodes. 
1. Synced all of them
2. `Node2` added transaction to the pool with `ttl = 4` **(`Node2` doesn't have money)**
3. `Node1` added transaction to the pool as well, with `ttl = 4` **(`Node3` doesn't have money)**
4. `Node0` is the **miner**. Those transactions persist in the pool(currently `height = 0`). 

On `height = 4` the transactions are still in pool. On `height = 5` the **pool is empty**.

This manual test proves that the garbage collector works as intended with synced nodes as well.
We don't know where the problem is coming from.

thepiwo (approved):

cytadela8 (approved):

PR #792, opened 2018-11-19
[GH-727] Block validation

d-velev:
closes #727 
A check for block gas limit (tx limit) should be added as well when that is done

PR #798, opened 2018-11-22
[GH-736] Beneficiary reward calculation

### LINKED ISSUE #736: Adjust miner reward chainstate calculation

d-velev:
closes #736 
Each generation's fees are stored in a cache, whenever the block that applies the reward for that generation is processed, that cache entry is cleared. The cache should also be persisted but I'll leave that for the persistence issue.


apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -3,115 +3,153 @@ defmodule Aecore.Chain.BlockValidation do
...
 
-  @spec single_validate_block(Block.t()) :: :ok | {:error, String.t()}
-  def single_validate_block(
-        %Block{
-          header: %Header{txs_hash: txs_hash, version: version} = header,
-          txs: txs
-        } = block
-      ) do
-    block_txs_count = length(txs)
+  defp validate_key_block(
```
thepiwo: Maybe we should move this code to the block itself , as `validate`

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -3,115 +3,153 @@ defmodule Aecore.Chain.BlockValidation do
...
+
+      !is_solution_valid?(header) ->
+        {:error, "#{__MODULE__}: Invalid PoW solution"}
+
+      true ->
+        :ok
+    end
+  end
+
+  defp validate_micro_block(
```
thepiwo: Maybe we should move this code to the micro block itself , as `validate`

apps/aecore/lib/aecore/chain/chainstate.ex
```diff
@@ -106,23 +106,51 @@ defmodule Aecore.Chain.Chainstate do
...
-          AccountStateTree.update(chainstate.accounts, miner_pubkey, fn acc ->
+      next_generation_beneficiary =
+        (block_height + 1) |> Chain.generation_fees() |> Map.keys() |> List.first()
+
+      accounts_with_current_beneficiary_reward =
+        AccountStateTree.update(chainstate.accounts, current_generation_beneficiary, fn acc ->
+          Account.apply_transfer!(
+            acc,
+            block_height,
+            round(
```
thepiwo, line 134: Is `round` correct here?

apps/aecore/lib/aecore/chain/chainstate.ex
```diff
@@ -106,23 +106,51 @@ defmodule Aecore.Chain.Chainstate do
...
+          next_generation_beneficiary,
+          fn acc ->
             Account.apply_transfer!(
               acc,
               block_height,
-              GovernanceConstants.coinbase_transaction_amount() +
-                Miner.calculate_miner_reward(txs, chainstate)
+              round(
+                current_generation_fees *
+                  GovernanceConstants.previous_generation_fee_reward_multiplier()
```
thepiwo, line 154: Why is the constant `previous`, not `next`?

d-velev: I wrote a comment above it, the generation is previous to the one that was generated by the next leader.

apps/aecore/lib/aecore/chain/chainstate.ex
d-velev: yes, we need this to work with integers when serializing (it's done this way in epoch as well)

apps/aecore/lib/aecore/chain/block_validation.ex
d-velev: I'll do this in the block validation PR

apps/aecore/lib/aecore/chain/chainstate.ex
thepiwo: okay, thanks

apps/aecore/lib/aecore/chain/chainstate.ex
thepiwo: okay

thepiwo: could you send a link to where this is done in epoch?

apps/aecore/lib/aecore/chain/block_validation.ex
thepiwo: ok

apps/aecore/lib/aecore/chain/chainstate.ex
d-velev: https://github.com/aeternity/epoch/pull/1585
It was changed here, I guess I haven't updated my master in a long time, I'll adjust this tomorrow as this makes the current beneficiary get less of a reward

thepiwo: thanks

thepiwo:
so as I do understand, this is blocked by https://github.com/aeternity/elixir-node/pull/792

apps/aecore/lib/aecore/chain/block_validation.ex
```diff
@@ -3,115 +3,153 @@ defmodule Aecore.Chain.BlockValidation do
...
+           header: %{time: prev_block_time, prev_key_hash: prev_key_hash} = prev_header
+         } = prev_block
+       ) do
+    prev_key_block =
+      case prev_block do
+        %KeyBlock{} ->
+          prev_block
+
+        %MicroBlock{} ->
+          {:ok, key_block} = Chain.get_block(prev_key_hash)
```
cytadela8: I think it would be better if you would pass `prev_key_block` as argument to this function and `calculate_and_validate_block`.
1. This is how we handle `prev_block`.
2. This way we don't do calls to Channel.Worker. (We remove circular dependency between models.)

apps/aecore/lib/aecore/chain/chainstate.ex
```diff
@@ -106,23 +106,51 @@ defmodule Aecore.Chain.Chainstate do
     }
   end
 
-  defp calculate_miner_reward_chain_state(txs, chainstate, block_height, miner) do
-    case miner do
-      @genesis_miner ->
-        chainstate
+  defp calculate_miner_reward_chain_state(chainstate, block_height) do
+    if block_height > GovernanceConstants.beneficiary_reward_lock_time() do
+      current_generation_beneficiary =
```
cytadela8, line 120: How does this work? ~I don't think `Map.keys` guarantees returning in the order of adding.~ I think it would be better to to get the header of block at proper height.

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -66,7 +66,8 @@ defmodule Aecore.Chain.Worker do
        top_hash: genesis_block_hash,
        top_height: 0,
        total_diff: Persistence.get_total_difficulty(),
-       node_is_leader: false
+       node_is_leader: false,
+       generation_fees: %{}
```
cytadela8, line 70: I think it would be better to move the data stored here to blocks (to the unnamed block_info map (the one with "refs"))

apps/aecore/lib/aecore/chain/chainstate.ex
cytadela8: After further inspection - I see that this will be the height of beneficiary block not pubkey.
So:
1. This height could be calculated with some simple substraction
2. It's still better to get the header by height from Chain.Worker in case of forks.

apps/aecore/lib/aecore/chain/worker.ex
```diff
@@ -421,10 +440,11 @@ defmodule Aecore.Chain.Worker do
     state_update = %{
       state
       | blocks_data_map: hundred_blocks_data_map,
-        node_is_leader: new_node_is_leader
+        node_is_leader: new_node_is_leader,
+        generation_fees: new_generation_fees
     }
 
-    if top_height < height do
+    if top_height <= height && !block_is_genesis do
```
cytadela8, line 475: I'm not sure about the Bitcoin-NG consensus rules, but this part of `:add_validated_block` handles changing top hash. I'm woried we might change to a fork with the same height or one of the previous microblocks.

To be sure we don't do that previously checking for increasing heights was enough. Now we have micro blocks with the same height. I think we could check now `top_height < height || (top_height == height && top_timestamp < timestamp)`

apps/aecore/lib/aecore/miner/worker.ex
```diff
@@ -303,26 +302,26 @@ defmodule Aecore.Miner.Worker do
...
+    prev_key_hash = Header.top_key_block_hash(top_block.header)
+
+    minimum_distance =
+      if prev_hash == prev_key_hash do
+        @minimum_distance_from_key_block
+      else
+        GovernanceConstants.micro_block_distance()
+      end
+
+    :timer.sleep(minimum_distance)
```
cytadela8, line 328: This function is called by `Channel.Worker` I think we shouldn't have any "sleeps" here, because this would block the Channel.Worker. It would be better to have such stuff in some thread maybe Miner GenServer. (make a cast on the genserver, cast-s don't block) Generally handling of microblock mining should be improved. Probably not in this PR. Probably it would be better to have a separate module from that, as mining MicroBlocks is something completely different then mining KeyBlocks.

cytadela8 (dismissed):
Some comments. I think only `calculate_miner_reward_chain_state` comment is important. Rest are about style.

apps/aecore/lib/aecore/miner/worker.ex
d-velev: I've thought about having a separate module for micro block generation, decided that it could be done a bit later on as it works this way as well. And as for the sleep, maybe it should be done before sending the new block to our peers.

apps/aecore/lib/aecore/chain/worker.ex
d-velev: It's still more difficult chain wins (a shorter chain that is more difficult should win), we don't have that really (I guess except when syncing?) . I haven't thought about it too much but it's probably best to leave it for another issue.

apps/aecore/lib/aecore/chain/chainstate.ex
d-velev: Ok, this would mean removing the cache in general, don't you think it would be quite expensive to get all of the generation micro blocks from the db and iterate over them to get the fees again?

d-velev: How would we know which fork we are getting blocks from if we are getting them by height? 

>  I think it would be better to to get the header of block at proper height.

What do you mean by this? Just to get the beneficiary key?

apps/aecore/lib/aecore/chain/chainstate.ex
```diff
@@ -61,20 +60,21 @@ defmodule Aecore.Chain.Chainstate do
   end
 
   @spec calculate_and_validate_chain_state(
-          list(),
+          KeyBlock.t() | MicroBlock.t(),
           Chainstate.t(),
           non_neg_integer()
         ) :: {:ok, Chainstate.t()} | {:error, String.t()}
   def calculate_and_validate_chain_state(block, chainstate, block_height) do
```
gorbak25: You can remove the case statement by matching the type of a block in the function definition:
calculate_and_validate_chain_state(%KeyBlock{}, chainstate, block_height) ...
calculate_and_validate_chain_state(%MicroBlock{}, chainstate, block_height) ...

apps/aecore/lib/aecore/chain/chainstate.ex
```diff
@@ -106,23 +106,51 @@ defmodule Aecore.Chain.Chainstate do
     }
   end
 
-  defp calculate_miner_reward_chain_state(txs, chainstate, block_height, miner) do
-    case miner do
-      @genesis_miner ->
-        chainstate
+  defp calculate_miner_reward_chain_state(chainstate, block_height) do
```
gorbak25, line 118: If we have a fork like:
```
           [Gen2a] - ... - [Gen#{GovernanceConstants.beneficiary_reward_lock_time()+1}a] 
           //
[Gen1]--
            \\
            [Gen2b] - ... - [Gen#{GovernanceConstants.beneficiary_reward_lock_time()+1}b]
```
The reward calculation will apply wrong rewards on one of the branches. This situation could happen if a group of miners decides to build blocks on top of the mallicoius chain instead of building on top the most difficult chain. This could lead to miscalculations in the main chain which could invalidate the the main chain and lead to double spend attacks.

Make sure on what fork the block reward was calculated.

apps/aecore/lib/aecore/chain/worker.ex
gorbak25: Are we validating if the new block is linked to the genesis block?

apps/aecore/lib/aecore/chain/chainstate.ex
gorbak25: @cytadela8 
https://github.com/aeternity/elixir-node/pull/798/files#diff-76c70afbd7cea50d60f10c30ae67b41cR555
We are querying the beneficiary pub key.

We are not checking whether we are applying the rewards for the correct fork (see comment above)

gorbak25 (changes requested):
Looks good. Fork handling while calculating rewards needs to be introduced.

thepiwo:
I assume this is unblocked now

apps/aecore/lib/aecore/chain/worker.ex
d-velev: Isn't it impossible for it not to be linked to the genesis block if the genesis is always our first block and the next block has to be validated in pair with it?

apps/aecore/lib/aecore/chain/chainstate.ex
d-velev: I know, we need to handle forks generally, not only when calculating rewards. Currently the check for new blocks in add_validated_block is if top_height <= new_block_height, we should be comparing difficulties not height. Haven't thought about a way but I won't do it here...

thepiwo:
@d-velev whats the status of this? will you merge master-mg in here?

d-velev:
@thepiwo I have to move the sleep before adding micro blocks, I'll get this ready to merge in an hour or so

thepiwo:
@grzegorz225 @DanielaIvanova could you review?

thepiwo (approved):
should be good for now

DanielaIvanova (approved):

gorbak25 (approved):

Issue #810, opened 2018-11-29
Get Key Block by height


cytadela8:
Adjust get_block_by_height functions family for key blocks.

thepiwo:
resolved by #811

PR #811, opened 2018-11-29
[GH-810] Adjust get block by height for KeyBlocks

cytadela8:
Small adjustments to return proper KeyBlocks when calling get_block_by_height functions family.

resolves #811


thepiwo (approved):
thank you

d-velev (approved):

